PR #57, opened 2014-09-24
BUILD: Add scripts to parse all Makefile.am and build with CMake.

berenm:
As the other PR (#32) has been closed because of branch rewrite stuff, here it is again for the record.

This is the same AutoMake parsing stuff as before so that the CMake build system is automatically kept (more or less) in sync with the Autotools one.

There also is a separate commit for unit testing with Catch and a sample test suite for the TransformationMatrix class.


cc9cii:
This is great, thanks very much. I'm on windows and this PR allows me to get going (not that xoreos compiles with MSVC, it appears that system.h, stdlib.h, etc, being included have different signatures).

By the way, I'm using GUI version of CMake, v2.8.12.2.  I had to make some changes, replacing [[ with " and adding extra escapes, etc.

DrMcCoy:
> Why isn't it accepted/merged?

That's essentially down to me not liking CMake. At all. I like the idea of parsing the autotools files, and I'm really not happy with shutting out MSVC-using people either. But, yeah, I can't stand CMake, for various reasons I'm not going to get into again here.

I *still* [1] think the better solution is to take [ScummVM's create_project](https://github.com/scummvm/scummvm/tree/master/devtools/create_project), which spits out MSVC project files (even on non-Windows systems, something CMake doesn't for some reason) and adapt that to parse our autotools files.

[1] Even though it gets unlikelier by the day that someone interested in doing this comes along, unfortunately :/

> not that xoreos compiles with MSVC, it appears that system.h, filepath.h, etc needs some changes

I would, however, be highly interested in what these changes are, and integrate those into the main xoreos tree. I don't have a running Windows, so I can't test any of this, and xoreos compiles fine into a Windows binary with a cross-compiling MinGW.

berenm:
> By the way, I'm using GUI version of CMake, v2.8.12.2. I had to make some changes, replacing [[ with " and adding extra escapes, etc.

This is a syntax introduced in CMake 3.0.0, it was to avoid too much quote escaping in the strings.

> Even though it gets unlikelier by the day that someone interested in doing this comes along, unfortunately :/

Probably because CMake does it well enough already, and that *many* people are using CMake, and *very few* are interested in writing build-system related code.

> That's essentially down to me not liking CMake.

The configure step of this CMake PR takes around two second when the original autotools checks take almost a minute on my brand new SSD. Just trolling :).

cc9cii:
> I would, however, be highly interested in what these changes are,

Sure, once I get things working I'll submit a PR.  At the moment it's not quite working (2DA.zip is actually in Data directory)

> C:\GOG Games\Neverwinter Nights 2 Complete>xoreos.exe -p.
WARNING: No target specified, but found a target with a matching path!
Target "xoreos"
Initialising shaders...
shader default/default.vert loaded
shader default/default.frag loaded
shader default/color.frag loaded
Initialising default surfaces...
Initialising default materials...
Initialising default mesh containers...
Graphics subsystem initialized
Sound subsystem initialized
Event subsystem initialized
Detected game "Neverwinter Nights 2"
[  0%] Loading user game config
[  5%] Declare string encodings
[ 10%] Setting base directory
[ 15%] Adding extra archive directories
[ 20%] Loading main resource files
Shutting down
ERROR: No such archive file "2da.zip"
Cleaning up shaders...

DrMcCoy:
> The configure step of this CMake PR takes around two second when the original autotools checks take almost a minute on my brand new SSD

Yes, that's for the most part, down on boost.m4. Because it finds all different ways the boost libraries might be named and want to be linked with. And checks for other libraries.

There's no way I can use the CMake thing for my cross-compiling MinGW, with statically linking boost, for example.

DrMcCoy:
> 2DA.zip is actually in Data directory

Yes, and xoreos should be able to search through that. It does register the data directory as directory for ZIP resources files (see src/engines/nwn2/nwn2.cpp:201) At least it works here both on Linux, and with wine with the MinGW build...

There might be an issue with the filepath stuff. :/

DrMcCoy:
> At least it works here both on Linux, and with wine with the MinGW build...

Or maybe not. Just tried with a MinGW build in wine and it doesn't work anymore. It seems I recently broke that. My bad. :/

I'll fix it.

cc9cii:
> There's no way I can use the CMake thing for my cross-compiling MinGW, with statically linking boost, for example.

That should be possible with CMake (perhaps it is cross compiling that causes trouble?).

In any case, why can't autotools coexist with CMake?  (I guess one reason might be people like me update exclusively on CMake and autotools miss out on updates/fixes ;-)  But I don't know enough to update autotools so nothing is lost there ;-) :-D

Seriously, though, MSVC needs a bunch of additional stuff to get working properly and CMake just supports them better IMHO.

DrMcCoy:
Okay, with 16e9823965bc1230afe3ec73a84846b884ab34ce, the problem of Windows not finding the archives should be fixed. It's a bit of a hacky way to use absolute() for, though, which is why it broke on Windows in the first place... I might need to rethink the whole shebang, again...

cc9cii:
Confirming that your fix works.  Thanks for that.  Loads an internal cell but doesn't do much else (was there meant to be a menu that responds to mouse and/or keyboard?)

DrMcCoy:
>  Loads an internal cell but doesn't do much else

Yes, for NWN2, that's about it. You can open a debug console with Ctrl+D, though, and load other areas and modules.

> was there meant to be a menu that responds to mouse and/or keyboard

Not for NWN2. We have a partial main menu for NWN, KotOR and KotOR2, and a partial ingame menu for NWN.

In NWN2, the menu definition is done by a XML files. Very, very broken and non-standard-compliant XML files. We need a custom parser for them first, because a stock XML library will throw errors left and right.

berenm:
> There's no way I can use the CMake thing for my cross-compiling MinGW, with statically linking boost, for example.

> That should be possible with CMake (perhaps it is cross compiling that causes trouble?).

This is definitely possible, and quite easy, to cross compile with CMake. There are some platform toolchain descriptions, included in this PR, that can be used for cross-compilation.

Regarding Boost static link, it's probably a option to add. ``cmake --help-module FindBoost`` tells me that it could be controlled by Boost_USE_STATIC_LIBS option.

> In any case, why can't autotools coexist with CMake? (I guess one reason might be people like me update exclusively on CMake and autotools miss out on updates/fixes ;-) But I don't know enough to update autotools so nothing is lost there ;-) :-D

This is how I meant this pull request to work. It parses the Automake descriptions to find targets, and as long as these files stay simple, CMake will not require duplicating the target descriptions. The only duplication is about library checks, and config.h generation, which is minimal.

The LLVM project for example, although they use a more complicated setup, have the same CMake / Autotools hybrid build system.

DrMcCoy:
> Yes, for NWN2, that's about it

Oh, and you can "fly" through NWN2 areas in a kind of spectator mode. WASD for movement; middle mouse button + mouse move = camera rotate; Q, E, page up, page down also rotate the camera.

.gitignore
```diff
@@ -46,5 +46,6 @@ INSTALL
 /gitstamp/gitstamp
 # Unix binary
 /src/xoreos
-# Window binary
```
DrMcCoy: This changeset doesn't belong here.

CMakeLists.txt
```diff
@@ -0,0 +1,228 @@
...
+  set(GLEW_LIBRARIES    glew)
+  set(GLEW_INCLUDE_DIRS glew)
+
+  parse_configure(configure.ac glew)
+endif()
+include_directories(${GLEW_INCLUDE_DIRS})
+list(APPEND XOREOS_LIBRARIES ${GLEW_LIBRARIES})
+
+
+find_package(Lua51)
```
DrMcCoy: I'm not sure we actually *ever* want a system Lua instead of our own. We have to support the Witcher scripts, which are compiled, and therefore pinned to a specific Lua version. And we will probably need to hack the loader anyway, to do icky stuff to the bytecode (endianess changes, for one thing).

CMakeLists.txt
```diff
@@ -0,0 +1,228 @@
...
+
+
+# -------------------------------------------------------------------------
+# packaging
+# see: http://www.vtk.org/Wiki/CMake:CPackPackageGenerators
+set(CPACK_GENERATOR "TGZ")
+
+set(CPACK_PACKAGE_NAME    ${PROJECT_NAME})
+set(CPACK_PACKAGE_VERSION ${xoreos_VERSION})
+set(CPACK_PACKAGE_CONTACT "Sven Hesse <drmccoy@drmccoy.de>")
```
DrMcCoy: I think that contact should rather by the mailing list at xoreos-devel@xoreos.org, no?

CMakeLists.txt
```diff
@@ -0,0 +1,228 @@
...
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Xoreos Engine: A new implementation of BioWare's Aurora engine.")
+set(CPACK_PACKAGE_DESCRIPTION "Xoreos Engine: A new implementation of BioWare's Aurora engine.")
+set(CPACK_PACKAGE_VENDOR "Xoreos Project")
```
DrMcCoy: If it doesn't stand out too much, I'd rather like "xoreos" to always be all-lowercase. The only place were we don't do that is the Mac Bundle name thing, because I've been told it looks bad otherwise.

CMakeLists.txt
```diff
@@ -0,0 +1,228 @@
...
+  # MacOSX bundle, unchecked packaging
+  # list(APPEND CPACK_GENERATOR "BUNDLE")
+  # set(CPACK_BUNDLE_NAME "Xoreos")
+  # set(CPACK_BUNDLE_PLIST dists/macosx/Info.plist)
+
+elseif(CMAKE_HOST_UNIX)
+  find_program(DPKG_CMD dpkg)
+  if(DPKG_CMD)
+    list(APPEND CPACK_GENERATOR "DEB")
+    set(CPACK_DEBIAN_PACKAGE_DEPENDS "libc6, libgcc1, libz1, libboost-all-dev, libsdl1.2-dev, libopenal1, libgl1, libfreetype6, libglew1.7, libmad0, libfaad2, libvorbis0a, libxvidcore4")
```
DrMcCoy: Make that libsdl2-dev instead of libsdl1.2-dev

CMakeLists.txt
berenm: Well, I'm not sure if packaging even has to be supported by the CMake toolchain. It was to be on-par with autotools but maybe not useful.

CMakeLists.txt
```diff
@@ -0,0 +1,228 @@
...
+
+# -------------------------------------------------------------------------
+# subfolders where built binaries and libraries will be located, relative to the build folder
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+
+
+# -------------------------------------------------------------------------
+# find the required libraries
```
DrMcCoy: You're missing liblzma, which is used for BZF archives (which are found in the iOS version of KotOR)

CMakeLists.txt
DrMcCoy: True. Hell, I haven't really thought much about proper packing anyway. That's not something we're ready for yet. :P

CMakeLists.txt
```diff
@@ -0,0 +1,228 @@
...
+
+# endianess detection, could be replaced by including Boost.Config 
+include(TestBigEndian)
+test_big_endian(XOREOS_BIG_ENDIAN)
+if(XOREOS_BIG_ENDIAN)
+  add_definitions(-DXOREOS_BIG_ENDIAN=1)
+else()
+  add_definitions(-DXOREOS_LITTLE_ENDIAN=1)
+endif()
+
```
DrMcCoy: I don't suppose there's a way to define XOREOS_REV, XOREOS_REVDESC and XOREOS_BUILDDATE (the gitstamp / common/Makefile.am hackery)? Those need to trigger a recompile for version.o on git changes, after all...

CMakeLists.txt
```diff
@@ -0,0 +1,228 @@
...
+# -------------------------------------------------------------------------
+# find the required libraries
+set(XOREOS_LIBRARIES "")
+
+find_package(ZLIB REQUIRED)
+include_directories(${ZLIB_INCLUDE_DIRS})
+list(APPEND XOREOS_LIBRARIES ${ZLIB_LIBRARIES})
+
+find_package(Boost COMPONENTS filesystem system regex date_time REQUIRED)
```
DrMcCoy: You're missing boost.atomic there. (I take it the components are only for non-header-only libraries?)

DrMcCoy:
Okay, since this seems to be the only way to support MSVC and stop turning MSVC-only people away, I'm considering accepting this changes as a kind of unsupported, alternative "use this if you absolutely have to, but don't yell at us if it's not working" kind of thing.

And you, @berenm, will have to keep it up-to-date with library dependency and config.h changes. I don't expect those to come often, of course. Nor do I expect to complicate our Makefile.am files, so those should keep being easily parseable, I guess.

CMakeLists.txt
berenm: Should be doable, I'll have a look.

berenm:
Sounds perfectly fine, please assign any related issue to me and I will take care of it.

CMakeLists.txt
turol: 32- and 64-bit Lua bytecodes are not compatible either so you'll have to find a workaround for that too.

DrMcCoy:
Okay, thanks. :)

To recap, please change these things:
- remove that .gitignore changeset
- add checks for liblzma and boost.atomic
- never use the system's Lua library
- either remove the packaging or change the mail contact and libsdl1.2-dev to libsdl2-dev
- change "Xoreos" to all-lower "xoreos" where appropriate

Nice to haves:
- the gitstamp and date/time defines

CMakeLists.txt
DrMcCoy: Yeah, I'm already dreading that...

berenm:
Alright, I did the changes.

I removed packaging stuff, as there was no CMake install directives to install things into the target tree, so the packages were empty. It can be added back later on, if needed.

Git version string should now be added as preprocessor definition when compiling ``src/common/version.cpp``, but the dirty flag is not 100% accurate. For some reason, when creating Makefiles, make feels that it has to rebuild the whole common library's sources although the preproc macros are for version.cpp only... When using ninja, everything works as expected. :neutral_face:

berenm:
XOREOS_REV was not set to the right portion of XOREOS_REVDESC, it's fixed now.

DrMcCoy:
Okay, merged it. Thanks!

PR #67, opened 2015-01-26
GRAPHICS: Add shader handling foundation

mirv-sillyfish:
This adds files used as the foundation of shader handling. No
initialisation is called, and no shaders are actually yet used
anywhere.


DrMcCoy:
Just some few minor things remaining.

I'd actually fix them myself and merge the thing (*), if it weren't for the ShaderInstance namespace. Is that also supposed to go into Graphics::Shader?

(*) I think I annoyed you enough with those fussy details; sorry. You probably get what you should look out for by now. :)

mirv-sillyfish:
There is a reason for ShaderInstance not going into Graphics::Shader, but it's not apparent right now. On thinking further, maybe it won't be apparent for a while, if at all, so into Graphics::Shader it goes!

DrMcCoy:
Merged it, thanks! :)

PR #73, opened 2015-02-19
GRAPHICS: Use new[] instead of malloc()

mirv-sillyfish:
Index and vertex buffers can safely use new[] instead of malloc
and not run into alignment issues.


DrMcCoy:
Merged, thanks :)

PR #77, opened 2015-03-02
GRAPHICS: Mesh bind and render separation

mirv-sillyfish:
Mesh rendering can be performed separate to actual binding. This
allows multiple render calls without rebinding the same data, which
would otherwise be a waste of GL calls.


DrMcCoy:
Can you remove code duplication by having renderImmediate() call the three separate functions, or is there a reason that won't work?

mirv-sillyfish:
Just thought to cut down on the checks for GL2.x or 3.x - but otherwise, there's no reason renderImmediate can't call the functions.

PR #86, opened 2015-05-02
MSVC2013

cc9cii:
These are minimal amount of changes required for building on Windows with CMake and MSVC.  I've refrained from making further changes in case this PR results in many updates before acceptance.  (for example the following line seems like an error to me, but perhaps \\( is valid and **required** on gcc?  I simply don't know.)
```
  console->printf("Entering area \"%s\" \(\"%s\")", _currentArea->getResRef().c_str(),
```

I am also unsure on the project policies regarding platform specific warnings, i.e. are they meant to be guarded with #ifdef/#endif


DrMcCoy:
First of all, thanks!

> but perhaps \( is valid and **required** on gcc?

Ah, nope, that's me messing up. :)

> I am also unsure on the project policies regarding platform specific warnings

Err, what exactly do you mean? The #pragmas you #ifdef'd out in 2579d20e6d939aee19d359f6ef709261bcdc83f2?  Or something else?

DrMcCoy:
Okay, I went ahead and added that _Pragma() macro, commited the include and stroull fixes, and cherry-picked 281bde794c0f5f8d0b3aad8cb638fd1bdbaafc36. :)

I'll wait for @berenm's thoughts on the CMake stuff, because I really have no clue there.

How do you want to be credited in the Contributors list in the AUTHORS file? ```cc9cii <cc9c@iinet.net.au>```? Or a different nickname and/or a real name to go with that?

cc9cii:
> >   I am also unsure on the project policies regarding platform specific warnings
>
> Err, what exactly do you mean? The #pragmas you #ifdef'd out in 2579d20? Or something else?

For example, I get a warnings on lines like this, apparently a performance issue on MSVC:
```c++
bool invertRestrict = row.getInt("InvertRestrict");
```
Does that now become:
```c++
#ifndef _MSC_VER
bool invertRestrict = row.getInt("InvertRestrict");
#else
bool invertRestrict = row.getInt("InvertRestrict") != 0;
#endif
```

cc9cii:
> Any reason that's not \#include \<cstdlib\> for consistency's sake?

Yes, it can be \#include \<cstdlib\>, I should have thought about that ;-)

> How do you want to be credited in the Contributors list in the AUTHORS file? 

cc9cii is fine.

DrMcCoy:
Oh. No, please don't wrap those changes into ifdefs.

Technically, that code there does warrant a warning, because the conversion from int to bool doesn't preserve all possible values. gcc and clang don't warn there (and I can't even seem to coax them into generating warnings there). I didn't know that MSVC does warn there.

I would very much appreciate if you could fix these all, i.e. change them all to use != 0 (without #ifdef'ing), and submit a pull request for them, if you don't mind that extra work.

If there's other similar issues, where the code would actually be more correct if changed, I would love having them fixed as well.

> cc9cii is fine.

Okay :)

cc9cii:
> If there's other similar issues, where the code would actually be more correct if changed, I would love having them fixed as well.


There are literally thousands of these warnings (loss of precision from double to float, size_t to int, etc, etc) that occur due to different platforms having different size of int or MSVC being just more pedantic.  Fixing those will result in ugly casts (either c style or static_cast) so you may have second thoughts on that ;-)

DrMcCoy:
Hmm, right, I hadn't thought about the myriads of double->float (especially since I tend to use double literals, even when assigning to float) and size_t->uint conversions...

cc9cii:
Apologies for the delay.  I've made changes as per berenm's comments in commit  4a43611 a couple of weeks ago.  I also added a commit to suppress macro redefinition warnings.

DrMcCoy:
Oooh, I didn't see that you edited your reply, and completely forgot about this PR. I'm sorry. :/
Please, in the future, just add a new reply, so that I get an email as a reminder.

Yes, please resubmit everything nice and clean, if that's no trouble for you.

cc9cii:
Ok, I'll push a fresh PR

DrMcCoy:
What's the status of the fresh PR? :)

Or did I miss something again? If so, my apologies.

cc9cii:
So sorry, got distracted while preparing a fresh PR (had to go back to all my notes to see what I did before).  Question: is keeping the repository clean critically important?  Software development is often "messy" and mistakes are made, corrected, etc.  It seems that it is a little artificial to have submit a new PR each time, just a thought.  Anyway, you may close this PR since I'll be raising a new one.  Regards,

DrMcCoy:
> So sorry, got distracted

Sure, no problem. :)

> It seems that it is a little artificial to have submit a new PR each time

Well, you can also rebase this PR with the changes (but please add a comment when you did, so that I get notified by email). But yes, even when rebased, it should be rebased into a clean state. When I merge/rebase this onto the master branch, I do want it to be clean.

I.e. I don't care about the PR as such being clean, but the commits in them. If that makes sense.

> Software development is often "messy"

Sure, but that doesn't mean you have to mess it up on purpose, especially when you can very easily clean it up at this stage.

When I try out or implement new features, I do that in a local branch too, and then when I have it down, I rebase the branch to hell and back until it's clean and logical, before I rebase it onto master and push it. I don't want the history to accurately display my dozens of commits only stating "WIP" in the commit message; I want the history to be relatively logical and easy to look through in the future.

So, please, keep the PR tidy. Thanks. :)

cc9cii:
New PR submitted, closing.

PR #87, opened 2015-05-03
Warnings cleanup

cc9cii:
I am not sure whether you preferred this style (commit 320b772) or instead add IGNORE_UNUSED_VARIABLES macro to these files as well.

The removal suppression of the warnings on MSVC build for all files will be submitted once the CMakefiles are sorted out.

> I would very much appreciate if you could fix these all, i.e. change them all to use != 0 (without #ifdef'ing), and submit a pull request for them, if you don't mind that extra work.

Commit  8acfab5

> if the disabling of these warning is in any way applicable to MSVC, could you maybe add a #define for IGNORE_UNUSED_VARIABLES for MSVC in there?

Commit 9c9dffc

### DIFF
--- a/src/aurora/erffile.cpp
+++ b/src/aurora/erffile.cpp
@@ -86,7 +86,7 @@ void ERFFile::load() {
 
 			if (!_noResources)
 				readResources(erf, erfHeader);
-		} catch (Common::Exception &e) {
+		} catch (Common::Exception &/*e*/) {
 			delete[] erfHeader.stringTable;
 			throw;
 		}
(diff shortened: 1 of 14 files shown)


DrMcCoy:
Thanks! :)

Actually, for 320b772eb9d66a89c804f8f445ea8c590156d6ae, I'd rather like to use [our UNUSED() macro](https://github.com/cc9cii/xoreos/blob/master/src/common/system.h#L115), just to make it explicit what's happening there.

Of course, when compiling C++ code, it could/should leave the name out completely. Dunno why I hadn't thought about that previously. I commited a change to do just that with 1c379fea6452d6005873539d640344f7d805b4bf, and modified your 320b772eb9d66a89c804f8f445ea8c590156d6ae to use UNUSED() into 55ec8f7a5481b469376b17abdc4126bbd85c0024.

I cherry-picked the other two commit as e5e6d55e52e520c790f84828049ef7aede8ca3fc and 24e61d8b49de546e433bc7952888393270ebca9a. :)

PR #89, opened 2015-05-12
GRAPHICS: Load jade mesh bounding shape info.

logzero:
Verified with N_Zu_ mesh from players.mod


DrMcCoy:
Oh, neat! :)

Merged, thanks.

PR #94, opened 2015-07-31
Compile msvc

cc9cii:


DrMcCoy:
Merged it, thank you very much! :)

I took the liberty of shortening the first line of the commit messages a bit, though. They first line really shouldn't exceed 70 characters (and should not include a full stop at the end).

Can you apply the CMake fixes to [xoreos-tools](https://github.com/xoreos/xoreos-tools) too, if it's not too much of a bother? And actually check if that works on Windows? I basically just copied @berenm's xoreos CMake files over there, and removed the library dependency stuff that's not needed for xoreos-tools. It works on GNU/Linux here, but I have no idea if it works on Windows with MSVC.

PR #101, opened 2015-11-02
BUILD: Keep MSVC compiler switches unique to MSVC

kevL:
when using CMake.


DrMcCoy:
Rebased it onto the three commits I did a few minutes prior and merged it as df0c75f24da2d69a7dde194561b91a1214da5203. Thanks! :)

PR #102, opened 2015-11-03
sloooow slow going

kevL:
never used CMake before ...
I'm about to revamp the strategy/system on my end, just to get things working, and worry about integration later.

- any words of advice? regarding specifying dependencies (headers & libs) w/ CMake, and how this may or may not impinge on static vs. dynamic linking. Several CMake modules -- FindZLIB.cmake, eg -- appear to be missing. Or is that just me. I can go into some detail but am keeping things brief here

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,12 +40,13 @@ endif()
 
 # platform specific warning settings
 if(CMAKE_HOST_WIN32)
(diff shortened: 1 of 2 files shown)


kevL:
(darn comments appearing as commit-description ..) will fix it up.

kevL:
now it's screwed :\
Can't change title. Trying to re-PR just opens this again.

PR #103, opened 2015-11-03
BUILD: Fix messages for SDL2

kevL:
when using the FindSDL2 module in CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,12 +40,13 @@ endif()
 
 # platform specific warning settings
 if(CMAKE_HOST_WIN32)
-  if(CMAKE_CXX_FLAGS MATCHES "/W[0-4]")
-    string(REGEX REPLACE "/W[0-4]" "/W4" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-  else()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4")
-  endif()
   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+    if(CMAKE_CXX_FLAGS MATCHES "/W[0-4]")
+      string(REGEX REPLACE "/W[0-4]" "/W4" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    else()
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4")
+    endif()
+
     set(WARNINGS_DISABLE
       # ignore
       4250 # 'class1' : inherits 'class2::member' via dominance
--- a/cmake/FindSDL2.cmake
+++ b/cmake/FindSDL2.cmake
@@ -2,16 +2,16 @@
 # Once done this will define
 #
 # SDL2_INCLUDE_DIRS - where to find SDL.h, etc.
-# SDL2_LIBRARIES - List of libraries when using libmad.
-# SDL2_FOUND - True if libmad found.
+# SDL2_LIBRARIES - List of libraries when using SDL2.
+# SDL2_FOUND - True if SDL2 found.
 
 if(WIN32)
   find_path(SDL2_INCLUDE_DIR SDL.h $ENV{PROGRAMFILES}/SDL2/include DOC "The directory where SDL.h resides")
-  find_library(SDL2_LIBRARY NAMES SDL2 PATHS $ENV{PROGRAMFILES}/SDL2/lib DOC "The libmad library")
+  find_library(SDL2_LIBRARY NAMES SDL2 PATHS $ENV{PROGRAMFILES}/SDL2/lib DOC "The SDL2 library")
 
 else(WIN32)
   find_path(SDL2_INCLUDE_DIR SDL.h /usr/include/SDL2 DOC "The directory where SDL.h resides")
-  find_library(SDL2_LIBRARY NAMES SDL2 DOC "The libmad library")
+  find_library(SDL2_LIBRARY NAMES SDL2 DOC "The SDL2 library")
 
 endif(WIN32)
 
@@ -30,7 +30,7 @@ mark_as_advanced(SDL2_LIBRARY)
 mark_as_advanced(SDL2_FOUND)
 
 if(NOT SDL2_FOUND)
-  set(SDL2_DIR_MESSAGE "libmad was not found. Make sure SDL2_LIBRARY and SDL2_INCLUDE_DIR are set.")
+  set(SDL2_DIR_MESSAGE "SDL2 was not found. Make sure SDL2_LIBRARY and SDL2_INCLUDE_DIR are set.")
   if(NOT SDL2_FIND_QUIETLY)
     message(STATUS "${SDL2_DIR_MESSAGE}")
   else(NOT SDL2_FIND_QUIETLY)


kevL:
k, that's better. (wondering why 'it' didn't just do this the first time)

kevL:
and ... why does it seem like this is adding 3 commits
o.O

when at least one of them "MSVC compiler switches" should have already been taken care of.

DrMcCoy:
> and ... why does it seem like this is adding 3 commits

Because you have a merge commit in your local checkout now. I had added three commits on my own before cherry-picking your commit ontop.

This left the "official" repository history different from your own own.
The xoreos/xoreos repository history looked like this:

- Your commit
- My commit 3
- My commit 2
- My commit 1
- Old commits

While your looked like this:

- Your commit
- Old commits

When you then pulled your origin/master into your tree, git created a merge commit to bind these two histories together.

You can reset your checkout back to the offical history by doing a "git reset --hard origin/master" after fetching (which is done either explictly through "git fetch origin master", by doing a "git remote update", or as part of the git pull that created your merge commit to begin with). If you have already new commits in your tree and then the official repository suddenly also gets different new ones, you can do a "git rebase origin/master" to rebase your changes ontop of the other new changes (you need to fetch first, though, either explictly or through a "git remote update"; not git pull, because that will automatically try to merge). If that fails somehow, maybe because a merge is already there, you can create a tempary branch so that your work is not lost (*), then go back to your master, reset it, and cherry-pick your changes back onto the master branch.

(*) Technically, the work isn't lost even without a branch, as you can always see your personal work history in "git reflog" but inaccessible commits have an expiration date (default 2 weeks, I think) and might be purged when git does cleanups

Having the merge commit in your history doesn't exactly matter in terms of "will the code still work", but it makes the history a bit more messier to work with. Comparing differences between the official repository and your changes might include older differences you have to manually ignore, for example. Pull requests will also include old changes, as you can see here, which means I'll have to manually cherry-pick the real changes.

I hope that makes it a bit more clearer, but I admit that there are messy git internal that can be a bit daunting for newcomers to git. You might want to read the [git book](https://git-scm.com/book/en/v2), but, again, I admit it might feel a bit overwhelming. Sorry for that.

DrMcCoy:
As for your comment on the close PR #102:

> regarding specifying dependencies (headers & libs) w/ CMake, and how this may or may not impinge on static vs. dynamic linking

Frankly, I have no idea. I don't use CMake, I use the autotools buildsystem on my end (which doesn't work together with Visual Studio on Windows, and probably won't work together with netbeans either). I *think* CMake defaults to dynamic linking, though.

> Several CMake modules -- FindZLIB.cmake, eg -- appear to be missing

Those *should* come with the default CMake installation, I thought.

You might want to ask @berenm (if you excuse me tagging you here because of this :P) about CMake specific things. I personally know barely anything about CMake, sorry.

DrMcCoy:
In either case, I cherry-picked and pushed your SDL2 message fix, thanks! :)

I'll close this PR, but the comments will stay open. Feel free to ask me about git, and I'll try to clarify as well as I'm able to.

kevL:
ok, thanks for the willingness to help. On my end, I use GitExtensions here; and i can sorta see in its schematic how 'master branch' sweeps through (includes) the MSVC-compiler commit before getting merged back into master-proper. So, effectively 'master' had two parallel branches i Guess (yours/mine).

- will try to keep an eye out for stuff like that. +reset hard

Rebase of course always comes with that warning: Do it wrong and we'll hunt you down ....


Anyway, regarding CMake. I tested two strategies that work for me to get past errors & apparently missing Find* modules. One is to create subdirs of /xoreos -> deps/include and deps/lib

then point the modules there. The other is to pass in the include_dir and perhaps library (...) on the cmake commandline. But at this point it's all just a dirty, hacky way to get cmake to stop erroring out and let me get on to where the uh, real errors are.

btw, I'm using Netbeans on WinXP 32-bit, and intend to use CMake->make->GNU compiler/linker w/ the MSYS subsystem.

And to think they used to do this on a single commandline. yeesh ...

berenm:
This CMake issue look very weird to me. If you could paste the CMake error messages, it would probably be clearer.

DrMcCoy:
> Rebase of course always comes with that warning: Do it wrong and we'll hunt you down ....

Sure, but even if you break your tree, you can always go back to an earlier state. With "git reflog", you can see your old states, with the revision ID in front. Give that revision ID to git reset --hard, and you're back to how the repository was then.

kevL:
hey beren,

Here's the console output (cmd.exe, although Bash shell is the same)
running CMake 3.3.2
against the master repository.

default run: http://pastebin.com/5pLA41mp

MSYS run: http://pastebin.com/hKkK4DtX
cache vars: http://pastebin.com/aBXQK8PY


The output and error logs just stop after endianness is checked. But i can get them to you if you like,



Ps. my GIT probls appear resolved for now.

berenm:
Alright, it's what I thought:

CMake does not complain about a missing `FindZLIB.cmake`, but about not being able to find the zlib library on your setup. When configuring the project, it will try to look for zlib at predefined locations (on Linux it looks for the headers in `/usr/include`, and for the libraries in `/usr/lib/<arch-triplet>`; on Windows, it's probably trying some path like `C:/Program Files/Zlib`).

Apparently it isn't able to find zlib at this predefined location on your system, and it complains about it. If you have installed zlib and the other dependencies in a folder different that the expected one, you will have to specify these folders using the corresponding CMake variables (as indicated: `ZLIB_LIBRARY` and `ZLIB_INCLUDE_DIR`, and equivalent variables for other dependencies).

It's not specific to CMake btw, you would have the same issue with any other build configuration tools if you don't have the dependencies where it expects them to be.

kevL:
That makes some sense.

btw, pls check line 13 in 'FindSDL2.cmake'

else(WIN32)
  find_path(SDL2_INCLUDE_DIR SDL.h /usr/include/SDL2 DOC "The directory where SDL.h resides")


Should "/usr/include/SDL2" really be there?

berenm:
Probably, that's the expected location I was talking about, what's wrong with it?

kevL:
seems unlikely to be on a win32 machine ...
+ none of the other custom Find* modules have a path string there.

berenm:
It's not in the WIN32 part of the if. The WIN32 part has `$ENV{PROGRAMFILES}/SDL2/include` as hint path. CMake language has this nasty syntax where it was required to have the same parameters to all of the if / else / endif control statements. 

NB: It's not required anymore, but as the FindSDL2.cmake is a dumb copy-paste from an old CMake module, it's written with the oldstyle syntax.

kevL:
ok, you're da boss.

So... is there a convenient hook i can use for my currently unfound deps? Or do I have to deal with commandline input?

berenm:
You have a few possible solutions:
0. Install dependencies in the expected location.
1. Use -D<variable>=<path> flags on CMake command-line for each variable it complains about with the correct path.
2. Use cmake-gui to manually add the missing paths in such variables.
3. Install all the dependencies headers and libraries in the same folders and fill the `CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH` with the folders where headers (resp. libraries) should be found.
4. Install all the dependencies in a single folder, with an `include` and `lib` subfolders containing headers (resp. libraries), and provide it to CMake in the `CMAKE_PREFIX_PATH` variable.

I would recommend 0., and I know by experience that 1. and 2. work fine.
3. and 4. should work as documented in https://cmake.org/cmake/help/v3.0/command/find_path.html and https://cmake.org/cmake/help/v3.0/command/find_library.html, but no guarantees.

kevL:
thanks. will try some stuff out ..

maybe 1. for headers
and 4. for libraries

or, admittedly, insert a private/personal hook to a custom module that does it.

PR #105, opened 2015-11-19
COMMON: Avoid endianness redefinition

kevL:
The CMake front-end defines XOREOS_LITTLE_ENDIAN already, if
applicable. This prevents warnings about macro redefinitions.

### DIFF
--- a/src/common/system.h
+++ b/src/common/system.h
@@ -89,7 +89,9 @@
 
 #elif defined(__MINGW32__)
 
-	#define XOREOS_LITTLE_ENDIAN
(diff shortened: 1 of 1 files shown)


DrMcCoy:
Good catch, thanks!

I extended it a bit to also cover a theoretical redefinition of XOREOS_BIG_ENDIAN, also applied it to the MSVC and default UNIX case, and commited it as 4b129c68d55d6b25fa25ae52fed38637153636f6. :)

kevL:
ok :)

PR #110, opened 2015-12-04
Rewrite manpage to use semantic -mdoc macros.

bentley:
This updates the macros in the manpage to the semantic -mdoc macros.

-mdoc is a drop‐in replacement for -man. It is supported out of the box by the default manpage formatters in Linux, OpenBSD, FreeBSD, Mac OS X, Illumos, and Minix. No change is necessary when installing the manpage.

-mdoc macros are semantic, and aside from groff (what most Linux distributions use to format manpages) there are several great tools for such manuals, especially the mandoc suite, which supports advanced searching, conversion to semantic HTML, and linting.

### DIFF
--- a/man/xoreos.6
+++ b/man/xoreos.6
@@ -1,95 +1,89 @@
-.de URL
-\\$2 \(laURL: \\$1 \(ra\\$3
-..
-.if \n[.g] .mso www.tmac
-
-.hw xoreos
-.hw $XDG\_CONFIG\_HOME-/-xoreos-/-xoreos.conf
-.hw $XDG\_CONFIG\_HOME
-.hw $HOME-/-.config/
-.hw $HOME-/-Library-/-Preferences-/-xoreos-/-xoreos.conf
-.hw xoreos.conf
-.hw $APPDATA
-.hw $USERPROFILE
-
-.TH xoreos 6 2015-07-24 xoreos
-.SH NAME
-xoreos - A reimplementation of BioWare's Aurora engine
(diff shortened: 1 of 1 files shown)


DrMcCoy:
Oh, I didn't know about that! Thanks :)

I'll look at this more closely tomorrow, after I had some sleep and regained the ability to think clearly. :P

DrMcCoy:
Except for these two minor issues, I really quite like it. :)

DrMcCoy:
Oh, and the commit message is wrong. It should be `DOCS: Rewrite manpage to use semantic -mdoc macros`, i.e. a preceeding "DOCS: " to specify the section it concerns, and no trailing period.

bentley:
Add the Ns, and fixed the commit message.

There are several good resources for writing manpages. One is [manpages.bsd.lv](http://manpages.bsd.lv/), which has a tutorial and a really interesting piece on the history of manpage formatters. Another is mandoc’s [mdoc(7) manpage](http://mdocml.bsd.lv/man/mdoc.7.html#x5c264464), which covers dates:

“The month is the full English month name, the day is an optionally zero-padded numeral, and the year is the full four-digit year. Other arguments are not portable…” So that’s why I went with that date format.

DrMcCoy:
> Other arguments are not portable

Ah, okay, then I guess my preference for the ISO date format has to take a bow.

In either case, I pushed it as 150bc282bc85e88429258fb0ce6becdbe97f3024. Thanks! :)

PR #111, opened 2015-12-05
MAIN: Fail before probing if path is a file

ChaoticMind:
When trying to launch xoreos for the first time, I accidentally pointed the path (in -p) to nwmain.exe instead of the parent folder. The usage function is clear that I should point to the folder, but I didn't immediately notice it and I was confused because of the strange unrelated opengl/audio errors that I got instead.

Now we fail early with a clear error message instead.

### DIFF
--- a/src/xoreos.cpp
+++ b/src/xoreos.cpp
@@ -147,8 +147,11 @@ int main(int argc, char **argv) {
 		error("Invalid path \"%s\"", dirArg.c_str());
 	}
 
-	if (!Common::FilePath::isDirectory(baseDir) && !Common::FilePath::isRegularFile(baseDir))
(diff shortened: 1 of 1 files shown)


DrMcCoy:
Maybe, though, when probing on the file fails, we should automatically retry on the directory containing the file?

DrMcCoy:
Also, a probing failure right now raises an exception, just like all other failures in xoreos, and is handled the same way, by printing the exception and quitting (without cleaning up, too). It would be nice if that could be done cleaner, somehow, so that a probing failure is a non-lethal, or something, failure that provides the error and still cleans up.

In either case, closing this pull request.

Issue #99, opened 2015-09-15
Make keyboard states recursive/re-entrant


DrMcCoy:
This is a fairly simple and low-priority feature request that would be perfect for newcomer to tackle; see http://www.firsttimersonly.com/. If you are already experienced in programming and/or contributing to free/libre open source software projects, please have a look at [our general TODO list](https://wiki.xoreos.org/index.php?title=TODO) instead.

---

[```EventsManager::enableKeyRepeat()```](https://github.com/xoreos/xoreos/blob/master/src/events/events.cpp#L260) is used to switch on repeating keys. I.e. when it's enabled, holding down a key will fire several KeyDown events. When it's disabled, holding down a key will only fire one single KeyDown event. [```EventsManager::enableTextInput()```](https://github.com/xoreos/xoreos/blob/master/src/events/events.cpp#L264) is used to signal that the input of a whole textual string is requested instead of single key presses for game interaction. For example, entering a name in a character creator vs. pressing the "w" key to move forward. This is especially crucial for a potential future mobile port, where ```enableTextInput()``` might drop-down an on-screen keyboard.

Currently, these features are enabled/disabled with a simple on/off method. You call ```EventMan.enableKeyRepeat(true)``` to enable the feature, and ```EventMan.enableKeyRepeat(false)``` to disable the feature. Even if you call ```EventMan.enableKeyRepeat(true)``` twice in a row, only a single ```EventMan.enableKeyRepeat(false)``` will disable the feature.

However, we have potential situations were we are already in a state were the feature is enabled, but then something happens that would also enable the state. For example, we are in a character creator, inside the GUI field for a character name, which expects a textual input. Then we press Ctrl+D to drop down the debug console, which also expects a textual input. When we then close the console, it would disable the textual input feature, leaving us in a character creator name field without enabled textual input.

To implement this recursive behaviour, we would have a value count up every time the feature is enabled, and down every time it is disabled. If the integer is > 0, we really enable the feature, and only disable it when the count reaches 0 again.

---

[...]

PR #113, opened 2015-12-09
EVENTS: Add counter to EventsManager::keyRepeat

Koffeinfriedhof:
Make keyboard states recursive/re-entrant

https://github.com/xoreos/xoreos/issues/99

### DIFF
--- a/src/events/events.cpp
+++ b/src/events/events.cpp
@@ -54,7 +54,8 @@ const EventsManager::RequestHandler EventsManager::_requestHandler[kITCEventMAX]
 
 
 EventsManager::EventsManager() : _ready(false), _quitRequested(false), _doQuit(false),
-	_fatalError(false), _queueSize(0), _fullQueue(false), _repeat(false), _repeatCounter(0) {
+	_fatalError(false), _queueSize(0), _fullQueue(false), _repeat(false), _repeatCounter(0),
+	_textInputCounter(0) {
 
 }
 
@@ -77,6 +78,8 @@ void EventsManager::init() {
 
 	std::srand(getTimestamp());
 
+	// Forcing enableTextInput to be disabled requires _textInputCounter = 1 to not underrun the counter.
+	_textInputCounter = 1;
 	enableTextInput(false);
 
 	_repeatCounter = 0;
@@ -109,6 +112,8 @@ void EventsManager::reset() {
 	initJoysticks();
 
 	_repeatCounter = 0;
+
+	_textInputCounter = 0;
 }
 
 bool EventsManager::ready() const {
@@ -274,10 +279,18 @@ void EventsManager::enableKeyRepeat(bool repeat) {
 }
 
 void EventsManager::enableTextInput(bool textInput) {
-	if (textInput)
-		SDL_StartTextInput();
-	else
-		SDL_StopTextInput();
+	if (!textInput && _textInputCounter == 0)
+		throw Common::Exception("EventsManager::enableTextInput(): Counter underrun");
+
+	if (textInput) {
+		if(_textInputCounter == 0)
+			SDL_StartTextInput();
+		_textInputCounter++;
+	} else {
+		if (_textInputCounter == 1)
+			SDL_StopTextInput();
+		_textInputCounter--;
+	}
 }
 
 Common::UString EventsManager::getTextInput(const Event &event) {
--- a/src/events/events.cpp
+++ b/src/events/events.cpp
@@ -54,7 +54,7 @@ const EventsManager::RequestHandler EventsManager::_requestHandler[kITCEventMAX]
 
 
 EventsManager::EventsManager() : _ready(false), _quitRequested(false), _doQuit(false),
-	_fatalError(false), _queueSize(0), _fullQueue(false), _repeat(false) {
+	_fatalError(false), _queueSize(0), _fullQueue(false), _repeat(false), _repeatCounter(0) {
 
 }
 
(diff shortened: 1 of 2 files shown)


DrMcCoy:
That looks better, yes. :)

Since it belongs to the same task, essentially, the next step would be to convert `enableTextInput()` as well. Similar, but instead of setting `_repeat`, you'd be calling `SDL_StartTextInput()` and `SDL_StopTextInput()`.

This would go into another commit, but onto the same branch / pull request. Try keeping it into this very same PR, without closing it and reopening a new one.

DrMcCoy:
Oh, and if you're done and want me to have another look over your PR, please leave a comment here, saying so. I do not get notifications for added commits, but I do get them for new comments. :)

Those two commits, c506154383f7dc820562c0a594c9fef9660dc948 and 08ed9608ba3abf761fee3bf929861fe9d100d9f9, should be squashed together. You could have done that with git commit --amend (which modifies last commit), or now with an interactive git rebase. Afterwards, you can git push -f to your branch, updating this PR.

DrMcCoy:
Oh, also, you can see that new commits have been added to the tree in https://github.com/xoreos/xoreos.

If you haven't already, add that repository as another remote to your fork's repository (naming it "upstream" is common). You can then update that remote whenever new commits have been done, and pull them into your master branch.

Whenever you update your pull request anyway, you can rebase the feature-branch ontop of the changed upstream, keeping it up-to-date with the current development of xoreos. This makes it easier to review. Also, when there's a conflict with things that have been changed in the meantime, and your proposed changes, you'll get to see this conflict, and can resolved it there and then.

Koffeinfriedhof:
I've played around with git rebase and --amend and I'm quite unsure what I've done now :D 

Tried refactoring it due to readability

DrMcCoy:
It seems like you somehow sandwiched my commits between yours?

What you can do is create a new branch off of this one, then go back to this, git reset --hard to upstream master, and then cherry-pick your commits (which are still held in that other branch) onto this branch. And then you can delete that backup branch again.

(Theoretically, it can be done without a backup branch, because the commits will still be reachable through git reflog.)

Koffeinfriedhof:
I've now cleaned up my branch using your email-guide! Thank you a lot for this great support!

DrMcCoy:
Could you squash 39cb1ea614f21d16435d6208f83bf831630c88bb and 5a82c34c3da36e9414438085dd320a23bee9dc3a into a18d4330a13fe930e59c82091d08a572ba3b83f3 as well? I.e. the end result should be two commits: one adding a counter to `EventsManager::keyRepeat()`, and one adding a counter to `EventsManager::enableTextInput()` (including the cleanups).

Also, the commit message should be "Add counter", not "Added counter". We want commit messages in indicative mood, so that you can say "If applied, this commit will X", where X is the commit message. Roughly, at least. And ignoring the "SUBSYSTEM:" thing. :P

Koffeinfriedhof:
Done.

Koffeinfriedhof:
I've tried it! Please watch coding style again, because I've changed some settings in my IDE. The indent was different from your settings (4 spaces = 1 tab).

DrMcCoy:
Please, squash these into their respective commit again. The result should be two commits, one for adding the counter to keyRepeat (including the Exception) and one adding the counter to textInput (including the Exception).

In the end, it should look like that this was your code all along.

DrMcCoy:
Apart from that tiny nitpick in the throws and the squashing, it looks pretty good now. :)

If you don't know how to pick the commit apart: with git reset --soft, you can return to a previews commit and not change the code in the files (so the changes are still there). Then with git add -p, you can selectively add lines from the files into the index, and commit them with git commit. So you can create the one changed file into two commits, one for the exception change in enableKeyRepeat, and one for the exception change in enableTextInput. These can then be squashed into the respective base commits with an interactive rebase.

Koffeinfriedhof:
Thank you for guiding me through git :) I think I've got it now.

DrMcCoy:
Neat. :)

There is, unfortunately, another issue. I just ran xoreos with your changes, and it immediately dies with "ERROR: EventsManager::enableTextInput(): Counter underrun".

The problem is the `enableTextInput(false)` line in EventsManager::init(), in events.cpp:81. This is there to make sure that the init() function always disables text input.

The fix I propose is to set `_textInputCounter` to 1 before that `enableTextInput(false)` line. (You can then also remove the line below that sets it to 0). This, again, should be squashed into the commit enableTextInput commit, since it still belongs in there.

While you're at it, can you add an explanatory sentence to the commit message of both commits (not in the first line; leave the second line empty, and then start writing in the third line)? Something like "This is part of the implementation to make keyboard states re-entrant (see GitHub issue #99)".

Koffeinfriedhof:
Wouldnt it be better to just handle this in the functions itself? Remove the throw exceptions and check if _counter is 0 before decreasing.
Or is it better to keep that exception and use 1 as default value for the constructor. Has this to be commented in any way?

DrMcCoy:
No, the general case should still be that it's a usage error. The init case is a special case, an exception.

Yeah, that's a good idea, add a short comment above these two lines (the `_textInputCount = 1`, and the `enableTextInput(false)`), explaining that we're forcing text input to be always disabled in the init phase. :)

Koffeinfriedhof:
Done. Please check.

DrMcCoy:
Just another tiny nitpick, but otherwise great :)

DrMcCoy:
Hmm, also, sorry for the late reply. I just noticed you already did that yesterday. For some reason, my mails seem to be getting delayed by a day now... Possibly my greylisting is running wild, or something. Need to check that.

DrMcCoy:
Okay, this looks good now. Thanks! :D

I pushed it as 6e4f4435ab44517e71347e6cc6bd3b5e1089d287 and b9da98aa6c4816b3b3f4f8c2d0de1c496c4f6d83.

PR #117, opened 2016-02-16
DIST: Add Fedora specific spec file

TC01:
So, I made a Fedora-specific spec file for xoreos. 

The BuildRequires are sufficient such that it builds under [mock](https://fedoraproject.org/wiki/Mock) (the Fedora chroot buildsystem for building packages), and I was able to install and run the result. :)

It's also compliant with current Fedora packaging guidelines, but otherwise is pretty similar to the [OBS spec](https://build.opensuse.org/package/view_file/home:DrMcCoy:xoreos/xoreos/xoreos.spec?expand=1) that @DrMcCoy linked me on IRC when I asked about this.

I'd be happy to make Fedora-specific specs for xoreos-tools and phaethon as well.

**rpmfusion dependencies**

There are three dependencies (libmad, faad, and xvidcore) that aren't in the official Fedora repos, but are instead in [rpmfusion-free](http://rpmfusion.org/). rpmfusion-free mostly contains Free Software that Fedora Legal doesn't want to ship due to patent issues (e.g. MP3 support, etc.), so to actually build on a Fedora system you'd need the rpmfusion-free repository.

I assume OBS takes care of that, since the OBS spec builds, but I've never actually used it for Fedora packages.

I'd offer to set up a [copr](https://copr.fedorainfracloud.org/) to build xoreos packages for Fedora in a native Fedora environment, but I can't due to the above, sadly. :(

### DIFF
--- a/dists/fedora/build-from-git.sh
+++ b/dists/fedora/build-from-git.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# This is a helper script to update the Release tag of the specfile
+# and then invoke rpmbuild --build-in-place with the correct options
+# to build from the root of the git checkout.
+
+# You need rpmbuild to build the rpm.
+
+# Thanks to StackOverflow:
+# https://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in
+RPM_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+XOREOS_ROOT=`git rev-parse --show-toplevel`
+SPEC_NAME=${RPM_DIR}/xoreos.spec
+
+# Compute the release snapshot, create a temporary spec. Also redine _rpmdir
+SNAPSHOT=`date +"%Y%m%d"`
+sed "s,\%global snapshot .*,\%global snapshot ${SNAPSHOT}git," < ${SPEC_NAME} > ${SPEC_NAME}.tmp
+sed "1s,^,\%define _rpmdir ${RPM_DIR}\n," -i ${SPEC_NAME}.tmp
+
+# Create a buildroot directory.
+mkdir -p ${RPM_DIR}/buildroot
+
+# Invoke rpmbuild from the root of the git repo (so --build-in-place works).
+# Also set the --buildroot and --
+cd ${XOREOS_ROOT}
+rpmbuild --build-in-place --rmspec --buildroot ${RPM_DIR}/buildroot -bb ${SPEC_NAME}.tmp
+rm -rf ${RPM_DIR}/buildroot
--- a/dists/fedora/xoreos.spec
+++ b/dists/fedora/xoreos.spec
@@ -1,0 +1,85 @@
+# If you want to build the current git checkout, run "build-from-git.sh".
+# If you want to build the last stable release of xoreos instead,
+# build from this specfile directly.
+
+# Note: xoreos depends on packages from rpmfusion-free.
+# So it won't be able to build in e.g. a Fedora Copr
+# (https://copr.fedorainfracloud.org/)
+# Presumaby OBS will handle this properly?
+
+# Globals, overridden by build script.
+%global snapshot 0
+
+Name:           xoreos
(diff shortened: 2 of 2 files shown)


DrMcCoy:
Hmm, is there a way to build directly from the repository, i.e. the current master branch instead of the release tarball? The Debian/Ubuntu file builds directly from a checkout, while the Arch file checks out the repository.

Also, the commit message is not [according to our guidelines](https://wiki.xoreos.org/index.php?title=Commit_Guidelines). Specifically, it should start with a subsystem ("DIST: " in this case), be in the imperative mood (so "Add" instead of "Added") and should not end with a period.

TC01:
> Hmm, is there a way to build directly from the repository, i.e. the current master branch instead of the release tarball? The Debian/Ubuntu file builds directly from a checkout, while the Arch file checks out the repository.

Oh, hm. 

Well, Source0 can be set to https://github.com/xoreos/xoreos/archive/master.zip. The setup macro's invocation needs to be modified slightly in that case but it still should work.

And then a line could be stuck into the %prep section that downloads that archive.

Alternatively instead of putting it in %prep, a small build script could be included that downloads the archive (or creates it from the checkout) and sets the release tag to something involving the current date and then invokes rpmbuild.

What do you think?

> Also, the commit message is not according to our guidelines. Specifically, it should start with a subsystem ("DIST: " in this case), be in the imperative mood (so "Add" instead of "Added") and should not end with a period.

Oh, whoops, sorry about that! I'll reword when fixing the above.

DrMcCoy:
> Source0 can be set to https://github.com/xoreos/xoreos/archive/master.zip

Hmm, I take it then that it can't invoke git to download a full repo checkout, i.e. what the Arch file does? Downloading the zip is certainly an option, but then we're losing the .git directory, and so the git version string will not be compiled into the binary. Which is a bit of a bummer, IMHO.

> Alternatively instead of putting it in %prep, a small build script could be included that downloads the archive (or creates it from the checkout)

I for one would prefer that then, create it from the checkout. That would mean that the Fedora file would then work similar to the Debian/Ubuntu file, in that it needs to run from within a repo checkout.

Unless that's a completely weird thing to do in Fedora. I don't really know much about this distribution, frankly. :)

TC01:
> I for one would prefer that then, create it from the checkout. That would mean that the Fedora file would then work similar to the Debian/Ubuntu file, in that it needs to run from within a repo checkout.

> Unless that's a completely weird thing to do in Fedora. I don't really know much about this distribution, frankly. :)

So apparently I didn't know as much about rpm(build) as I thought I did, because it has a ```-build-in-place``` option. So running ```rpmbuild --build-in-place -bb xoreos.spec``` appears to just do the right thing if it's ran from the root of the checkout and ignores Source0. :) So a script to tar things up isn't actually needed, and this way we get the git version string too.

(And also, if someone wants to build an actual package for the last released version, they can do that, too).

I think I'll still include a build script anyway that sets the right flags on rpmbuild for the in-place build and also updates the spec (to stick the current date in the release tag).

TC01:
Done (and reworded commit message). The ```build-from-git.sh``` shellscript in dists/fedora does the following:

* Creates a temporary copy of the spec file, updating the Release tag using sed to refer to the current date. The version of a xoreos RPM will be either something like "0.0.4-1.fc23" or "0.0.4-1.20160217git.fc23"; so the latter will upgrade the former. (Then obviously the upgrade to the next stable version would upgrade both packages).

* Moves to the root of the xoreos git checkout.

* Invokes ```rpmbuild --build-in-place``` with the right arguments to build the package in dists/fedora/ and to not build a SRPM, just the binary RPM. And obviously this will build from the current checkout rather than from the Source URL in the spec file.

* "make clean" is now invoked before building.
 
I'm happy to adjust further. I *think* this is a reasonably sane way to do this, though. :)

DrMcCoy:
Looks good, thanks :)! I pushed it as 5fa334f12e067cf7ad4990073cf2c8fd93953511.

If you could do that same for xoreos-tools and Phaethon as well, that would be great.

PR #118, opened 2016-02-19
BUILD: Make sure to actually respect libiconv compile flags.

christiank:
Options pertaining to libiconv were being added to a list that wasn't being
used anywhere else.


DrMcCoy:
Ah, yes, good catch! Thanks! :)

Pushed it as c7c3d8b5036b12c53cd364d788ac964b221b14f1 (slight change of the commit message, to remove the period at the end of the first line).

PR #119, opened 2016-03-30
kotor1 menu background rendered with its own class

farmboy0:


src/graphics/resolution.h
```diff
@@ -0,0 +1,81 @@
...
+	resolutions.push_back(Resolution(1152,  864));
+	resolutions.push_back(Resolution(1024,  768));
+	resolutions.push_back(Resolution(800 ,  600));
+	resolutions.push_back(Resolution(640 ,  480));
+	resolutions.push_back(Resolution(320 ,  240));
+	resolutions.push_back(Resolution(320 ,  200));
+	return resolutions;
+}
+
+static std::vector<Resolution> resolutions = initResolutions();
```
DrMcCoy: Good idea moving this to the graphics code. :)

But: using a global std::vector<>, filled by the initialization code like this isn't really all that great. We want to avoid doing that if possible. I'd rather you use a plain static array of Resolution structs. That can be filled out directly, too, and can live in the data section of the binary. (I probably should have done that in the first place :) )

Also, we don't need this data to be mutable, right? So it should be const. Also, also, please name it something like "kResolutions".

See kLanguageStrings in src/aurora/language_strings.h (https://github.com/xoreos/xoreos/blob/master/src/aurora/language_strings.h#L41) for how it should probably look like. You can then use the ARRAYSIZE macro to iterate over this array (see https://github.com/xoreos/xoreos/blob/master/src/aurora/language.cpp#L258).

Another nitpick: It should be "Refactor", not "Refaktor" in the commit message. :P

DrMcCoy:
These upper two commits, the resolution list ones, really should be in a separate pull request, though.

src/engines/kotor/gui/guibackground.h
```diff
@@ -0,0 +1,86 @@
...
+
+static const Common::UString &kBackgroundTypeMenu   = "back";
+static const Common::UString &kBackgroundTypeComp0  = "comp0";
+static const Common::UString &kBackgroundTypeComp1  = "comp1";
+static const Common::UString &kBackgroundTypeLoad   = "load";
+static const Common::UString &kBackgroundTypeMap    = "map";
+static const Common::UString &kBackgroundTypePazaak = "pazaak";
+static const Common::UString &kBackgroundTypeStore  = "store";
+
+class GuiBackground : public Graphics::GUIElement, public Events::Notifyable {
```
DrMcCoy: Is there a reason you used "Gui" instead of "GUI"? I think everywhere else I spelled it "GUI".

src/engines/kotor/gui/guibackground.cpp
```diff
@@ -0,0 +1,142 @@
...
+	if (ResMan.hasResource(name, Aurora::kResourceImage)) {
+		_texture = TextureMan.get(name);
+
+		_vertexX1 = (_screenWidth - width) / 2;
+		_vertexX2 = _vertexX1 + width;
+		_vertexY1 = (_screenHeight - height) / 2;
+		_vertexY2 = _vertexY1 + height;
+
+		_render = true;
+		status("found background %s", _texture.getName().c_str());
```
DrMcCoy: Minor nitpick, but isn't it a bit too spammy for status()? Maybe put it in a debugC() instead?

src/engines/kotor/gui/guibackground.cpp
```diff
@@ -0,0 +1,142 @@
...
+	glMatrixMode(GL_PROJECTION);
+	glLoadIdentity();
+	glOrtho(0.0, _screenWidth, 0.0, _screenHeight, -1.0, 1.0);
+
+	glMatrixMode(GL_MODELVIEW);
+	glLoadIdentity();
+
+	glPushMatrix();
+
+	glBegin(GL_QUADS);
```
DrMcCoy: That's then of course another piece of code that needs to be rewritten once the OpenGL rewrite is further along. :P

src/engines/kotor/gui/guibackground.cpp
```diff
@@ -0,0 +1,142 @@
...
+			if (tryBackground(Graphics::resolutions[i].width, Graphics::resolutions[i].height)) {
+				return;
+			}
+		}
+	}
+}
+
+bool GuiBackground::tryBackground(int width, int height) {
+	Common::UString name = Common::UString::format("%ux%u%s", width, height, _type.c_str());
+	if (ResMan.hasResource(name, Aurora::kResourceImage)) {
```
DrMcCoy: Hmm, you could just do TextureMan.get(), but wrap it in a try {} block. Or is there a reason you want to differentiate between "Texture resource doesn't exist" and "Texture resource fails to load"?

src/engines/kotor/gui/gui.cpp
```diff
@@ -265,13 +277,10 @@ WidgetListBox *GUI::getListBox(const Common::UString &tag, bool vital) {
...
 void GUI::addBackground(const Common::UString &background) {
-	WidgetPanel *backPanel =
-		new WidgetPanel(*this, "Background", "1600x1200" + background,
-		                -800.0f, -600.0f, 1600.0f, 1200.0f);
-
-	backPanel->movePosition(0.0f, 0.0f, 10.0f);
-
-	addWidget(backPanel);
+	if (!_background)
+		_background = new GuiBackground(background);
```
DrMcCoy: You need to delete _background in the destructor.

src/engines/kotor/gui/guibackground.cpp
farmboy0: yes, there is actually.
"Texture resource doesn't exist" is not an error, it is expected.
"Texture resource fails to load" would be an actual error and should be treated as such.

src/engines/kotor/gui/guibackground.cpp
farmboy0: it is in good company with GUIQuad and sonic/AreaBackground :)

src/engines/kotor/gui/guibackground.h
farmboy0: no reason.

src/graphics/resolution.h
farmboy0: why dont you like vectors?
cant they be made const?

src/engines/kotor/gui/gui.cpp
farmboy0: oops

src/graphics/resolution.h
DrMcCoy: Well, it would still be allocated on the heap and during the init code vs. being pulled into static memory wholesale with the data section; and it adds to the global constructors. It's simply unnecessary.

src/engines/kotor/gui/guibackground.cpp
DrMcCoy: Okay :)

PR #121, opened 2016-05-07
move window handling to its own class

farmboy0:


farmboy0:
The following functionality should be tested: 
* changing FSAA level
* switching between fullscreen and windowed and back
* changing window size in windowed and fullscreen mode

src/graphics/graphics.cpp
```diff
@@ -957,8 +761,8 @@ Renderable *GraphicsManager::getGUIObjectAt(float x, float y) const {
 		return 0;
 
 	// Map the screen coordinates to our OpenGL GUI screen coordinates
-	x =            x  - (_width  / 2.0f);
-	y = (_height - y) - (_height / 2.0f);
+	x =            x  - (WindowMan.getWindowWidth()  / 2.0f);
```
Supermanu: Minor nitpick here: space alignment not needed anymore.

DrMcCoy:
Yeah, this looks good, IMHO. I like it.

I especially like the use of an enum for the renderer type; that's probably what I should have done in the first place, yes.

Two things I noticed (but are not a fault of your code, just general observations for the TODO list):
- When changing the FSAA settings in NWN (either the main menu or the ingame options menu), the highlight bounding boxes suddenly appear offset by half an object's height. This doesn't seem to change the cursor/object interactions, though. When restarting xoreos with the new FSAA settings, they're okay again, too. Very weird.
- The game loader progress bar looks silly in higher FSAA settings, because the bar is actually a string of Unicode box glyphs. With FSAA up, the gaps between them are visible. That really shouldn't be, but the bar should probably be drawn in a better way anyway, I guess.

Do you want this to be merged now or should this wait for after @mirv-sillyfish's stuff? #123 rebases cleanly (with an automatic 3-way merge) on top of this change, but then needs a little fix in graphics.cpp to replace _width/_height by WindowMan.getWindowWidth()/WindowMan.getWindowHeight() to make it compile again.

DrMcCoy:
Merged with 287d2d8..812b1af. Thanks! :)

I did, however, remove the SDL_WINDOW_RESIZABLE flag since it's currently not working anyway. I added "Let the user resize the xoreos window" as a TODO. If you have a working implementation, please do a PR for it. I'm really not sure how well it'll work though, what with OpenGL needing to be recreated for every change, but yeah, if I see it working, sure.

Also note that Sonic, as a Nintendo DS game, is inherently fixed-size. Currently, the engine code even overrides a user config choice there. In the future, we might add something to scale the 2D images? But even that would probably work best in full integer steps. So an engine might explictly want to disable user scaling.

I also added two issues, #131 (boundbox displacement while changing FSAA) and #132 (borked fullscreen resolution change, and added the progress bar thing as another TODO.

PR #123, opened 2016-06-28
Ithasreturned

mirv-sillyfish:
GRAPHICS: shader support

Support runtime generated shaders, and shader based rendering of world geometry.
GUI is not rendered using shaders. OpenGL compatibility profile still required for GL3.2
to be used.


src/graphics/shader/shaderbuilder.cpp
```diff
@@ -273,12 +301,101 @@ ShaderBuilder::~ShaderBuilder() {
...
 		header = header_default_3vert;
 		body = body_default_start_3vert;
+	} else {
+		header = header_default_2vert;
+		body = body_default_start_2vert;
+	}
+}
 
-		switch (flags) {
+void ShaderBuilder::addVertexShaderString(Common::UString &header, Common::UString &body, uint32 passType, uint32 blendType, bool isGL3) {
```
DrMcCoy: The parameter `blendType` is unused in this method (and will throw an "unused parameter" warning). Please wrap `blendType` in the `UNUSED()` macro to document that it is unused (and silence that warning).

DrMcCoy:
Keep in mind that commit messages should, after the subsystem tag, start with a capital letter and shouldn't end with a period. I.e. it should be "GRAPHICS: Render decals after primary mesh data" instead of "GRAPHICS: render decals after primary mesh data."

PR #124, opened 2016-07-03
Fixes

farmboy0:
Some small fixes

### DIFF
--- a/src/aurora/types.h
+++ b/src/aurora/types.h
@@ -385,7 +385,7 @@ enum GameID {
 };
 
 enum ResourceType {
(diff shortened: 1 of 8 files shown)


src/aurora/types.h
```diff
@@ -385,7 +385,7 @@ enum GameID {
 };
 
 enum ResourceType {
-	kResourceImage,  ///< A music resource.
+	kResourceImage,  ///< A image resource.
```
DrMcCoy: Should be "An image", though. :P

Also, the commit message should say "Fix", not "Fixed".

PR #125, opened 2016-08-15
First part of removing states from models

farmboy0:
This part introduces special node structures to hold mesh and dangly data.
It also allows attaching models directly to other models as this fixes a bug with the current animation system.


DrMcCoy:
Hmm, for some reason, characters are now slightly transparent in your branch.

This is noticable, for example, on the hair of Bim in the NWN OC Prelude. And in Chapter 2, on, for example, Linu:

lilutransparent lilusolid

It *could* be related to the environment mapping on her armor. Or maybe not, since it happens on heads, too (just not this noticable, for the most part).

src/graphics/aurora/modelnode.h
```diff
@@ -118,6 +118,63 @@ class ModelNode {
...
+
+		Dangly();
+	};
+
+	struct MeshData {
+		VertexBuffer vertexBuffer; ///< Node geometry vertex buffer.
+		IndexBuffer indexBuffer;   ///< Node geometry index buffer.
+	
+		std::vector<TextureHandle> textures; ///< Textures.
+
```
DrMcCoy: Trailing whitespace in lines 138 and 140

DrMcCoy:
Apart from that tiny style issue in 32484c3 and that bigger transparency issue, the PR looks rather good. Good job! :)

DrMcCoy:
Hmm, while checking your branch for leaks, I found something "interesting":

There's apparently at least two animations with the same name in at least one model. The model in question is "PLC_H09", the animation name is "die", and the model shows up in the entry area of the Chapter 1 of the NWN OC. I think it might be the mat on the ground where the infected lie, or maybe the infected themselves.

[Here is a small patch against your branch](https://gist.github.com/DrMcCoy/9f629e4c6b9481a043d154d938398261) that spits out a warning (and also for duplicate states). That also takes care of the memory leak, I think (because std::map::insert() won't overwrite a value, we're basically allocating the extra Animation and then throwing it away without delete'ing it; the patch just bails *before* allocating the Animation).

I'm not really sure how to handle this. Which of the two animations is the real one? Are the different, even? Does it matter? 

(I probably should have added a check there earlier, too. I just didn't think that this could happen, but, well, BioWare. :P)

This is also not necessarily related to your PR as such, but somewhat connected. I'm mainly just thinking out loud here, because it might interest you and so I'm not forgetting this. And maybe someone wants to chime in and/or research this further. :)

farmboy0:
I'll check it out and find a solution after comparing the values for both animations.
Btw one of the 2 checks should suffice because animations and states are the same thing in the models.

PR #126, opened 2016-08-18
Updated events.cpp file and tried to fix the bug

AbhiramU:
Added the  condition  which checks if the _textInputCounter is greater than zero then it will be a text input or else it wont be a text input

### DIFF
--- a/src/events/events.cpp
+++ b/src/events/events.cpp
@@ -282,14 +282,16 @@ void EventsManager::enableTextInput(bool textInput) {
 	if (!textInput && _textInputCounter == 0)
 		throw Common::Exception("EventsManager::enableTextInput(): Counter underrun");
 
-	if (textInput) {
-		if (_textInputCounter == 0)
-			SDL_StartTextInput();
+	if(textInput) {
 		_textInputCounter++;
 	} else {
-		if (_textInputCounter == 1)
-			SDL_StopTextInput();
-		_textInputCounter--;
+		_textInputCounter--;	
+	}
+	
+ 	if (_textInputCounter > 0) {
+ 		SDL_StartTextInput();
+ 	} else {
+		SDL_StopTextInput();
 	}
 }
 


DrMcCoy:
Hey there and thanks! :)

There's a few little issues with your pull request here, though:
First of all, what exactly is "the bug"? I'm a bit unclear about what the issue you're fixing is.

Then, please have a look at [the Developer Central page on our wiki](https://wiki.xoreos.org/index.php?title=Developer_Central), especially the [Code Formatting Conventions](https://wiki.xoreos.org/index.php?title=Code_Formatting_Conventions) and the [Commit Guidelines](https://wiki.xoreos.org/index.php?title=Commit_Guidelines).

Specifically, your commit message does not follow the guidelines in this way:
- You need to add a submodules prefix, in this case "EVENTS: "
- You should state what was fixed, not just "the bug"
- The long explanation after the first line needs to be wrapped at around 72 characters. I usally do 70 character, because the first line also mustn't exceed 70 characters

So a better commit message would be "EVENTS: Fix bug that [...]", with a short description what the bug is there (not exceeded 70 characters in total), then a blank line and then a longer explanation in the following lines (wrapped at about 72 characters).

The commit itself also doesn't strictly follow our guidelines:
- There's trailing whitespace (a tab, in this case) in lines 288 and 290
- Lines 291, 292 and 293 contain a space in the first column

If you could fix these issues, that would be great. And please, explain in the pull request text what the original bug is. Thanks! :)

DrMcCoy:
Oh, I'm seeing you have the awesome-for-beginners and up-for-grabs repositories forked.

Are you trying to implement https://github.com/xoreos/xoreos/issues/99 ? Because if so, I'm afraid somebody already has done this earlier. Sorry.

And, unless I'm mistaken, the current implementation is correct and works. Unless, of course, I *am* mistaken. Do correct me there, then. :)

AbhiramU:
Thank you for your valuable feedback.  I am new to open source. Hence forth will follow the guidelines as you mentioned..


Get Outlook for Android






On Thu, Aug 18, 2016 at 5:31 PM +0530, "Sven Hesse" <notifications@github.com> wrote:





Oh, I'm seeing you have the awesome-for-beginners and up-for-grabs repositories forked.

Are you trying to implement https://github.com/xoreos/xoreos/issues/99 ? Because if so, I'm afraid somebody already has done this earlier. Sorry.

And, unless I'm mistaken, the current implementation is correct and works. Unless, of course, I *am* mistaken. Do correct me there, then. :)

DrMcCoy:
Since this pull request seems to not be applicable anymore, I'm closing this.

@AbhiramU, we unfortunately do not have any easy issues for newcomers, at least not that I can think of. Reimplementing games, like xoreos does, is a complex task, with many parts completely "open" and unbound, without any rails newcomers can grab on.

You are of course free to look at [the rambly TODO list on our wiki](https://wiki.xoreos.org/index.php?title=TODO), but I'm afraid you won't find small and simple tasks there. I'm sorry, I would like to have easy tasks I could use to help steer new people into becoming contributors, but try as I might, I can't really think of anything.

If you have any comments, questions, anything, feel free to reach out to me. Either here, #xoreos on Freenode IRC or by mail, drmccoy@drmccoy.de, whatever you feel like. Cheers.

PR #129, opened 2016-10-13
BUILD: Appveyor integration for Windows builds.

berenm:
At long last, it's alive: https://ci.appveyor.com/project/berenm/xoreos

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,162 @@
+version: 1.0.{build}
+os: Visual Studio 2015
+clone_depth: 1
+
+matrix:
+  fast_finish: true
+
+platform: x86
+configuration: Debug
+
+environment:
+  global:
+    BOOST_ROOT: C:\Libraries\boost_1_59_0
+    BOOST_INCLUDEDIR: C:\Libraries\boost_1_59_0
+    FREETYPE_DIR: C:\projects\xoreos\libs\freetype-2.3.5-1-lib
+    LIBSDL2: SDL2-2.0.4
+    LIBFT2: freetype-2.3.5-1
+    LIBICONV: libiconv-1.9.2-1
+    LIBMAD: libmad-0.15.1b
+    LIBFAAD: faad2-2.7
+    LIBOGG: libogg-1.3.2
+    LIBVORBIS: libvorbis-1.3.5
+    LIBXVID: xvidcore-1.3.4
+
+    AUTORECONF_CMD: WANT_AUTOMAKE=latest AUTOMAKE='automake --foreign' autoreconf -fi
+    CONFIGURE_CMD: CC=cl CXX=cl ./configure --prefix=/
+    INSTALL_CMD: make install
+
+  matrix:
+    - CMAKE_GENERATOR: Visual Studio 14 2015
+      MSVC_ARCH: amd64_x86
+      MSBUILD_PLATFORM: Win32
+      PROGRAM_FILES_DIR: C:\Program Files (x86)
+
+      BOOST_LIBRARYDIR: C:/Libraries/boost_1_59_0/lib32-msvc-14.0
+
+      LIBOPENAL_DESTDIR: C:/Programs/OpenAL
+      LIBSDL2_DESTDIR: C:/Programs/SDL2
+      LIBMAD_DESTDIR: C:/Programs/MAD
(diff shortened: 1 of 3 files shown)


DrMcCoy:
Neat, thanks! :)

Now, what exactly do I have to do (besides merging that PR) to set up AppVeyor for xoreos?

berenm:
I think it's more or less like for Travis-CI. You have to register to Appveyor with your Github account, create a new project for xoreos. In the project settings, I have configured:

Custom configuration .yml file name: .appveyor.yml
Skip branches without appveyor.yml: check
Rolling builds: check

I left all the rest to the default settings.

There might be something to do with the notifications, but I believe this can be configured in the .appveyor.yml, like for Travis. The file reference is available here: https://www.appveyor.com/docs/appveyor-yml/

DrMcCoy:
Okay, merged with 71c748f261dabab563b752468ac7923111d10268 (I did change the build version number, though, just so that it fits with our versioning scheme a bit better).

Let's see if it works. :)

DrMcCoy:
Can you do this for xoreos-tools and Phaethon, too?

xoreos-tools should, I guess, be relatively straight-forward. It also doesn't depend on Boost at the moment, but pulling that in is on the TODO list, to have access to Boost's smart pointers, etc.

Phaethon might be more work, though, since it also depends on wxWidgets.

DrMcCoy:
Hmm, well, that build there failed with a connection time out while trying to fetch iconv... I'll restart :P

berenm:
Yeah, there are some connectivity issues with dependencies hosted on sourceforge... I enabled the cache mechanism in order to workaround them, and to speed up the build a little bit, but it requires a successful build at least to start caching things.

Otherwise, it could use mirrors but I didn't spend too much time looking into it.

berenm:
I'll try to see what I can do for the other repositories.

DrMcCoy:
If I want to explicitly disable an MSVC warning (like C4514, "unreferenced inline function has been removed"), I just have to add 4514 to the WARNINGS_DISABLE variable in CMakeLists.txt, right?

Because to me, this warning looks very useless and spams quite a lot.

DrMcCoy:
Errm, hmm, for some reason, you're building with /W3, but the CMakeList by default adds /W4. Maybe that should be changed instead?

berenm:
I don't think there is any specific compilation flags in the appveyor integration. Whatever is added in the CMakeLists should be taken into account in addition to the default MSVC flags.

DrMcCoy:
Interestingly, that commit just now also didn't change anything. Maybe I'm just confused?

Please have a look at https://ci.appveyor.com/project/DrMcCoy/xoreos/build/0.0.4%2BAppVeyor.4 , which looks very spammy with lots of warnings flags that shouldn't be enabled with anything less than /W4 (if I'm reading the MSVC online docs right).

As is, that unfortunately makes the log very not useful, because useful warnings are being drowned out. And AppVoyer even truncates it at a certain point. :/

berenm:
I'm looking at it right now, it's kind of surprising because I don't remember anything like that at all when I did my builds. If you look at my latest build before the PR it wasn't that verbose: https://ci.appveyor.com/project/berenm/xoreos/build/1.0.229

I was based on an older master so maybe is it it?

DrMcCoy:
Ooh, yes, I *did* change something. I did add compiler flags, including "-Wall". I only meant to do that for non-MSVC, but, yeah, there's no guards around it and I guess MSVC picks that up as well.

Entirely my bad, then.

DrMcCoy:
Hmm, for some reason AppVeyor broke. Seems to be during compilation of xvid if I'm reading this correctly?

I pushed a new branch set to an old commit that used to compile correctly but now fails.
Here's the failing log: https://ci.appveyor.com/project/DrMcCoy/xoreos/build/0.0.4+AppVeyor.237
And this is the old non-failing log, same commit: https://ci.appveyor.com/project/DrMcCoy/xoreos/build/0.0.4+AppVeyor.218

But in the earlier log, it seems it got the old binaries from cache, so it probably actually broke way earlier?

@berenm, If you find some time, could you check what's the problem there? Thanks! :)

berenm:
I think this should be it: https://github.com/xoreos/xoreos/pull/157

DrMcCoy:
For some reason, the build now randomly breaks during, I think, OpenAL compilation?
@berenm, could you take a look? Thanks!

PR #130, opened 2016-10-13
BUILD: Remove useless tree command in appveyor integration

berenm:
As in the title, it's a leftover of my debugging

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -92,7 +92,6 @@ install:
   - choco install nasm
   - set PATH="C:\Program Files (x86)\NASM";%PATH%
   - cmake -E chdir libs cmake -E tar xf %LIBXVID%.tar.gz
-  - tree /f libs\xvidcore\build\win32
   - if not exist libs\xvidcore\build\win32\libxvidcore.vcxproj cmake -E chdir libs/xvidcore devenv /upgrade build\win32\libxvidcore.sln
   - cmake -E chdir libs/xvidcore msbuild build\win32\libxvidcore.vcxproj /p:Configuration=Release /p:Platform=%MSBUILD_PLATFORM% /verbosity:quiet
   - cmake -E make_directory %LIBXVID_DESTDIR%
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,29 +61,31 @@ set_check_compiler_flag_cxx("-std=c++03")
 # Compiler warning flags
 
 # Extra warnings valid for both C and C++
-set_check_compiler_flag_cxx("-Wall")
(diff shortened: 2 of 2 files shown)


berenm:
Well I didn't want to add another commit to this one... wtf Github

DrMcCoy:
Yeah, PR are linked to the branch you used when making the PR. So pushing to master updated the PR.

That's why it's usually advised to create a feature branch for each PR.

berenm:
Yep, I'm still not very used to Github workflow.

Anyway, feel free to pick or redo the changes, I believe it should fix the verbosity issue.

DrMcCoy:
Thanks :)

berenm:
The `if ("${CMAKE_CXX_COMPILER_ID}" NOT STREQUAL "MSVC")` actually doesn't build, it should be `if (NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")`

DrMcCoy:
Aaah, yes, that looks way better now :)

DrMcCoy:
Hmm, is there a reason to use `if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")` instead of `if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")`? CMake seems to warn about the three `STREQUAL "MSVC"` lines here: https://ci.appveyor.com/project/DrMcCoy/xoreos/build/0.0.4+AppVeyor.8#L269 . The reason seems to be a policy change of when to dereference the variable further? Not sure how pressing that is, but I really wouldn't want that to break some time in the future. :P

berenm:
I think it's CMake language arcane. It is possible to explicitly expand variables like "${VARIABLE}", but CMake is also able to implicitly expand strings that look like variables.

I understand before some CMake version, "VARIABLE" would be expanded to the content of VARIABLE if such variable existed. Now they removed this absolute non-sense, and only implicitly expand variables if they are used unquoted. Explicit expansion does not change.

What this means for the current case is that MSVC is an existing CMake variable. It is quoted but with the old behavior, it was probably expanded to 1 when compiling with MSVC. On the other side, CMAKE_CXX_COMPILER_ID is explicitly expanded to "MSVC", which in turn is recursively expanded again to 1. When not compiling with MSVC, the MSVC variable does not exist, so is not expanded implicitly and no warning is issued.

With the new and less silly behavior, "MSVC" is quoted and will not be expanded. CMAKE_CXX_COMPILER_ID is explicitly expanded to "MSVC", which is a quoted string and the logic stops here.

I think this warning can be solved by requiring a more recent version of CMake as minimal requirement. Or maybe by unquoting the CMAKE_CXX_COMPILER_ID, which is the modernish way of writing CMake files AFAIK.

PR #133, opened 2016-12-21
Unit tests

DrMcCoy:
This PR adds general unit tests (using Google Test) to xoreos.

It's not 100% coverage, far from it. There's currently 1034 indidual tests in 70 file (with ~ one class per file), covering most of Common, many of the Aurora classes and 3 files for image decoders.

Included in the changeset is also some CMake changes. @berenm, if you could double-check that what I did there makes sense, that would be great. Thanks.

I'm also enabling the unit tests in Travis CI and AppVeyor. The former seems to work, but for some reason, [AppVeyor throws errors there](https://ci.appveyor.com/project/DrMcCoy/xoreos/build/0.0.4+AppVeyor.72#L4096). "Exception: Other", whatever *that* is supposed to mean. I don't suppose you might know what the issues is there, or how I can get more usual debug info, @berenm? Because that is just weird :/.

Going forward, more unit tests would be great. For example:
- PE EXE file in common/ and PE file archive in aurora/.
- Image decoders. Possible, but annoying to write.
- Nintendo DS formats.
- Aurora::ResourceManager? Would need functions to manipulate some structures more directly.
- FFT/DCT/MDCT/RDFT? Would be fiddly.
- Sound decoders, probably rather difficult to test. Video decoders doubly so.
- Fonts and models would be neat, but also difficult. And currently, the loaders directly use OpenGL objects...
- Engine-/game-level tests? World object structures? Script functions?

Likewise, new or changed functions in classes already covered by the tests should now be getting unit tests as well.

Comments, criticism, etc. welcome, of course. :)


tests/aurora/2dafile.cpp
```diff
@@ -0,0 +1,709 @@
...
+#include "src/common/ustring.h"
+#include "src/common/error.h"
+#include "src/common/memreadstream.h"
+#include "src/common/memwritestream.h"
+
+#include "src/aurora/types.h"
+#include "src/aurora/2dafile.h"
+#include "src/aurora/gdafile.h"
+
+static const byte k2DAASCII[] =
```
clone2727: Should be `char`?

Yes, I know it ends up in a `MemoryReadStream`, but it seems odd to store text as byte here.

tests/aurora/biffile.cpp
```diff
@@ -0,0 +1,338 @@
...
+	0x72,0x65,0x63,0x6B,0x2C,0x20,0x62,0x6F,0x75,0x6E,0x64,0x6C,0x65,0x73,0x73,0x20,
+	0x61,0x6E,0x64,0x20,0x62,0x61,0x72,0x65,0x0A,0x54,0x68,0x65,0x20,0x6C,0x6F,0x6E,
+	0x65,0x20,0x61,0x6E,0x64,0x20,0x6C,0x65,0x76,0x65,0x6C,0x20,0x73,0x61,0x6E,0x64,
+	0x73,0x20,0x73,0x74,0x72,0x65,0x74,0x63,0x68,0x20,0x66,0x61,0x72,0x20,0x61,0x77,
+	0x61,0x79,0x2E
+};
+
+GTEST_TEST(BIFFile10, getNameHashAlgo) {
+	Common::MemoryReadStream *stream =
+		new Common::MemoryReadStream(kBIF10File, sizeof(kBIF10File));
```
clone2727: Having to type the variable twice always bugs me, especially when it's easy to have a copy/paste error. Maybe it would be better to have something like this as a function in `MemoryReadStream`:

```C++
template <size_t N>
static MemoryReadStream *CreateFromArray(const byte (&array)[N]) {
    return new MemoryReadStream(array, N);
}
```

tests/aurora/bzffile.cpp
```diff
@@ -0,0 +1,184 @@
...
+	"I met a traveller from an antique land\n"
+	"Who said: Two vast and trunkless legs of stone\n"
+	"Stand in the desert. Near them, on the sand,\n"
+	"Half sunk, a shattered visage lies, whose frown,\n"
+	"And wrinkled lip, and sneer of cold command,\n"
+	"Tell that its sculptor well those passions read\n"
+	"Which yet survive, stamped on these lifeless things,\n"
+	"The hand that mocked them and the heart that fed:\n"
+	"And on the pedestal these words appear:\n"
+	"\'My name is Ozymandias, king of kings:\n"
```
clone2727: You don't need to escape a single quote from within double quotes.

DrMcCoy:
Okay, after *a lot* of fiddling (grrrr), it seems that many unit tests fail with error code `-1073741515`, which is `0xC0000135`, which is `STATUS_DLL_NOT_FOUND`. Of course, it doesn't say *which* DLL...

The ones that do work are:
- tests_common_test_scopedptr
- tests_common_test_disposableptr
- tests_common_test_ptrlist
- tests_common_test_ptrvector
- tests_common_test_ptrmap
- tests_common_test_binsearch
- tests_common_test_maths
- tests_common_test_vector3
- tests_common_test_matrix4x4
- tests_common_test_boundingbox

All other unit tests fail. The ones that work are all small, contained classes/templates and might not even need any DLL at all...

DrMcCoy:
Okay, the tests all build and run in AppVeyor now. The ones that read/write files still fail, though, because the paths don't exist.

cmake/CMakeAM.cmake
```diff
@@ -125,13 +131,74 @@ function(convert_automake_internal AM_INPUT_FILE CMAKE_OUTPUT_FILE)
...
+      list(APPEND FLAGS_INCLUDES "${FLAG}")
+    else()
+      list(APPEND FLAGS_MISC "${FLAG}")
+    endif()
+
+    math(EXPR i "${i} + 1")
+  endwhile()
+
+  string(REPLACE ";" " " FLAGS_MISC "${FLAGS_MISC}")
+  set_target_properties(${TARGET} PROPERTIES APPEND_STRING PROPERTY COMPILE_FLAGS "${FLAGS_MISC}")
```
berenm: Does this work ? I didn't know about this `APPEND_STRING` keyword, and it is only documented for the `set_property` function, same question for the three other places it is used.

berenm:
Sorry for the late reply, I see that you managed to figure out why the tests weren't running on Appveyor, that's nice. The CMake part looks good to me.

cmake/CMakeAM.cmake
DrMcCoy: Yeah, interestingly enough it does work. Or at least it does for me, when I tested it.

I found it recommended in a StackOverflow question here: http://stackoverflow.com/a/19161129 , that's how I got that idea. :P

Is there a difference between `set_target_properties(...)` and `set_property(TARGET ...)`? Should I rather use one over the other?

DrMcCoy:
Neat, thanks for checking! :)

Yeah, got the AppVeyor stuff figured out now (although copying the DLLs is a bit hacky). Took a lot of just trying out things, because I don't know jack about Windows anymore.

Because I have no idea how to get Windows to tell me *which* DLL is missing, I created an "artifact" in AppVeyor consisting of a test binary, downloaded it and looked into it with objdump on my machine. Turns out iconv.dll is needed for some reason (because of the Boost libraries, maybe?), so I added that to the downloads. And also copied all the DLLs needed to the bin directory.

I also now changed the tests that write into files to use the tmp directory (using Boost to query that directory) and fixed two unrelated issues (one in a unit test, one in common/). Now the unit tests all run and pass in AppVeyor as well. :)

tests/aurora/ndsrom.cpp
```diff
@@ -0,0 +1,182 @@
...
+	"Look on my works, ye Mighty, and despair!'\n"
+	"Nothing beside remains. Round the decay\n"
+	"Of that colossal wreck, boundless and bare\n"
+	"The lone and level sands stretch far away.";
+
+/* Percy Bysshe Shelley's "Ozymandias", within a NDS ROM file.
+ * Note: this isn't really a valid, standard NDS file. It only
+ * fills the header fields we actually care about. If the NDSFile
+ * class is extended to care about the other fields as well, this
+ * file data has to be changed as well. */
```
clone2727: Did you manually generate this? I think the test would be better if generated from some tool.

tests/common/encoding.h
```diff
@@ -0,0 +1,47 @@
...
+#include <cstdio>
+#include <cstdlib>
+
+#include "src/common/encoding.h"
+
+#ifndef SKIP_RETURN_CODE
+	#define SKIP_RETURN_CODE 77
+#endif
+
+static const int kSkipTest = SKIP_RETURN_CODE;
```
clone2727: Why bother with the `kSkipTest` constant at this point when the `SKIP_RETURN_CODE` macro already exists?

Shouldn't this logic be in a common place anyway?

tests/common/filepath.cpp
```diff
@@ -0,0 +1,188 @@
...
+	EXPECT_STREQ(Common::FilePath::changeExtension("/path/"        , "new"   ).c_str(), "/path/.new");
+	EXPECT_STREQ(Common::FilePath::changeExtension("/path/"        , ""      ).c_str(), "/path/");
+}
+
+GTEST_TEST_F(FilePath, relativize) {
+	EXPECT_STREQ(Common::FilePath::relativize("/path/to", "/path/to/file.ext").c_str(), "file.ext");
+	EXPECT_STREQ(Common::FilePath::relativize("/other"  , "/path/to/file.ext").c_str(), "");
+
+	/* Common::FilePath::relativize() only works with normalized paths...
+	 * These here are common failure modes.
```
clone2727: I don't like the behavior of `relativize` only working with normalized paths -- can't it just do the normalization itself?

(Also, `makeRelative` would be a nicer name, IMO)

tests/common/util.cpp
```diff
@@ -0,0 +1,184 @@
...
+	EXPECT_DOUBLE_EQ(readNintendoFixedPoint(0x00000000,  true, 15, 16),      0.00);
+	EXPECT_DOUBLE_EQ(readNintendoFixedPoint(0x00010000,  true, 15, 16),      1.00);
+	EXPECT_DOUBLE_EQ(readNintendoFixedPoint(0xFFFF0000,  true, 15, 16), -    1.00);
+	EXPECT_DOUBLE_EQ(readNintendoFixedPoint(0xFFFF0000, false, 16, 16),  65535.00);
+
+	/* The fractional part in Nintendo's fixed point format is expressed in simple
+	 * fractions of 2^fBits. That's pretty different from IEEE floats, so their
+	 * representation of non-trivial fractions diverge quite a bit. Which means
+	 * checking that the result is within a few double ULP doesn't make much sense.
+	 */
```
clone2727: Wouldn't it make more sense not to treat the fixed point numbers as floating point?

DrMcCoy: (I already talked with @clone2727 about this, so this is just for the record: nope, they have to be treated as floating point numbers, because the numbers are thrown into Matrix4x4 and OpenGL.)

PR #134, opened 2017-05-07
DeS3TC cursor fix

Nostritius:
I recognized a bug in the graphics system which caused it to crash when loading a compressed cursor image on a non-s3tc system. I added three lines to make the cursor image decompress on such systems.

### DIFF
--- a/src/graphics/aurora/cursor.cpp
+++ b/src/graphics/aurora/cursor.cpp
@@ -30,6 +30,8 @@
 #include "src/aurora/types.h"
 #include "src/aurora/resman.h"
 
+#include "src/graphics/graphics.h"
+
 #include "src/graphics/images/decoder.h"
 #include "src/graphics/images/txi.h"
 #include "src/graphics/images/tga.h"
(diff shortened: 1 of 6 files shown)


src/graphics/aurora/cursor.cpp
```diff
@@ -112,6 +114,12 @@ void Cursor::load() {
 
 	TXI *txi = new TXI();
 	txi->getFeatures().filter = false;
+	
+	//if we need DeS3TC, we decompress the cursor image
```
DrMcCoy: A space after the `//`, and then start with with uppercase, please.

I.e.
`// If we ...`

src/graphics/aurora/cursor.cpp
```diff
@@ -112,6 +114,12 @@ void Cursor::load() {
...
+	if(GfxMan.needManualDeS3TC())
```
DrMcCoy: Please put a space after the if, and add the opening brace to the same line

`if (GfxMan.needManualDeS3TC()) {`

PR #135, opened 2017-05-07
GRAPHICS: Fix a bug with missing S3TC decompression on cursors

Nostritius:
I recognized a bug in the graphics system which caused it to crash when loading a compressed cursor image on a non-s3tc system. I added some lines to make the cursor image decompress on such systems.


DrMcCoy:
Merged it, thanks! :)

PR #137, opened 2017-05-15
AURORA: Add functionality to add child guis to a gui

Nostritius:
While I was working on the KotOR menus, i recognized, that KotOR is using combined guis, where the actual interface is consisting of two or more seperated gui files which are independently switched. I added the possibility to do that in the xoreos gui class.

### DIFF
--- a/src/engines/aurora/gui.cpp
+++ b/src/engines/aurora/gui.cpp
@@ -70,6 +70,11 @@ void GUI::show() {
 			widget.show();
 	}
 
+	std::list<GUI*>::iterator iter;
+	for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+		(*iter)->show();
+	}
+
 	GfxMan.unlockFrame();
 }
 
@@ -79,6 +84,11 @@ void GUI::hide() {
 	// Hide all widgets
 	for (WidgetList::iterator widget = _widgets.begin(); widget != _widgets.end(); ++widget)
 		(*widget)->hide();
+
+	std::list<GUI*>::iterator iter;
+	for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+		(*iter)->hide();
+	}
 
 	GfxMan.unlockFrame();
 }
@@ -96,6 +106,13 @@ uint32 GUI::run(uint32 startCode) {
 	while (_returnCode == kReturnCodeNone) {
 		// Call the periodic run callback
 		callbackRun();
+
+		std::list<GUI*>::iterator iter;
+		// Call the periodic run callback of the children guis
+		for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+			(*iter)->callbackRun();
+		}
+
 		if (_returnCode != kReturnCodeNone)
 			break;
 
@@ -105,10 +122,25 @@ uint32 GUI::run(uint32 startCode) {
 
 		// Handle events
 		Events::Event event;
-		while (EventMan.pollEvent(event))
+		while (EventMan.pollEvent(event)) {
 			addEvent(event);
+			for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+				(*iter)->addEvent(event);
+			}
+		}
 
 		processEventQueue();
+		for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+			(*iter)->processEventQueue();
+		}
+
+		// If the _returnCode changed of a child gui we propagate it to the
+		// main gui
+		for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+			if((*iter)->_returnCode != _returnCode) {
+				_returnCode = (*iter)->_returnCode;
+			}
+		}
 
 		// Delay for a while
 		if (!EventMan.quitRequested() && (_returnCode != kReturnCodeNone))
@@ -184,6 +216,18 @@ void GUI::callbackRun() {
 }
 
 void GUI::callbackActive(Widget &UNUSED(widget)) {
+}
+
+
+void GUI::addChild(GUI* gui)
+{
+	_childrenguis.push_back(gui);
+}
+
+void GUI::removeChild(GUI * gui)
+{
+	gui->hide();
+	_childrenguis.remove(gui);
 }
 
 void GUI::addWidget(Widget *widget) {
--- a/src/engines/aurora/gui.cpp
+++ b/src/engines/aurora/gui.cpp
@@ -70,8 +70,7 @@ void GUI::show() {
 			widget.show();
 	}
 
-	std::list<GUI*>::iterator iter;
-	for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+	for (std::list<GUI *>::iterator iter = _childGUIs.begin(); iter != _childGUIs.end(); ++iter)	{
 		(*iter)->show();
 	}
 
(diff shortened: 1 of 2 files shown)


src/engines/aurora/gui.cpp
```diff
@@ -70,6 +70,11 @@ void GUI::show() {
 			widget.show();
 	}
 
+	std::list<GUI*>::iterator iter;
+	for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
```
DrMcCoy: Do ++iter, please

DrMcCoy: I guess you defined the iterator before the for-loop because of the line length? In general, I prefer to have the definition within the loop, to limit the scope of the variable

src/engines/aurora/gui.cpp
```diff
@@ -105,10 +122,25 @@ uint32 GUI::run(uint32 startCode) {
...
+				(*iter)->addEvent(event);
+			}
+		}
 
 		processEventQueue();
+		for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+			(*iter)->processEventQueue();
+		}
+
+		// If the _returnCode changed of a child gui we propagate it to the
```
DrMcCoy: If it's a multiline comment, use /* */

src/engines/aurora/gui.h
```diff
@@ -31,6 +31,7 @@
 #include <boost/noncopyable.hpp>
 
 #include "src/common/ustring.h"
+#include "src/common/scopedptr.h"
```
DrMcCoy: You're not using ScopedPtr, though

src/engines/aurora/gui.h
```diff
@@ -74,7 +75,6 @@ class GUI : boost::noncopyable {
 
 	GUI *_sub; ///< The currently running sub GUI.
 
-
```
DrMcCoy: Why? :P

src/engines/aurora/gui.h
```diff
@@ -115,10 +115,14 @@ class GUI : boost::noncopyable {
 	/** Callback that's triggered when a widget was activated. */
 	virtual void callbackActive(Widget &widget);
 
+	void addChild(GUI*);
+	void removeChild(GUI*);
 private:
 	typedef std::list<Widget *> WidgetList;
 	typedef std::map<Common::UString, Widget *> WidgetMap;
 
+	std::list<GUI*> _childrenguis;
```
DrMcCoy: Hmm, that should probably be rather named "_childGUIs".

child, because GUIs is already the plural (and you generally only plural the last part in a compund word). And "GUIs", because acronyms/initialisms in the xoreos codebase are always uppercased when it's the 2nd+ part in a camelCase word (see also, for example "showFPS").

src/engines/aurora/gui.cpp
```diff
@@ -105,10 +122,25 @@ uint32 GUI::run(uint32 startCode) {
...
+		// main gui
+		for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+			if((*iter)->_returnCode != _returnCode) {
```
DrMcCoy: `if (`, not `if(`

src/engines/aurora/gui.cpp
```diff
@@ -186,6 +218,18 @@ void GUI::callbackRun() {
 void GUI::callbackActive(Widget &UNUSED(widget)) {
 }
 
+
+void GUI::addChild(GUI* gui)
+{
+	_childrenguis.push_back(gui);
+}
+
+void GUI::removeChild(GUI * gui)
```
DrMcCoy: There's a space after the `*` that should be there

src/engines/aurora/gui.cpp
```diff
@@ -186,6 +218,18 @@ void GUI::callbackRun() {
 void GUI::callbackActive(Widget &UNUSED(widget)) {
 }
 
+
+void GUI::addChild(GUI* gui)
```
DrMcCoy: `GUI *gui`, not `GUI* gui`

src/engines/aurora/gui.h
```diff
@@ -115,10 +115,14 @@ class GUI : boost::noncopyable {
 	/** Callback that's triggered when a widget was activated. */
 	virtual void callbackActive(Widget &widget);
 
+	void addChild(GUI*);
```
DrMcCoy: `GUI *gui`

DrMcCoy: Also, you should probably add comments that these two methods should not be called in any of the GUI event handlers (because that will mess with the iterator in run()). And that the child GUIs will not have their ownership transferred to the parent GUI (i.e. the child GUIs still have to be deallocated as normal).

src/engines/aurora/gui.cpp
Nostritius: yes, and also because i can reuse the iterator over all children iterations.

src/engines/aurora/gui.h
Nostritius: because of the event handlers, i will look into it, so that a change can be made inside a callback. That is definitly needed when a event should add a new child GUI.

Nostritius:
I have now fixed it, that the addChild and removeChild methods can be called inside a callback and corrected (hopefully) all styling issues.

src/engines/aurora/gui.cpp
```diff
@@ -70,8 +70,7 @@ void GUI::show() {
 			widget.show();
 	}
 
-	std::list<GUI*>::iterator iter;
-	for (iter = _childrenguis.begin(); iter != _childrenguis.end(); iter++)	{
+	for (std::list<GUI*>::iterator iter = _childGUIs.begin(); iter != _childGUIs.end(); iter++)	{
```
DrMcCoy: You still have `iter++` instead of `++iter` in these loops

src/engines/aurora/gui.cpp
```diff
@@ -219,15 +219,16 @@ void GUI::callbackActive(Widget &UNUSED(widget)) {
 }
 
 
-void GUI::addChild(GUI* gui)
+void GUI::addChild(GUI *gui)
 {
```
DrMcCoy: The `{` belongs at the end of the previous line

src/engines/aurora/gui.cpp
```diff
@@ -219,15 +219,16 @@ void GUI::callbackActive(Widget &UNUSED(widget)) {
...
-	_childrenguis.push_back(gui);
+	_childGUIs.push_back(gui);
+	gui->show();
 }
 
-void GUI::removeChild(GUI * gui)
+void GUI::removeChild(GUI *gui)
 {
```
DrMcCoy: The `{` belongs at the end of the previous line

src/engines/aurora/gui.h
DrMcCoy: You still have `GUI*` instead of `GUI *gui` here

src/engines/aurora/gui.h
```diff
@@ -115,13 +115,15 @@ class GUI : boost::noncopyable {
 	/** Callback that's triggered when a widget was activated. */
 	virtual void callbackActive(Widget &widget);
 
+	/** Add a child GUI object to this GUI. Ownership odf the pointer is not transferred */
```
DrMcCoy: Typo: "odf"

src/engines/aurora/gui.cpp
DrMcCoy: Also, that `std::list<GUI*>` (here and everywhere else) should be `std::list<GUI *>`

DrMcCoy:
Merged it as b99751fec920b10c48aceaffc57bc70b149af5cd, thanks! :)

PR #138, opened 2017-05-25
GRAPHICS: Add an optional opengl scissor test for the quad rendering

Nostritius:
While working on the KotOR GUI, I recognized, that some textures were too large for placing them exactly like the original game does. The Problem was that they have borders equal with the background, but they overlap with other elements. I added the possiblity to remove this borders while rendering using the scissor test. I will add the gui specific stuff later.

### DIFF
--- a/src/graphics/aurora/guiquad.cpp
+++ b/src/graphics/aurora/guiquad.cpp
@@ -40,7 +40,7 @@ GUIQuad::GUIQuad(const Common::UString &texture,
 	_r(1.0f), _g(1.0f), _b(1.0f), _a(1.0f),
 	_x1 (x1) , _y1 (y1) , _x2 (x2) , _y2 (y2) ,
 	_tX1(tX1), _tY1(tY1), _tX2(tX2), _tY2(tY2),
-	_xor(false) {
+	_xor(false), _scissor(false) {
 
 	try {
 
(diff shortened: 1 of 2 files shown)


src/graphics/aurora/guiquad.cpp
```diff
@@ -140,6 +140,13 @@ void GUIQuad::setTexture(TextureHandle texture) {
 	unlockFrameIfVisible();
 }
 
+void GUIQuad::setScissor(int x, int y, int width, int height) {
+	_scissorx = x;
+	_scissory = y;
+	_scissorwidth = width;
+	_scissorheight = height;
```
DrMcCoy: These should be `_scissorX`, `_scissorY`, `_scissorWidth` and `_scissorHeight`.

src/graphics/aurora/guiquad.cpp
```diff
@@ -199,6 +214,18 @@ void GUIQuad::render(RenderPass pass) {
...
+	if (_scissor) {
+		glEnable(GL_SCISSOR_TEST);
+		GLint viewport[4];
+		glGetIntegerv(GL_VIEWPORT, viewport);
+		glScissor(
+				viewport[2]/2 + _x1 + _scissorx,
+				viewport[3]/2 + _y1 + _scissory,
+				_scissorwidth,
+				_scissorheight
+		);
```
DrMcCoy: Please use tabs for indenting, spaces for alignment. And put the first argument(s) on the same line as the function call itself.

I.e. it should rather looks like this:

```
		glScissor(viewport[2]/2 + _x1 + _scissorx,
		          viewport[3]/2 + _y1 + _scissory,
		          _scissorwidth,
		          _scissorheight);

```
Note that the tabs go to the level of glScissors, and then there's spaces to align the arguments to the opening parenthesis.

src/graphics/aurora/guiquad.h
```diff
@@ -71,6 +74,8 @@ class GUIQuad : public GUIElement {
 
 	void setXOR(bool enabled); ///< Enable/Disable XOR mode.
 
+	void setScissor(bool enabled); ///< Enable/Disable Scissor Test
```
DrMcCoy: There should be a full stop at the end of the comment there.

DrMcCoy:
Merged it as 1cf2577241a7042250e1f01bda1fa6b7e2077fb9, thanks! :)

PR #139, opened 2017-05-26
KOTOR: Fix a bug for widgets with an initially empty text

Nostritius:
The KotORWidget class does not generate a text rendering object if the text is initially empty. This caused the program to crash on attempts to change the text of empty labels. I fixed this by removing the condition that the text shouldn't be empty.


DrMcCoy:
Merged it as 28e413ddaa18fb14ad69974179c998f4c8c13384, thanks! :)

PR #140, opened 2017-06-03
GRAPHICS: Add callback to GUI for handling hovering

Nostritius:
While working on the KotOR character creation menus i was missing a possibility to identify widgets when they are hovered (the mouse is over the widget, but does not click). An example for this is the KotOR class selection menu, where you go over the six classes and get information about them in the text box below.

### DIFF
--- a/src/engines/aurora/gui.cpp
+++ b/src/engines/aurora/gui.cpp
@@ -217,6 +217,9 @@ void GUI::callbackRun() {
 void GUI::callbackActive(Widget &UNUSED(widget)) {
 }
 
+void GUI::callbackHover(Widget &UNUSED(widget)) {
+}
+
 
 void GUI::addChild(GUI *gui) {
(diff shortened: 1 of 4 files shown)


DrMcCoy:
Hmm, shouldn't Widget::enter() and Widget::leave() be exactly what you need? Those get triggered when the mouse enters and leaves a widget.

So the "hovered" state is equivalent to enter having been triggered. Or am I missing something?

(Also, I'm not sure your code there is even functional yet? Where are you setting the hovered state?)

Nostritius:
Yes, i plan to implement the State changes over enter and leave for specific widgets. Yes, it is equivalent when enter is triggered. I already tested the code, but removed it to add it later maybe if this request is accepted.

Nostritius:
I decided to take a different approach for solving this issue.

PR #141, opened 2017-06-04
KOTOR: Add some features for the KotOR Buttons

Nostritius:
when creating a character (either a quick or custom character) in KotOR, the buttons of the steps (like potrait, name ...) behave different then the default button in two ways. They are permanently highlighted, if they are the actual step and have no hovering sound. I added two methods for the KotOR Buttons to enable and disable this features.

### DIFF
--- a/src/engines/kotor/gui/widgets/button.h
+++ b/src/engines/kotor/gui/widgets/button.h
@@ -39,6 +39,7 @@ public:
 	~WidgetButton();
 
     void setPermanentHighlight(bool);
+	void setDisableHoverSound(bool);
 
 	virtual void load(const Aurora::GFF3Struct &gff);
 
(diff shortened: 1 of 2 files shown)


src/engines/kotor/gui/widgets/button.h
```diff
@@ -55,6 +64,6 @@ class WidgetButton : public KotORWidget {
 
 } // End of namespace KotOR
 
-} // End of namespace Engines
+} // End odnf search mate-x*f namespace Engines
```
DrMcCoy: This is wrong :P

DrMcCoy:
Can you split this into two commits, one for the permanent highlight, one for the sound disabling?

DrMcCoy:
And a third commit for the hover sound, please

Nostritius:
Ok, now i have divided the code and deleted the click sound disabling because i actually don't need it.

DrMcCoy:
Commited it as 048d8fab0...1f448039f, thanks! :)

PR #142, opened 2017-06-15
KOTOR: Character generation menu

Nostritius:
This is a simplified and yet extendable version of the KotOR character creation menu. At the moment you can only create a quick character. I disabled the custom character because most of the stuff, used there is practically not yet implemented. This Code should allow you to create a character and after that see the first level including the created player character standing on the starting point.
What is currently partly or not working:
- The 3d views of the characters
- The custom character creation. You can enter the menu, but the buttons have no functionality
- The buttons in the quick character menu have a glitch which makes the text invisible when disabling them
- The naked body of the characters, because they are weird to handle because of different skin colors
- The name menu is currently only a stub where you can neither create a random name nor create a custom name

### DIFF
--- a/src/engines/kotor/gui/chargen/chargeninfo.cpp
+++ b/src/engines/kotor/gui/chargen/chargeninfo.cpp
@@ -1,0 +1,175 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  The class for storing character information for generation.
+ */
+
+#include "src/common/strutil.h"
+#include "src/common/error.h"
+#include "chargeninfo.h"
+
+namespace Engines {
+
+namespace KotOR {
+
+CharacterGenerationInfo *CharacterGenerationInfo::createRandomMaleSoldier() {
+	CharacterGenerationInfo *info = new CharacterGenerationInfo();
+	info->_gender = kGenderMale;
+	info->_class = kClassSoldier;
+	info->_face = std::rand() % 5;
+	info->_skin = Skin(std::rand() % kSkinMAX);
+	info->_name = "";
+
+	return info;
+}
+
+CharacterGenerationInfo *CharacterGenerationInfo::createRandomMaleScout() {
+	CharacterGenerationInfo *info = new CharacterGenerationInfo();
+	info->_gender = kGenderMale;
+	info->_class = kClassScout;
+	info->_face = std::rand() % 5;
+	info->_skin = Skin(std::rand() % kSkinMAX);
+	info->_name = "";
+
+	return info;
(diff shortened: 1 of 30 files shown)


D8174:
You should do a find & replace for "Charachter" -> "Character"

src/engines/kotor/gui/chargen/classselection.cpp
```diff
@@ -0,0 +1,133 @@
...
+namespace KotOR {
+
+ClassSelectionMenu::ClassSelectionMenu(Module *module, ::Engines::Console *console) : GUI(console),
+	_module(module) {
+
+	load("classsel");
+
+	addBackground(kBackgroundTypeMenu);
+
+	// get the six class buttons
```
DrMcCoy: Upper case at the start of the comment, please.

src/engines/kotor/gui/chargen/classselection.cpp
```diff
@@ -0,0 +1,133 @@
...
+	// get the title label
+	_labelTitle = getLabel("LBL_CLASS");
+	_labelTitle->setText("");
+
+	// get the class descriptions
+	_soldierDesc = TalkMan.getString(32111);
+	_scoutDesc = TalkMan.getString(32110);
+	_scoundrelDesc = TalkMan.getString(32109);
+
+	//combine the titles with gender prefix and class name
```
DrMcCoy: A space after `//`, please.

src/engines/kotor/gui/chargen/classselection.cpp
```diff
@@ -0,0 +1,133 @@
...
+	femalePrefix = TalkMan.getString(359);
+	_soldierMaleTitle = malePrefix + " " + TalkMan.getString(134);
+	_soldierFemaleTitle = femalePrefix + " " + TalkMan.getString(134);
+	_scoutMaleTitle = malePrefix + " " + TalkMan.getString(133);
+	_scoutFemaleTitle = femalePrefix + " " + TalkMan.getString(133);
+	_scoundrelMaleTitle = malePrefix + " " + TalkMan.getString(135);
+	_scoundrelFemaleTitle = femalePrefix + " " + TalkMan.getString(135);
+}
+
+void ClassSelectionMenu::callbackRun() {
```
DrMcCoy: Hmm, doesn't that set/change the text every time the callback is called? I think it would be best if you remembered over which button the mouse is hovering and only set the text when the selection changes.

src/engines/kotor/gui/main/main.cpp
```diff
@@ -47,11 +51,21 @@ MainMenu::MainMenu(Module &module, bool isXbox, ::Engines::Console *console) : G
...
 	addBackground(kBackgroundTypeMenu);
+
+	startMainMusic();
 }
 
 MainMenu::~MainMenu() {
 }
 
+void MainMenu::createClassSelection()
+{
```
DrMcCoy: The { belongs on the previous line. For the other methods below as well.

src/engines/kotor/gui/main/main.cpp
```diff
@@ -113,14 +142,29 @@ void MainMenu::initWidget(Widget &widget) {
...
-		try {
+		/*try {
 			_module->load("end_m01aa");
 		} catch (...) {
 			Common::exceptionDispatcherWarning();
 			return;
 		}
 
-		_returnCode = 2;
+		_returnCode = 2;*/
```
DrMcCoy: Don't comment out code; just delete it.

src/engines/kotor/gui/chargen/chargeninfo.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+void CharachterGenerationInfo::setName(Common::UString name) {
+	_name = name;
+}
+
+void CharachterGenerationInfo::setSkin(Skin skin) {
+	_skin = skin;
+}
+
+void CharachterGenerationInfo::setFace(uint8_t face) {
```
DrMcCoy: Please use the type `uint8` instead of `uint8_t`, for consistency's sake.

Nostritius:
Ok, now i have (hopefully) everything corrected.

DrMcCoy:
You missed a few style things. :P

I took the liberty of fixing them and merging the PR as 0779ebd55...b52966e47 :)

PR #143, opened 2017-06-17
KOTOR: Add the nwscript function GetObjectByTag

Nostritius:
Add the nwscript function "GetObjectByTag" as a simple practice to the script system


DrMcCoy:
Pushed as 225e1ca24801c57efddedab92a7ad52e40f44459, thanks! :)

Could you check if that applies to KotOR2 as well and also send a PR for that then?

PR #144, opened 2017-06-17
GRAPHICS: Add a fade utility class to the engine

Nostritius:
I added a Class for handling Screen fades with one color. This is when the screen slowly turns from black to the game view or in the other direction. I also integrated support for this in KotOR


src/graphics/aurora/fadequad.cpp
```diff
@@ -0,0 +1,122 @@
...
+void FadeQuad::fadeOut() {
+	_fadeType = kFadeOut;
+	_start = EventMan.getTimestamp();
+	show();
+}
+
+void FadeQuad::calculateDistance() {
+}
+
+void FadeQuad::render(Graphics::RenderPass pass) {
```
DrMcCoy: You should probably make use of the RenderPass parameter here. Right now, the quad would be rendered twice each frame: one for the opaque objects render pass and then again for the transparent objects render pass.

Look at how src/engines/nwn/gui/legal.cpp does it.
(And in fact, maybe you can even make the NWN legal billboard use this class, then? It's not exactly the same, but maybe...)

As it is, the code currently throws a warning because the pass parameter is unused. Please wrap unused parameters in the UNUSED() macro.

src/graphics/aurora/fadequad.h
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  The Quad to handle one color fades.
+ */
+
+#ifndef XOREOS_FADEQUAD_H
```
DrMcCoy: The include guard should be GRAPHICS_AURORA_FADEQUAD_H

src/graphics/aurora/fadequad.h
```diff
@@ -0,0 +1,77 @@
...
+	uint32 _wait;
+	uint32 _run;
+	uint32 _start;
+};
+
+} // End of namespace Aurora
+
+} // End of namespace Graphics
+
+#endif //XOREOS_FADEQUAD_H
```
DrMcCoy: There should be a space after the `//` here (also, should be the correct name as above)

src/graphics/aurora/fadequad.h
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Graphics {
+
+namespace Aurora {
+
+class FadeQuad : public Renderable {
+public:
+	FadeQuad();
+
+	/** Get the current color of the quad */
```
DrMcCoy: There needs to be a period at the end of the sentence here and at the end of all the other doxygen comments.

src/graphics/aurora/fadequad.h
```diff
@@ -0,0 +1,77 @@
...
+	void getColor(float &r, float &g, float &b) const;
+	/** Set the current color of the quad. */
+	void setColor(float r, float g, float b);
+
+	/** Get and set the time to wait until fading begins */
+	void getWaitTime(float&) const;
```
DrMcCoy: `float &`, with a space, and please also add the parameter's name.

src/graphics/aurora/fadequad.cpp
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  The Quad to handle one color fades.
+ */
+
+#include <cfloat>
```
DrMcCoy: You should include "src/common/maths.h" instead.

src/graphics/aurora/fadequad.cpp
```diff
@@ -0,0 +1,122 @@
...
+	b = _b;
+}
+
+void FadeQuad::setColor(float r, float g, float b) {
+	_r = r;
+	_g = g;
+	_b = b;
+}
+
+void FadeQuad::getWaitTime(float & wait) const {
```
DrMcCoy: `float &wait`

src/graphics/aurora/fadequad.cpp
```diff
@@ -0,0 +1,122 @@
...
+	wait = _wait/1000;
+}
+
+void FadeQuad::setWaitTime(float wait) {
+	_wait = wait*1000;
+}
+
+void FadeQuad::getRunTime(float & run) const {
```
DrMcCoy: `float &run`

src/graphics/aurora/fadequad.h
```diff
@@ -0,0 +1,77 @@
...
+	void setWaitTime(float);
+
+	/** Get and set the time the quad is fading */
+	void getRunTime(float&) const;
```
DrMcCoy: `float &`, with a space, and please also add the parameter's name.

src/graphics/aurora/fadequad.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+void FadeQuad::setColor(float r, float g, float b) {
+	_r = r;
+	_g = g;
+	_b = b;
+}
+
+void FadeQuad::getWaitTime(float & wait) const {
+	wait = _wait/1000;
```
DrMcCoy: Please add spaces around the operator here and in the other cases below. I.e. `wait / 1000`.

src/graphics/aurora/fadequad.cpp
```diff
@@ -0,0 +1,122 @@
...
+		time = time - _wait;
+
+		if (_run == 0.0f) {
+			if (_fadeType == kFadeOut)
+				_opacity = 1.0f;
+			if (_fadeType == kFadeIn)
+				_opacity = 0.0f;
+		} else {
+			if (_fadeType == kFadeOut)
+				_opacity = std::min(1.0f, static_cast<float>(time) / _run);
```
DrMcCoy: We have our own MIN()/MAX() templates in src/common/util.h.

src/engines/kotor/module.h
```diff
@@ -85,6 +87,8 @@ class Module : public KotOR::Object, public KotOR::ObjectContainer {
 	Area *getCurrentArea();
 	/** Return the currently playing PC. */
 	Creature *getPC();
+	/** Return the fade quad */
+	Graphics::Aurora::FadeQuad *getFadeQuad();
```
DrMcCoy: Another missing period at the end of the doxygen comment there. :P

Also, you should make that method return a reference instead of a pointer.

Nostritius:
Ok, i hope i have corrected everything in this update

Nostritius:
Ok, everything is now corrected and split up into two commits.

DrMcCoy:
Merged as 3d48f5b00...85d827342, thanks! :)

PR #149, opened 2017-07-22
KOTOR: Widget borders

Nostritius:
I have investigated the missing borders around some KotOR widgets and found out, that they are defined in the .gui file. I implemented a new class for them named BorderQuad and integrated it into the KotORWidget class.

### DIFF
--- a/src/engines/kotor/gui/widgets/kotorwidget.cpp
+++ b/src/engines/kotor/gui/widgets/kotorwidget.cpp
@@ -74,12 +74,16 @@ void KotORWidget::show() {
 		_quad->show();
 	if (_text)
 		_text->show();
+	if (_border)
+		_border->show();
 }
 
 void KotORWidget::hide() {
 	if (isInvisible())
 		return;
 
+	if (_border)
+		_border->hide();
 	if (_quad)
 		_quad->hide();
 	if (_text)
@@ -120,6 +124,10 @@ void KotORWidget::setPosition(float x, float y, float z) {
 		_text->getPosition(tX, tY, tZ);
 
 		_text->setPosition(tX - oX + x, tY - oY + y, tZ - oZ + z);
+	}
+
+	if (_border) {
+		_border->setPosition(x, y, z);
 	}
 }
 
@@ -179,6 +187,10 @@ void KotORWidget::load(const Aurora::GFF3Struct &gff) {
 
 	if (border.fill.empty())
 		_quad->setColor(0.0f, 0.0f, 0.0f, 0.0f);
+
+	if (!border.edge.empty() && !border.corner.empty()) {
+		_border.reset(new Graphics::Aurora::BorderQuad(border.edge, border.corner, extend.x, extend.y, extend.w, extend.h));
+	}
 
 	Text text = createText(gff);
 
--- a/src/graphics/aurora/borderquad.cpp
+++ b/src/graphics/aurora/borderquad.cpp
@@ -1,0 +1,190 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file
+ * A quad for generating borders.
+ */
+
+#include <cfloat>
+
+#include "borderquad.h"
+#include "textureman.h"
+
+namespace Graphics {
+
+namespace Aurora {
+
+BorderQuad::BorderQuad(const Common::UString &edge, const Common::UString &corner, float x, float y, float w, float h)
+		: Renderable(kRenderableTypeGUIFront),
+		  _edgeWidth(0), _edgeHeight(0), _cornerWidth(0), _cornerHeight(0),
+		  _x(x), _y(y), _w(w), _h(h) {
+	_edge = TextureMan.get(edge);
+	_corner = TextureMan.get(corner);
+
+	_cornerWidth = _corner.getTexture().getWidth();
+	_cornerHeight = _corner.getTexture().getHeight();
+	_edgeWidth = _edge.getTexture().getWidth();
+	_edgeHeight = _edge.getTexture().getHeight();
+
+	_distance = -FLT_MAX;
+
+	assert(!_corner.empty());
+	assert(!_edge.empty());
+}
+
+void BorderQuad::setPosition(float x, float y, float UNUSED(z)) {
+	lockFrameIfVisible();
+
+	/*
+	 * Because of some graphic glitches with non-integer floats, i have to
+	 * floor the values
+	 */
+	_x = std::floor(x);
+	_y = std::floor(y);
+
+	unlockFrameIfVisible();
+}
+
+void BorderQuad::getPosition(float &x, float &y, float &z) {
+	x = _x;
+	y = _y;
+	z = -FLT_MAX;
+}
+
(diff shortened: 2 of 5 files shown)


src/engines/kotor/gui/widgets/kotorwidget.cpp
```diff
@@ -74,12 +74,16 @@ void KotORWidget::show() {
...
 		_text->show();
+    if (_border)
+        _border->show();
 }
 
 void KotORWidget::hide() {
 	if (isInvisible())
 		return;
 
+    if (_border)
```
tdalman: Indentation level doesn't look correct.

src/engines/kotor/gui/widgets/kotorwidget.cpp
```diff
@@ -121,6 +125,13 @@ void KotORWidget::setPosition(float x, float y, float z) {
 
 		_text->setPosition(tX - oX + x, tY - oY + y, tZ - oZ + z);
 	}
+
+    if (_border) {
+        float bX, bY, bZ;
+		_border->getPosition(bX, bY, bZ);
+
+        _border->setPosition(x, y, z);
```
tdalman: _border->getPosition() has a side-effect, a comment would be in order. If it does not have a side-effect (i.e., it is pure), _border->getPosition() and bX, bY, and bZ are superfluous.

src/engines/kotor/gui/widgets/kotorwidget.cpp
```diff
@@ -121,6 +125,13 @@ void KotORWidget::setPosition(float x, float y, float z) {
 
 		_text->setPosition(tX - oX + x, tY - oY + y, tZ - oZ + z);
 	}
+
+    if (_border) {
+        float bX, bY, bZ;
+		_border->getPosition(bX, bY, bZ);
```
tdalman: Indentation looks funny again.

src/graphics/aurora/borderquad.cpp
```diff
@@ -0,0 +1,188 @@
...
+          _x(x), _y(y), _w(w), _h(h) {
+	_edge = TextureMan.get(edge);
+	_corner = TextureMan.get(corner);
+
+    _cornerWidth = _corner.getTexture().getWidth();
+    _cornerHeight = _corner.getTexture().getHeight();
+    _edgeWidth = _edge.getTexture().getWidth();
+    _edgeHeight = _edge.getTexture().getHeight();
+
+    _distance = -FLT_MAX;
```
tdalman: FLT_MIN? Also, is there a reason not use std::numeric_limits<float>?

src/graphics/aurora/borderquad.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    assert(!_corner.empty());
+    assert(!_edge.empty());
+}
+
+void BorderQuad::setPosition(float x, float y, float UNUSED(z)) {
+    lockFrameIfVisible();
+
+    _x = (int)x;
+    _y = (int)y;
```
tdalman: Can you leave a comment why you are casting to int? Also, why not use static_cast<>?

src/graphics/aurora/borderquad.h
```diff
@@ -0,0 +1,66 @@
...
+
+	virtual void calculateDistance();
+
+	void render(RenderPass pass);
+
+private:
+	TextureHandle _edge, _corner;
+
+    int _edgeWidth, _edgeHeight;
+    int _cornerWidth, _cornerHeight;
```
tdalman: Indentation.

src/graphics/aurora/rules.mk
```diff
@@ -49,6 +49,7 @@ src_graphics_aurora_libaurora_la_SOURCES += \
     src/graphics/aurora/animnode.h \
     src/graphics/aurora/animation.h \
     src/graphics/aurora/fadequad.h \
+    src/graphics/aurora/borderquad.h \
```
tdalman: This seems to violate the previous ordering of filenames (lexicographical).

src/graphics/aurora/borderquad.cpp
Nostritius: Well, simply because we did it this way in a lot of other parts of the engine.

src/graphics/aurora/borderquad.cpp
Nostritius: I will definitely leave a comment. I recognized a bug which occured when we have some x and y which are non-integer values. so i used this to create a whole number from it. Maybe something like std::floor would be handier and more understandable.

src/engines/kotor/gui/widgets/kotorwidget.cpp
Nostritius: About the Indentation, I recognized, that my IDE inserted some tabs while i was not looking. I will definitely correct those.

src/graphics/aurora/rules.mk
Nostritius: We have to my knowledge no convention for ordering the rules.mk files.

src/graphics/aurora/borderquad.cpp
tdalman: If that's the idiom used everywhere, this code shouldn't be changed then (sorry, I'm rather new and trying to understand).

src/graphics/aurora/rules.mk
DrMcCoy: Yeah, the ordering is not really lexicographical. When I added files, I've roughly put them from least specific to more specific, if that makes sense. Base classes before classes that inherit from those. for example.

But that's not really a hard rule. That rule breaks down for some cases, and other people added files in other orders. I'm not really that adamant about it. Not about this, at least ;)

src/graphics/aurora/borderquad.cpp
```diff
@@ -0,0 +1,190 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file
+ * A quad for generating borders
```
DrMcCoy: Doxygen comment should end with a period here. Also in the .h file.

src/graphics/aurora/borderquad.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include <cfloat>
+
+#include "borderquad.h"
+#include "textureman.h"
+
+namespace Graphics {
+
+namespace Aurora {
+
+BorderQuad::BorderQuad(Common::UString edge, Common::UString corner, float x, float y, float w, float h)
```
DrMcCoy: Those two Common::UString can be const references.

src/graphics/aurora/borderquad.cpp
```diff
@@ -0,0 +1,190 @@
...
+	_distance = -FLT_MAX;
+
+	assert(!_corner.empty());
+	assert(!_edge.empty());
+}
+
+void BorderQuad::setPosition(float x, float y, float UNUSED(z)) {
+	lockFrameIfVisible();
+
+	/**
```
DrMcCoy: That shouldn't be a Doxygen comment. I.e. remove the second `*`.

src/graphics/aurora/borderquad.cpp
```diff
@@ -0,0 +1,190 @@
...
+	glTexCoord2f(0, 1);
+	glVertex2f(_x + _w - _cornerWidth, _y + _h);
+	glTexCoord2f(1, 1);
+	glVertex2f(_x + _cornerWidth, _y + _h);
+	glEnd();
+}
+
+}
+
+}
```
DrMcCoy: These two `}` are missing the "End of namespace" comments.

src/engines/kotor/gui/widgets/kotorwidget.cpp
```diff
@@ -180,11 +188,15 @@ void KotORWidget::load(const Aurora::GFF3Struct &gff) {
...
+	if (!border.edge.empty() && !border.corner.empty()) {
+		_border.reset(new Graphics::Aurora::BorderQuad(border.edge, border.corner, extend.x, extend.y, extend.w, extend.h));
+	}
+
 	Text text = createText(gff);
 
 	if (!text.font.empty()) {
 		_text.reset(new Graphics::Aurora::HighlightableText(FontMan.get(text.font),
-		            text.text, text.r, text.g, text.b, 1.0f));
+					text.text, text.r, text.g, text.b, 1.0f));
```
DrMcCoy: You replaced the alignment space with tabs here. Don't. :P

Nostritius:
Ok, this should now have corrected the issues

DrMcCoy:
Merged it as 15a4b20ad...a1bea660c, thanks! :)

PR #150, opened 2017-07-30
KOTOR: Add checkbox toggle

fdde:


fdde:
I realised that this isn't adequate after looking at the code for nwn's checkboxes.

PR #151, opened 2017-07-30
KOTOR: Add slider moving

fdde:
I'm not sure if this is exactly how KotOR handles sliders but it's probably better than nothing.


fdde:
Not at all how KotOR handles sliders. :)

PR #152, opened 2017-08-02
KOTOR: Fix font bug in chargenname gui

Nostritius:
The chargenname gui uses a font, which is somehow corrupted. I added the possiblity to change a widgets font and changed it inside the name generation gui.

### DIFF
--- a/src/engines/kotor/gui/widgets/kotorwidget.cpp
+++ b/src/engines/kotor/gui/widgets/kotorwidget.cpp
@@ -136,6 +136,10 @@ float KotORWidget::getHeight() const {
 	return _height;
 }
 
+void KotORWidget::setFont(const Common::UString &fnt) {
+	_text->setFont(fnt);
+}
+
 void KotORWidget::setFill(const Common::UString &fill) {
 	if (!_quad) {
 		float x, y, z;
--- a/src/engines/kotor/gui/widgets/kotorwidget.h
+++ b/src/engines/kotor/gui/widgets/kotorwidget.h
@@ -60,6 +60,9 @@ public:
 	/** Create a scissor test over this widget. */
 	void setScissor(int x, int y, int width, int height);
 
+	/** Change the font for this widget. */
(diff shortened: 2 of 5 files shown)


src/graphics/aurora/text.h
```diff
@@ -54,6 +54,9 @@ class Text : public GUIElement {
 	void unsetColor();
 	void setAlign(float align);
 
+	/** Change the font of the text */
```
DrMcCoy: The comment should end with a period.

src/graphics/aurora/text.cpp
```diff
@@ -242,6 +243,10 @@ void Text::parseColors(const Common::UString &str, Common::UString &parsed,
 	}
 }
 
+void Text::setFont(Common::UString fnt) {
```
DrMcCoy: You're not changing the string inside the method, so it can be a const reference, i.e. `const Common::UString &`.

src/engines/kotor/gui/widgets/kotorwidget.cpp
```diff
@@ -136,6 +136,10 @@ float KotORWidget::getHeight() const {
 	return _height;
 }
 
+void KotORWidget::setFont(Common::UString fnt) {
```
DrMcCoy: Likewise, make that a const reference.

src/engines/kotor/gui/widgets/kotorwidget.h
```diff
@@ -60,6 +60,9 @@ class KotORWidget : public Widget {
 	/** Create a scissor test over this widget. */
 	void setScissor(int x, int y, int width, int height);
 
+	/** Change the font for this widget */
```
DrMcCoy: Likewise, comment should end with a period.

Nostritius:
This Pull Request should now be corrected regarding the requested changes

DrMcCoy:
Merged it as b75a4d7d3...0e9b18099, thanks! :)

PR #153, opened 2017-08-02
AURORA: Add a text input callback to the gui class

Nostritius:
Adds a simple solution for gui wide text input over a callback. For deleting characters i thought about another callback for keys which could also be used for ingame shortcuts.

### DIFF
--- a/src/engines/aurora/gui.cpp
+++ b/src/engines/aurora/gui.cpp
@@ -200,6 +200,8 @@ uint32 GUI::processEventQueue() {
 			mouseUp(*e);
 		else if (e->type == Events::kEventMouseWheel)
 			mouseWheel(*e);
+		else if (e->type == Events::kEventTextInput)
+			textInput(*e);
 	}
 
 	_eventQueue.clear();
@@ -217,6 +219,8 @@ void GUI::callbackRun() {
 void GUI::callbackActive(Widget &UNUSED(widget)) {
 }
 
+void GUI::callbackTextInput(const Common::UString &UNUSED(text)) {
+}
 
 void GUI::addChild(GUI *gui) {
 	_childGUIs.push_back(gui);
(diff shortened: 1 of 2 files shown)


src/engines/aurora/gui.h
```diff
@@ -114,6 +114,8 @@ class GUI : boost::noncopyable {
 	virtual void callbackRun();
 	/** Callback that's triggered when a widget was activated. */
 	virtual void callbackActive(Widget &widget);
+	/** Callback that's triggered when a text input is received */
```
DrMcCoy: The Doxygen comment should end with a period.

src/engines/aurora/gui.h
```diff
@@ -114,6 +114,8 @@ class GUI : boost::noncopyable {
...
+	virtual void callbackTextInput(Common::UString text);
```
DrMcCoy: Again, this should probably be a const reference. The callback shouldn't usually need to modify the text input at all.

src/engines/aurora/gui.cpp
```diff
@@ -508,6 +512,10 @@ void GUI::mouseWheel(const Events::Event &event) {
 	mouseWheel(_currentWidget, event);
 }
 
+void GUI::textInput(const Events::Event &event) {
```
DrMcCoy: P

src/engines/aurora/gui.cpp
```diff
@@ -217,6 +219,8 @@ void GUI::callbackRun() {
 void GUI::callbackActive(Widget &UNUSED(widget)) {
 }
 
+void GUI::callbackTextInput(Common::UString text) {
```
DrMcCoy: Please wrap the `text` in the `UNUSED()` macro here, since the parameter isn't used in this method.

DrMcCoy:
Again, except for a few style issues, this PR looks good to me. :)

Nostritius:
Ok, this should now have corrected the style issues

DrMcCoy:
Merged it as f9b464194, thanks! :)

PR #154, opened 2017-08-04
CMAKE: Add some lines for a make install

Nostritius:
Add some lines for installing the xoreos executable into ${CMAKE_INSTALL_PREFIX}/bin on unix based systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -452,3 +452,12 @@ if(XOREOS_BUILD_DOCUMENTATION)
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
   )
 endif()
+
+# -------------------------------------------------------------------------
+# installation
+if(UNIX)
+  include(GNUInstallDirs)
+  install(TARGETS xoreos
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}
+  )
+endif()


CMakeLists.txt
```diff
@@ -452,3 +452,12 @@ if(XOREOS_BUILD_DOCUMENTATION)
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
   )
 endif()
+
+# -------------------------------------------------------------------------
+# installation
+if(UNIX)
+  include(GNUInstallDirs)
```
tdalman: This is already included above. Not needed I think.

Otherwise ok.

DrMcCoy:
Hmm, that only installs the one xoreos binary, no? I'd really want it to install everything that the autotools "make install" installs, namely the manpage and the documentation.

Also, it would be great if it could re-use, for the binary at least, the information parsed from the automake files. It already uses them to generate the list of binaries to build, anyway. This way, installation could be easily added to xoreos-tools as well, without having to list all xoreos-tools binaries themselves.

Maybe you can even extend the automake parsing to also grab the documentation files and the manpages. They are stated in the automake files, after all. Especially for xoreos-tools with its many manpages that would be vital. Having this all stated for each build system just means that there's an easy way for someone to forget updating one of the lists.

(Minor thing: the subsystem tag in the commit message for CMake too should be "BUILD:", and then say in the first line that you're adding make install to CMake.)

Nostritius:
I think, i will do this another time completely new.

PR #155, opened 2017-08-20
GRAPHICS: Make Highlightable framerate independent

seedhartha:
Menu buttons in KotOR were blinking way too fast on newer hardware. Made it framerate independent.

### DIFF
--- a/src/graphics/aurora/highlightable.cpp
+++ b/src/graphics/aurora/highlightable.cpp
@@ -18,6 +18,7 @@
  * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "src/events/events.h"
 #include "src/graphics/aurora/highlightable.h"
 
 namespace Graphics {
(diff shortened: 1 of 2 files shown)


src/graphics/aurora/highlightable.cpp
```diff
@@ -102,6 +107,8 @@ void Highlightable::incrementColor(float initialR, float initialG, float initial
 		b = initialB;
 		a = initialA;
 	}
+
```
DrMcCoy: Trailing white space

src/graphics/aurora/highlightable.h
```diff
@@ -76,6 +78,8 @@ class Highlightable {
 	float _lowerBoundG;
 	float _lowerBoundB;
 	float _lowerBoundA;
+
```
DrMcCoy: Trailing white space

src/graphics/aurora/highlightable.cpp
```diff
@@ -88,10 +90,13 @@ void Highlightable::flipHighlightDelta() {
 
 void Highlightable::incrementColor(float initialR, float initialG, float initialB, float initialA,
                                    float &r, float &g, float &b, float &a) {
-	r = initialR + _deltaR;
-	g = initialG + _deltaG;
-	b = initialB + _deltaB;
-	a = initialA + _deltaA;
+	uint32 time = EventMan.getTimestamp();
+	double dt = (time - _prevIncTime) / 50.f;
```
DrMcCoy: Why do you use float constant, but assign the calculation to a double?

src/graphics/aurora/highlightable.h
seedhartha: I believe it looks better that way since there are already four groups of members separated by white spaces.
private_members
As for other issues, will fix those right away :)

DrMcCoy: No, the problem is not the empty line, that is good and wanted. But the line should be completely empty. Right now, there's still a (invisible) tab character in that line. Make the line completely empty by removing the tab.

DrMcCoy: Also, to catch these, you could configure your editor/IDE to show trailing whitespace (i.e. tabs and spaces at the end of lines) and/or show all white space.

This is how that section looks for me:
foo

PR #157, opened 2017-08-21
BUILD: Update NASM path for AppveyorCI builds

berenm:
Looks like NASM is now a 64bit application, installed in "Program Files" folder.


DrMcCoy:
Yay, thanks! :)

PR #158, opened 2017-08-22
KOTOR: Modify perspective with KotOR values

seedhartha:
Models were partially cut at close distances unlike in original game. Also fixes background rendering bug in module korr_m36aa. These values (FOV 55.0, near distance 0.1, far distance 10000.0) I got directly from KotOR executable.


DrMcCoy:
Merged it as b75a4d7d3aec29ef286c91ea05b376eb2c804068, thanks! :)

PR #159, opened 2017-08-25
GRAPHICS: Implement KotOR animations

seedhartha:
Mostly working KotOR animations implementation. Uses CPU for skeletal animation. There are some unhandled model node controller types (namely, compressed quaternions).

### DIFF
--- a/src/graphics/aurora/model.h
+++ b/src/graphics/aurora/model.h
@@ -74,6 +74,9 @@ public:
 	/** Change the environment map on this model. */
 	void setEnvironmentMap(const Common::UString &environmentMap = "");
 
+	/** Set the flag if the model has skinned animations. */
+	void setSkinned(bool skinned);
+
 	/** Is that point within the model's bounding box? */
 	bool isIn(float x, float y) const;
 	/** Is that point within the model's bounding box? */
@@ -134,6 +137,10 @@ public:
 	ModelNode *getNode(const Common::UString &stateName, const Common::UString &node);
 	/** Get the specified node, from the named state, if it exists. */
 	const ModelNode *getNode(const Common::UString &stateName, const Common::UString &node) const;
+	/** Get node by node number. */
+	ModelNode *getNode(uint16 nodeNumber);
+	/** Get node by node number. */
+	const ModelNode *getNode(uint16 nodeNumber) const;
 
 	/** Get all nodes in the current state. */
 	const std::list<ModelNode *> &getNodes();
@@ -244,6 +251,8 @@ protected:
 	/** The model's box after translate/rotate. */
 	Common::BoundingBox _absoluteBoundBox;
 
+	bool _skinned;
+
 
 	// Rendering
 
@@ -279,6 +288,7 @@ private:
 	float _animationLoopLength; ///< The length of one loop of the current animation.
 	float _animationLoopTime;   ///< The time the current loop of the current animation has played.
 
+	std::vector<ModelNode *> _animNodeMap;
 
 	/** Create the list of all state names. */
 	void createStateNamesList(std::list<Common::UString> *stateNames = 0);
@@ -307,6 +317,7 @@ public:
 	                      uint32 offset, uint32 count, std::vector<T> &values);
 
 	friend class ModelNode;
+	friend class Animation;
 };
 
 } // End of namespace Aurora
--- a/src/graphics/aurora/model_kotor.cpp
+++ b/src/graphics/aurora/model_kotor.cpp
@@ -107,7 +107,8 @@ namespace Aurora {
 
 Model_KotOR::ParserContext::ParserContext(const Common::UString &name,
                                           const Common::UString &t, bool k2) :
-	mdl(0), mdx(0), state(0), texture(t), kotor2(k2) {
+	mdl(0), mdx(0), state(0), texture(t), kotor2(k2), mdxStructSize(0), vertexCount(0),
+	offNodeData(0) {
 
 	try {
 
@@ -149,6 +150,9 @@ Model_KotOR::Model_KotOR(const Common::UString &name, bool kotor2, ModelType typ
 	ParserContext ctx(name, texture, kotor2);
 
 	load(ctx);
+
+	if (_skinned)
+		makeBoneNodeMap();
 
 	loadSuperModel(modelCache, kotor2);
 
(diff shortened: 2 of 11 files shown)


DrMcCoy:
Hej, just wanted to say that I haven't forgotten your PR. I haven't really had enough time at a stretch to properly look at it, sorry.

Still, thank you for your work here! :)

src/graphics/aurora/modelnode.cpp
```diff
@@ -162,13 +182,17 @@ void ModelNode::getAbsolutePosition(float &x, float &y, float &z) const {
 	z = _absolutePosition.getZ() * _model->_scale[2];
 }
 
-Common::Matrix4x4 ModelNode::getAsolutePosition() const {
+Common::Matrix4x4 ModelNode::getAbsolutePosition() const {
```
farmboy0: This should be in a seperate commit

src/graphics/aurora/modelnode.h
```diff
@@ -202,6 +218,8 @@ class ModelNode {
 	Common::BoundingBox _boundBox;
 	Common::BoundingBox _absoluteBoundBox;
 
+	uint16 _nodeNumber;
```
farmboy0: _nodeNumber isnt initialized in the constructor.

src/graphics/aurora/model_kotor.cpp
```diff
@@ -650,6 +744,51 @@ void ModelNode_KotOR::readMesh(Model_KotOR::ParserContext &ctx) {
...
+	_mesh->skin = new Skin();
+	_mesh->skin->boneMapping = new float[boneMappingCount];
+	_mesh->skin->boneMappingCount = boneMappingCount;
+	_mesh->skin->boneWeights = new float[_vertexCount * 4];
+	_mesh->skin->boneMappingId = new float[_vertexCount * 4];
+
+	uint32 pos = ctx.mdl->seek(ctx.offModelData + boneMappingOffset);
+	for (uint32 i = 0; i < boneMappingCount; i++)
+		_mesh->skin->boneMapping[i] = ctx.mdl->readIEEEFloatLE();
+	ctx.mdl->seek(pos);
```
farmboy0: This should jump to the end of the skin data but not only does it ignore the bone mappings it also ignores all the rest of the skin data whose definition you ignore in line 759.

src/graphics/aurora/model_kotor.h
```diff
@@ -93,9 +95,14 @@ class ModelNode_KotOR : public ModelNode {
 	void load(Model_KotOR::ParserContext &ctx);
 
 private:
+	uint32 _mdxStructSize;
+	uint16 _vertexCount;
+	uint32 _offNodeData;
```
farmboy0: These arent used past the parsing of the model and move to the ParserContext

src/graphics/aurora/modelnode.h
```diff
@@ -129,9 +130,20 @@ class ModelNode {
 		Dangly();
 	};
 
+	struct Skin {
+		float      *boneMapping;
+		uint32      boneMappingCount;
+		float      *boneWeights;
+		float      *boneMappingId;
+		ModelNode **boneNodeMap;
```
DrMcCoy: Could you convert these arrays to `std::vector`s, please?

In general, I'd like to keep explicit new/delete to a minimum, by either using STL containers or ScopedPtr/ScopedArray.

src/graphics/aurora/model_kotor.cpp
```diff
@@ -472,10 +506,64 @@ void ModelNode_KotOR::load(Model_KotOR::ParserContext &ctx) {
...
+		uint32 type = ctx.mdl->readUint32LE();
+		ctx.mdl->skip(2);
+		uint16 rowCount = ctx.mdl->readUint16LE();
+		uint16 timeIndex = ctx.mdl->readUint16LE();
+		uint16 dataIndex = ctx.mdl->readUint16LE();
+		uint8 columnCount = ctx.mdl->readByte();
+		ctx.mdl->skip(3);
+		switch (type) {
+			case kControllerTypePosition:
+				switch (columnCount) {
```
farmboy0: I think the inner switches should move to their own methods. Too much intentation here to make code easily readable.

src/graphics/aurora/model.h
```diff
@@ -177,6 +177,8 @@ class Model : public GLContainer, public Renderable {
 	void render(RenderPass pass);
 	void advanceTime(float dt);
 
+	bool _isSkinned;
```
farmboy0: This flag shouldn't be public.

src/graphics/aurora/model_kotor.cpp
```diff
@@ -387,6 +419,8 @@ void ModelNode_KotOR::load(Model_KotOR::ParserContext &ctx) {
 	uint16 superNode  = ctx.mdl->readUint16LE();
 	uint16 nodeNumber = ctx.mdl->readUint16LE();
 
+	_nodeNumber = nodeNumber;
```
farmboy0: This can be set and used directly

src/graphics/aurora/model_kotor.cpp
farmboy0: This should go back to the position after the skin header(where you are after line 759) . You cant assume that the start of the bone mapping matches the end of the skin header.

src/graphics/aurora/modelnode.cpp
seedhartha: Fixed

src/graphics/aurora/modelnode.h
seedhartha: Fixed

src/graphics/aurora/model_kotor.cpp
seedhartha: Not sure I understand you there. It doesn't read bone data from the end of the skin header. First it moves to the offset specified in skin header, reads the data, then returns to previous position.

src/graphics/aurora/model_kotor.cpp
seedhartha: Fixed

src/graphics/aurora/model.h
seedhartha: Made it protected

src/graphics/aurora/model_kotor.cpp
seedhartha: Fixed

src/graphics/aurora/modelnode.h
seedhartha: Fixed

src/graphics/aurora/model_kotor.h
seedhartha: Fixed

src/graphics/aurora/model_kotor.cpp
farmboy0: The previous position you go back to is the start of the bone data.
The previous position you should go back is the end of the skin header.

seedhartha: According to the seek method description and your own binary template I'm pretty sure it goes back to the end of skin header as is.

farmboy0: You're right. I didn't realize seek returns the previous position before the seek.

seedhartha:
All requested changes made. I had however accidentally pushed animation testing code if that's a problem.

farmboy0:
Theres 2 things left to do:
1. use std::vector<float> instead of the float arrays or float * as DrMcCoy asked
2. the debugging code needs to go

seedhartha:
Okay, that should be it.

DrMcCoy:
Okay, merged it, thanks! :)

PR #160, opened 2017-08-30
KOTOR: Add the typing possibility in the CharGenName class

Nostritius:
This PR adds an implemented callbackTextInput() for the CharGenName class. Now it is possible to type the character name with a maximum of 18 letters (this is the maximum of the original game).

### DIFF
--- a/src/engines/kotor/gui/chargen/chargenname.cpp
+++ b/src/engines/kotor/gui/chargen/chargenname.cpp
@@ -22,10 +22,13 @@
  *  The menu for modifying the name of the character.
  */
 
+#include "src/events/events.h"
+
 #include "src/engines/kotor/gui/widgets/kotorwidget.h"
 #include "src/engines/kotor/gui/widgets/label.h"
 
 #include "src/engines/kotor/gui/chargen/chargenname.h"
+#include "src/engines/kotor/gui/chargen/chargeninfo.h"
 
 namespace Engines {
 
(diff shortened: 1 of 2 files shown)


src/engines/kotor/gui/chargen/chargenname.cpp
```diff
@@ -36,23 +39,41 @@ CharacterGenerationNameMenu::CharacterGenerationNameMenu(CharacterGenerationInfo
 
 	load("name");
 
+	_name = "";
```
DrMcCoy: That isn't necessary. A Common::UString is initialized with an empty string by default.

src/engines/kotor/gui/chargen/chargenname.cpp
```diff
@@ -36,23 +39,41 @@ CharacterGenerationNameMenu::CharacterGenerationNameMenu(CharacterGenerationInfo
...
+
+void CharacterGenerationNameMenu::callbackTextInput(const Common::UString &text) {
+	// The name should not be longer than 18 letters (according to the original game).
+	if (_name.size() == 18) {
 		return;
 	}
+
+	_name += text;
+	_info.setName(_name);
+	_nameLabel->setText(_name + "_");
```
DrMcCoy: That is missing the ability to delete characters again with backspace, though.

I guess you also need a callbackKeyDown() for that.

Nostritius: Yes, this is already planned, but for now I assumed, it would be enough to enable typing. At I have to design a new callback for keys, but this was planned for another pull request.

DrMcCoy: Eh, do that other pull request first, please, and then revisit this pull request. I think it would be best for this PR to be more feature-complete before merging it. :)

Nostritius:

DrMcCoy:
Errr, what's the reason for the timer?

If that's meant for key repeat, should EventsManager::enableKeyRepeat() be the better way there?

Nostritius:
we have a keyRepeat? Nice, this is obviously way better than my timer solution.

DrMcCoy:
Merged it as 0f5581e02e193479e4542c6084e58016a1648b5c, thanks! :)

PR #170, opened 2017-09-05
AURORA: Add keyboard input callback for GUI class

Nostritius:
Adds a callback to GUI classes for receiving keyboard inputs.

### DIFF
--- a/src/engines/aurora/gui.cpp
+++ b/src/engines/aurora/gui.cpp
@@ -202,6 +202,10 @@ uint32 GUI::processEventQueue() {
 			mouseWheel(*e);
 		else if (e->type == Events::kEventTextInput)
 			textInput(*e);
+		else if (e->type == Events::kEventKeyDown)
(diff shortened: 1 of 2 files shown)


src/events/types.h
```diff
@@ -72,6 +72,13 @@ enum ITCEvent {
 	kITCEventMAX                     ///< For range checks.
 };
 
+/** Enum for keyboard keys. */
+enum Key {
+	kKeyBackspace   = SDLK_BACKSPACE
```
DrMcCoy: Right now, the rest of the code uses the SDLK_* values as is. That should be changed, yes, but that's a job for a different PR. As such, I don't think that enum belongs here yet.

Nostritius:
Ok, this version should now be up to date with the kKey enum

DrMcCoy:
Merged it as 51d1fb2ec, thanks! :)

PR #171, opened 2017-10-08
KOTOR: Add the nwscript function GetGender

Nostritius:
Add the nwscript function GetGender which returns the gender of a creature. Also I added the gender logic to the Creature class.


DrMcCoy:
Commited as cee09083c...7ed470255, thanks! :)

PR #172, opened 2017-10-08
KOTOR: Add the nwscript function GetLevelByClass

Nostritius:
Add the nwscript function GetLevelByClass which returns the class level of a creature. Also I added the class level logic to the Creature class. As far as I remember, a player can have multiple class levels because of the two times, he can choose a class (Soldier, Scoundrel and Scout, and later at Dantooine the Jedi classes. Then the old level is frozen and the new one is increased at every level up. Because of that, i designed it the way, a creature can have multiple levels for every class type.


DrMcCoy:
Commited as 7ed470255...aa47e8052, thanks! :)

PR #173, opened 2017-10-17
KOTOR2: Fix the border texture bug

Nostritius:
bildschirmfoto zu 2017-10-17 12-11-04
While testing KOTOR2, i realized, that after the border update it crashed with unknown textures. I looked  into this in detail and realized, that even without the borders, it is not exactly like the original. The appended picture shows the differences between the old layout left, the new layout right and the original game in the mid. I changed the used layout to the new layout.


DrMcCoy:
Commited as 3a24aad, thanks! :)

PR #174, opened 2017-10-17
KOTOR: Widget border fix

Nostritius:
I recognized, that in some cases, the borders became oversized. I investigated this, and found out, that there is a dimension parameter which controls the thickness of the border. I added handling for this parameter in the BorderQuad and KotORWidget classes.


DrMcCoy:
Commited as 3a24aad...cee0908, thanks! :)

PR #177, opened 2017-10-27
KOTOR: Add saving/loading and defaults to all checkboxes in options

fdde:


src/engines/kotor/gui/options/gameplay.cpp
```diff
@@ -63,6 +63,21 @@ void OptionsGameplayMenu::show() {
 
 	_difficulty = CLIP(ConfigMan.getInt("Difficulty Level", 0), 0, 2);
 	updateDifficulty(_difficulty);
+
+	_autoLevelUp = ConfigMan.getBool("Auto Level Up", false);
```
DrMcCoy: Hmm, please don't spaces in the config options. They basically map 1:1 to command line switches, and there a space obviously won't work. Also, while they are case-insensitive, I prefer if xoreos itself writes all-lower-case.

See how NWN does it:
`ConfigMan.setBool("mouseoverfeedback", mouseoverFeedback, true);`

src/engines/kotor/gui/options/gameplay.cpp
```diff
@@ -127,8 +167,20 @@ void OptionsGameplayMenu::updateDifficulty(int difficulty) {
 	}
 }
 
+void OptionsGameplayMenu::updateCheckbox(bool& option, const char* widget, bool state) {
```
DrMcCoy: That should be `bool &option` and `const char *widget`, i.e. the `*` and `&` should hug the name, not the type.

src/engines/kotor/gui/gui.cpp
```diff
@@ -279,6 +279,16 @@ void GUI::addBackground(const Common::UString &background) {
 		_background->setType(background);
 }
 
+void GUI::setCheckBoxState(const char* tag, bool state) {
```
DrMcCoy: Please make sure you put the `*` at the right place, with the name instead of the tag, all over this commit too

src/engines/kotor/gui/options/gameplay.cpp
```diff
@@ -102,18 +102,35 @@ void OptionsGameplayMenu::callbackActive(Widget &widget) {
...
+
+		_mouseMove = false;
+		setCheckBoxState("CB_INVERTCAM", _mouseMove);
+
+		_autoSave = false;
+		setCheckBoxState("CB_AUTOSAVE", _autoSave);
+
+		_reverseMinigameY = false;
+		setCheckBoxState("CB_REVERSE", _reverseMinigameY);
+
```
DrMcCoy: Trailing whitespace in that line

src/engines/kotor/gui/options/graphicsadv.h
```diff
@@ -36,8 +36,19 @@ class OptionsGraphicsAdvancedMenu : public GUI {
...
+	virtual void show();
+
+	virtual void adoptChanges();
+
 protected:
 	void callbackActive(Widget &widget);
+
+private:
+	int _textureQuality;
+
```
DrMcCoy: Trailing whitespace

src/engines/kotor/gui/options/mousesettings.cpp
```diff
@@ -38,18 +41,35 @@ OptionsMouseSettingsMenu::~OptionsMouseSettingsMenu() {
...
+	_reverseMouseButtons = ConfigMan.getBool("Reverse Mouse Buttons", false);
+	setCheckBoxState("CB_REVBUTTONS", _reverseMouseButtons);
+}
+
 void OptionsMouseSettingsMenu::callbackActive(Widget &widget) {
 
 	if (widget.getTag() == "BTN_DEFAULT") {
-
+		_reverseMouseButtons = false;
+	setCheckBoxState("CB_REVBUTTONS", _reverseMouseButtons);
```
DrMcCoy: The indenting is confusing here

src/engines/kotor/gui/options/gameplay.cpp
D8174: I was trying to match the "Difficulty Level" that was already there but I'll change them.

PR #179, opened 2017-10-29
KOTOR: Add colors for left/right buttons and set their highlighting to a static color

fdde:
I don't know if there is a more elegant way to do this.


DrMcCoy:
Merged it as 75c923a19, thanks! :)

Issue #169, opened 2017-09-02
KOTOR: Some textures not transparent


Nostritius:
I'm currently developing on the ingame GUI, and in there I recognized some weird behaviour of some textures, which are not transparent, as in the original game, but black. I already looked into the images and found out, that they actually have no alpha channel, but some Texture Information which could point to the way, they should be rendererd.
![bildschirmfoto zu 2017-09-02 16-19-41](https://user-images.githubusercontent.com/17826813/29996227-e7dc10b2-8ffa-11e7-8739-23a3d000d8e6.png)

adolfintel:
The grass and the tree leaves use that strange format too. I remember it being an issue on AMD drivers a few years ago

Nostritius:
You mean, the textures are rendered different to circumvent an amd driver bug, or is an amd driver bug causing this problem? Because this scene is rendered on a nvidia card.

adolfintel:
If I remember correctly (it was 2013), the game has some textures where it uses a keyed alpha, and some where the alpha of the background was like 0.01. This caused issues with the AMD driver back then where you could not see objects behind a transparent polygon.
Also if I remember correctly, some parts of the UI were rendered additively (most likely the ones where you have the black squares)

Nostritius:
Thanks for the information, the additive rendering did the trick. I implemented this in the new #181 PR.

PR #181, opened 2017-10-31
KOTOR: Basic KotOR ingame menu

Nostritius:
After solving #169 i was able to build a basic version of the ingame ui which displays all the HUD elements without doing anything else, while maintaining pickability of scene objects.


DrMcCoy:
Very nice, I like it. :)

However, I'm not sure just erroring out, completely shutting down the game, when the resolution doesn't match one of the supported ones, is all that great. I'd rather have the game continue without an ingame GUI for now, and a more explanatory message (including the supported resolutions) printed with warning().

Nostritius:
This code should now leave the menu with a warning and prints a warning while making the hud invisible.

PR #184, opened 2017-11-21
Base work for text rendering changes

farmboy0:
To make text layouting possible we need to give the text quad the same height and width as the underlaying widget quad. Layouting is done by defining a writing direction (ltr,rtl) and a block direction (these come later and are defined by the display language) and then aligning the text in the text quad according to vertical and horizontal alignment.

This PR
* renames the existing align to horizontal align
* add a vertical align
* defines common values for vertical and horizontal align
* adds a Text constructor with width and height parameters added

### DIFF
--- a/src/graphics/aurora/highlightabletext.h
+++ b/src/graphics/aurora/highlightabletext.h
@@ -32,7 +32,7 @@ class HighlightableText: public Text, public Highlightable {
 
   public:
 	HighlightableText(const FontHandle &font, const Common::UString &str,
-	     float r = 1.0f, float g = 1.0f, float b = 1.0f, float a = 1.0f, float halign = 0.0f);
+	     float r = 1.0f, float g = 1.0f, float b = 1.0f, float a = 1.0f, float halign = kHAlignLeft);
 	~HighlightableText();
 
 	void render(RenderPass pass);
--- a/src/graphics/aurora/highlightabletext.h
+++ b/src/graphics/aurora/highlightabletext.h
@@ -32,7 +32,8 @@ class HighlightableText: public Text, public Highlightable {
 
   public:
 	HighlightableText(const FontHandle &font, const Common::UString &str,
-	     float r = 1.0f, float g = 1.0f, float b = 1.0f, float a = 1.0f, float halign = kHAlignLeft);
+	     float r = 1.0f, float g = 1.0f, float b = 1.0f, float a = 1.0f,
+	     float halign = kHAlignLeft, float valign = kVAlignTop);
 	~HighlightableText();
 
 	void render(RenderPass pass);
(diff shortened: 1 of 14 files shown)


src/graphics/aurora/text.cpp
```diff
@@ -44,6 +44,17 @@ Text::Text(const FontHandle &font, const Common::UString &str,
 	_distance = -FLT_MAX;
 }
 
+Text::Text(const FontHandle &font, float w, float h, const Common::UString &str,
+		float r, float g, float b, float a, float halign, float valign) :
+	Graphics::GUIElement(Graphics::GUIElement::kGUIElementFront), _width(w), _height(h),
+	_r(r), _g(g), _b(b), _a(a), _font(font), _x(0.0f), _y(0.0f), _halign(halign),_valign(valign),
+	_disableColorTokens(false) {
```
DrMcCoy: I usually compile with -Werror, to catch warnings like these:

```
./src/graphics/aurora/text.h: In constructor ‘Graphics::Aurora::Text::Text(const Graphics::Aurora::FontHandle&, float, float, const Common::UString&, float, float, float, float, float, float)’:
./src/graphics/aurora/text.h:91:8: error: ‘Graphics::Aurora::Text::_height’ will be initialized after [-Werror=reorder]
  float _height;
        ^~~~~~~
./src/graphics/aurora/text.h:82:8: error:   ‘float Graphics::Aurora::Text::_r’ [-Werror=reorder]
  float _r, _g, _b, _a;
        ^~
src/graphics/aurora/text.cpp:47:1: error:   when initialized here [-Werror=reorder]
 Text::Text(const FontHandle &font, float w, float h, const Common::UString &str,
 ^~~~
```

src/graphics/aurora/highlightabletext.h
```diff
@@ -32,7 +32,8 @@ class HighlightableText: public Text, public Highlightable {
 
   public:
 	HighlightableText(const FontHandle &font, const Common::UString &str,
-	     float r = 1.0f, float g = 1.0f, float b = 1.0f, float a = 1.0f, float halign = kHAlignLeft);
+	     float r = 1.0f, float g = 1.0f, float b = 1.0f, float a = 1.0f,
+	     float halign = kHAlignLeft, float valign = kVAlignTop);
```
DrMcCoy: This isn't really visible on GitHub here, but this commit, which uses `kVAlignTop` comes before the 98d5bfd3c74b4205fc723716152fd67003ae3955, which defines `kVAlignTop` in the first place.

Which means that when going back to this commit, for example during a git bisect, this commit here won't compile. I'd like to avoid commits not compiling, if possible, even if later commits fix the error.

You should probably swap these two commits.

PR #185, opened 2017-11-22
GRAPHICS: Fix OSX context creation problem

Nostritius:
After I read, that the current version of xoreos is not working under osx, I began searching for the cause. I found out, that if you request an opengl 2.1 core context, osx actually returns an opengl 3.x core context. I fixed this by checking if the created context is really a 2.1 context and if not return to the third fallback.

### DIFF
--- a/src/graphics/windowman.cpp
+++ b/src/graphics/windowman.cpp
@@ -117,6 +117,19 @@ bool WindowManager::initRender(RenderType type, bool useDebug, int fsaa) {
 		return false;
 	}
 
+	/*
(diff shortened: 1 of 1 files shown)


DrMcCoy:
As @clone2727 said before you force-pushed (:P), there shouldn't be a warning if that's the common case.

Nostritius:
The warning shouldd now be removed.

DrMcCoy:
Merged as 25e996c8a51d4ea54ddd05d97da4f4641cdbe094, thanks! :)

PR #187, opened 2017-11-22
KOTOR: Add a basic ingame menu

Nostritius:
This PR adds a basic version of the kotor ingame menu.


farmboy0:
The GUIBack/Front solution had been developed because some dialogs put world elements between GUI elements. With the subscenequad there exists a better and more flexible solution now. I think it would be better to remove the GUIBack/Front code completely and treat all GUI elements as front.

DrMcCoy:
@farmboy0: The GUIBackObject queue is still used by Sonic, for the 2D area background that's shown behind the 3D objects. Not sure that can be done with SubSceneQuad instead.

farmboy0:
subscenequad is for 3D models, so no it cant. But it might better to implement this background as a sonic specific class like the guibackground for kotor1.

DrMcCoy:
Thanks, merged as fe9406066...a3a6c943e! :)

Changing the usage of the GUI stuff in KotOR to be always GUIFront would be a good next step, then, I guess.

PR #188, opened 2017-11-22
KOTOR2: Make the module listbox invisible

Nostritius:
The KotOR 2 main menu contained a widget called "LB_MODULES" which is definitely not in the orignal game and became visible after the list box update. I made it invisible.


DrMcCoy:
Merged it as 5c41b2915, thanks! :)

PR #195, opened 2017-12-02
AURORA: Add class for reading nfo files

Nostritius:
This PR adds a reader plus unit tests for reading nfo files. These files along with several other files are a save game in kotor. The nfo files save the metadata, that means the data, which can be seen from the save/load menu.


tests/aurora/nfofile.cpp
```diff
@@ -0,0 +1,112 @@
...
+		0x00, 0x04, 0x00, 0x00, 0x00, 0x05, 0x00, 0x00, 0x00, 0x06, 0x00, 0x00,
+		0x00, 0x07, 0x00, 0x00, 0x00, 0x08, 0x00, 0x00, 0x00, 0x09, 0x00, 0x00,
+		0x00, 0x0a, 0x00, 0x00, 0x00, 0x0b, 0x00, 0x00, 0x00, 0x0c, 0x00, 0x00,
+		0x00, 0x0d, 0x00, 0x00, 0x00, 0x0e, 0x00, 0x00, 0x00
+};
+
+GTEST_TEST(NFOFile, getPortrait) {
+	const Aurora::NFOFile nfo(new Common::MemoryReadStream(kNFOFile));
+
+	ASSERT_EQ(nfo.getPortrait0(), "po_pmha4");
```
DrMcCoy: For strings, you should use STREQ instead of EQ, in combination with c_str(). This gives better/clearer warning messages should the test fail (i.e. when it fails, it can print the actual string).

Also, use EXPECT_* instead of ASSERT_* where possible. If an ASSERT_* fails, that whole function will abort, while EXPECT_* will continue with the following lines. So with EXPECT_*, you're going to see if further tests also fail.

ASSERT_* are needed if the lines below depend on the test to succeed, for example when it would return a pointer that is then dereferenced. See also https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#assertions .

This line should then be
`EXPECT_STREQ(nfo.getPortrait0().c_str(), "po_pmha4");`

Likewise for the other lines below.

Nostritius:

DrMcCoy:
Thanks! :)

PR #198, opened 2017-12-03
GRAPHICS: Cache animation nodes

seedhartha:
Works for both KotOR and NWN.
FPS improvement on my machine (Ryzen 1600 + RX480):
KotOR (class selection) +62%
NWN (prelude, chapter 1-4) +6-16%


DrMcCoy:
Unfortunately, it segfaults in NWN, when the door in the prelude opens (talk to Bim, the one by the door, say "I don't need a tutorial", then the door opens, the animation plays and crash).

gdb and valgrind logs are here: https://gist.github.com/DrMcCoy/af0a801b11e323cd68a8b1e57fee8d5c

seedhartha:
Should be fixed now. Sorry for the inconvenience.

DrMcCoy:
Thanks! :D

PR #199, opened 2017-12-03
KOTOR: Confirm dialog

Nostritius:
This PR adds a confirm dialog for kotor which is used in several occasions.

### DIFF
--- a/src/graphics/aurora/borderquad.cpp
+++ b/src/graphics/aurora/borderquad.cpp
@@ -64,12 +64,13 @@ BorderQuad::BorderQuad(const Common::UString &edge, const Common::UString &corne
 	assert(!_edge.empty());
 }
 
-void BorderQuad::setPosition(float x, float y, float UNUSED(z)) {
+void BorderQuad::setPosition(float x, float y, float z) {
 	lockFrameIfVisible();
 
 	// Because of some graphic glitches with non-integer floats, i have to floor the values
 	_x = std::floor(x);
 	_y = std::floor(y);
+	_distance = z;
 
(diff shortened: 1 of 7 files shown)


src/graphics/aurora/borderquad.cpp
```diff
@@ -64,20 +64,21 @@ BorderQuad::BorderQuad(const Common::UString &edge, const Common::UString &corne
...
 }
 
-void BorderQuad::setPosition(float x, float y, float UNUSED(z)) {
+void BorderQuad::setPosition(float x, float y, float z) {
 	lockFrameIfVisible();
 
 	// Because of some graphic glitches with non-integer floats, i have to floor the values
 	_x = std::floor(x);
-	_y = std::floor(y);
+	_y = std::ceil(y);
```
DrMcCoy: Could you explain *why* you're ceiling the value instead of flooring it? Update the comment above with your explanation, please.

DrMcCoy:
Nice PR, thanks! :)

I'd just like you to explain your thought-processes a bit more in the commit messages, if possible. When I'm just sitting there, look at a commit and think "Why?", that's a point when some more explanation would be nice.

I know I'm also often not the best there, and people probably have that with some of my commits. I also need to work on that :)

Nostritius:
About the ceiling, i decided to remove this. It was about preventing a graphical glitch. Is there anything else i should explain? because the remainder is  (from my point of view) self explanatory.

DrMcCoy:
Merged it as 8f3b1b2a2...5627151ea, thanks! :)

PR #202, opened 2017-12-16
Gui scaling

farmboy0:
This is primarily for Jade.
All GUI elements are scaled to screen size.

### DIFF
--- a/src/graphics/graphics.h
+++ b/src/graphics/graphics.h
@@ -52,6 +52,11 @@ class Renderable;
 /** The graphics manager. */
 class GraphicsManager : public Common::Singleton<GraphicsManager>, public Events::Notifyable {
 public:
+	enum ScalingType {
+		kScalingNone = 0,
+		kScalingWindowSize
+	};
+
 	GraphicsManager();
 	~GraphicsManager();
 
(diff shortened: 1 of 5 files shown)


DrMcCoy:
Hmm, this scales while ignoring the aspect ratio of the original GUI. This will probably look ridiculous on those ultra-widescreen monitors (21:9, I think?).

Wouldn't it be better to scale while keeping the aspect ratio intact, centering in the dimension that doesn't fit (similar to how we're doing fullscreen videos right now)?

src/graphics/graphics.h
```diff
@@ -85,6 +90,10 @@ class GraphicsManager : public Common::Singleton<GraphicsManager>, public Events
 	/** Enable/Disable face culling. */
 	void setCullFace(bool enabled, GLenum mode = GL_BACK);
 
+	/** Configure scaling for the GUI */
```
DrMcCoy: Doxygen comments can't span multiple methods, unfortunately. You need to add one for each. (And end those comments with a fullstop, please.)

DrMcCoy:
Oh, and while GitHub doesn't show it, in the actual PR the Sonic commit is before the "Implement GUI scaling" commit. That probably should be the other way round, no? :P

Does Sonic actually work then, scaled? I mean, we're still forcing the resolution to 1x, too...

farmboy0:
The commit order was done on purpose so that when gui scaling is implemented sonic will use already the correct configuration.

farmboy0:
I havent implemented for sonic to work with any resolution.
While the GUI is already correctly scaled the world is not.
Thats why I left the window size restriction in place.

farmboy0:
> Hmm, this scales while ignoring the aspect ratio of the original GUI. This will probably look ridiculous on those ultra-widescreen monitors (21:9, I think?).

>Wouldn't it be better to scale while keeping the aspect ratio intact, centering in the dimension that doesn't fit (similar to how we're doing fullscreen videos right now)?

It shouldn't be too hard to add new scaling options.
The problem I see with this is that you will also need to render a black blackground or the world will show where the GUI isn't rendered.

DrMcCoy:
Thanks! :)

PR #206, opened 2017-12-28
KOTOR: Hit Points

Nostritius:
This PR adds everything necessary for handling hitpoints of objects and creatures. This includes the loading, handling and the appropriate nwscript functions.


src/engines/kotor/object.h
```diff
@@ -115,6 +128,10 @@ class Object : public Aurora::NWScript::Object, public KotOR::ScriptContainer {
 	bool _static; ///< Is the object static?
 	bool _usable; ///< Is the object usable?
 
+	int _currentHitPoints; ///< The current hitpoints of the object.
+	int _maxHitPoints; ///< The maximum hitpoints of the object.
+	bool _minOneHitPoint; ///< If the object should have at least one hitpoint.
```
DrMcCoy: Initialize those in the constructor, please. To 0, 0, false, I guess. Especially important for _minOneHitPoint, which isn't set in Placeable::loadObject().

src/engines/kotor/script/functions_object.cpp
```diff
@@ -72,6 +72,48 @@ void Functions::getObjectByTag(Aurora::NWScript::FunctionContext &ctx) {
 	ctx.getReturn() = search->get();
 }
 
+void Functions::getMinOneHP(Aurora::NWScript::FunctionContext &ctx) {
+	Aurora::NWScript::Object *object = ctx.getParams()[0].getObject();
+
+	Object *kotorObject = ObjectContainer::toObject(object);
+
+	ctx.getReturn() = kotorObject->getMinOneHitPoints();
```
DrMcCoy: You need to check for kotorObject being 0 before dereferencing.

`ctx.getReturn() = kotorObject ? kotorObject->getMinOneHitPoints() : 0`

should do it.

Nostritius:

DrMcCoy:
You forgot that != 0 check in `Functions::setMaxHitPoints()`.

Btw, we should really check what the original does for the set* functions, when the object is not valid. That throw in xoreos would mean the script gets aborted at that point. Might be that the original just ignores these calls on invalid objects instead.

Trying that out can be a bit complicated, though. You need to look for a copy of the OpenKnights nwnnsscomp (the NWScript compiler) that's been modified to work with KotOR (*), write a script that calls the functions on invalid objects and does something visible afterwards and put it into a modified module somewhere.

For now, I think it would be good to put a "TODO" comment somewhere at the top of functions_object.cpp.

(*) Might be somewhere on [Deadlystream](http://deadlystream.com/).

Nostritius:
The last issues should now be fixed.

src/engines/kotor/script/functions_object.cpp
```diff
@@ -72,6 +74,58 @@ void Functions::getObjectByTag(Aurora::NWScript::FunctionContext &ctx) {
...
+}
+
+void Functions::setMaxHitPoints(Aurora::NWScript::FunctionContext &ctx) {
+	Aurora::NWScript::Object *object = ctx.getParams()[0].getObject();
+	int maxHitPoints = ctx.getParams()[1].getInt();
+
+	if (maxHitPoints == 0)
+		maxHitPoints = 1;
+
+	Object *kotorObject = ObjectContainer::toObject(object);
```
DrMcCoy: You still don't check for kotorObject != 0 before dereferencing it the next two lines.

Nostritius:
This should now fix the invalid object issue

PR #208, opened 2017-12-29
KOTOR: Improved levels

Nostritius:
When I first wrote code for handling the kotor level system, i forgot an important aspect. The order in which the levels are added is actually relevant. This PR tries to deal with this, and adds some extended nwscript functions regarding this.


DrMcCoy:
How is the order relevant? IIRC, you choose one base class at game start, level always that class, then later pick a Jedi class and always level that. The order should always be the same, no?

src/engines/kotor/creature.cpp
```diff
@@ -317,7 +307,10 @@ void Creature::createPC(CharacterGenerationInfo *info) {
 	}
 
 	// set the specific class to level 1
-	_levels[info->getClass()] = 1;
+	_levels.clear();
+	_levels.emplace_back();
```
DrMcCoy: std::vector::emplace_back() is a C++11 feature, don't use that yet.

Just resize(1) instead of clear+emplace. It's shorter, too :P

src/engines/kotor/creature.cpp
```diff
@@ -96,11 +85,12 @@ Gender Creature::getGender() const {
 }
 
 int Creature::getLevel(const Class &c) const {
-	const std::map<Class, int>::const_iterator it = _levels.find(c);
-	if (it == _levels.end())
-		return 0;
+	for (int i = 0; i < _levels.size(); ++i) {
```
DrMcCoy: std::vector::size() returns a size_t, not an int.

Nostritius:
According to nwscript.nss it could have up to three different class levels, and gives no clue, that the first level is every time a normal class and the second is every time a jedi class. This functions explicitly require which class you selected first, second and so on. I wanted to make it more flexible to match the nwscript specification.

DrMcCoy:
Okay :)

Nostritius:

src/engines/kotor/creature.cpp
```diff
@@ -327,8 +341,7 @@ void Creature::createPC(CharacterGenerationInfo *info) {
 	}
 
 	// set the specific class to level 1
-	_levels.clear();
-	_levels.emplace_back();
+	_levels.resize(1);
```
DrMcCoy: I think you merged the emplace_back() fix into the wrong commit here.

https://github.com/xoreos/xoreos/pull/208/commits/6fd4e0ad9d621ea2ecc2f8c3d34dfb425f9c17b8#diff-d2c774ad01449ab50752df0719218776R331 adds it, and you remove it here. You probably meant to merge that fix into 6fd4e0ad9d621ea2ecc2f8c3d34dfb425f9c17b8 instead.

Nostritius:
The emplace_back issue should now be fixed

DrMcCoy:
Going in-game doesn't work anymore now, though. The entry module doesn't load.

```
WARNING: Failed loading module "end_m01aa"
    Because: GFF3: No such field
```

Probably related to the loading of creature class levels in 8dee9678682793036d79b967b52ca6aebb67cf12.

Nostritius:
The the issue should now be fixed.

PR #209, opened 2017-12-29
AURORA: Prevent the overwrite of a none-zero by a zero return code

Nostritius:
It is possible that a child gui overwrites a none-zero return code of the parent gui. This should prevent it.


DrMcCoy:
Merged it as 1361ea7e2, thanks! :)

PR #210, opened 2017-12-29
KOTOR: Race and SubRace handling

Nostritius:
This PR adds race handling for kotor including the nwscript functions.


DrMcCoy:
Merged it as aabf71141...5d1e72f42, thanks! :)

PR #212, opened 2018-01-04
KOTOR: Catch exception on creature subrace load

seedhartha:
Commit https://github.com/xoreos/xoreos/commit/372953e4196fca6cb7fcdad685a96c6386b7401a introduced a bug: due to unhandled exception on creature subrace load KotOR became unplayable past the character generation.

### DIFF
--- a/src/engines/kotor/creature.cpp
+++ b/src/engines/kotor/creature.cpp
@@ -201,7 +201,12 @@ void Creature::loadProperties(const Aurora::GFF3Struct &gff) {
 
 	// Race
 	_race = Race(gff.getSint("Race", _race));
-	_subRace = SubRace(gff.getSint("Subrace", _subRace));
+
+	try {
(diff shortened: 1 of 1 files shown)


farmboy0:
Are you sure this is the best way to do it?
If a value is loaded that is not defined in the subrace enum it should be added.
If there is no subrace value to load, the code could check for its existence first or give a default value.

Nostritius:
I 'm actually investigating the error, and found out, that there is a "SubraceIndex" field. The "Subrace" seems not to be used. I created a PR which should fix this problem in #213.

DrMcCoy:
GFF3Struct::getSint() should return the second parameter as a default, if the field doesn't exist. That's why there's _subRace again there, so that the value is unchanged if the field doesn't exist. No exception is thrown in that case.

However, since the field is a string and not an int, that's why there's an exception there (which should be "GFF3: Field is not an int type").

So yeah, #213 seems to be the proper fix for this bug, sorry. :P

PR #213, opened 2018-01-04
KOTOR: Fix subrace loading bug

Nostritius:
After #212 I searched for the cause of the error and found out, that the Subrace field is actually a string field and is not used, even on wookies. I found out, that there is a second field called "SubraceIndex" which contains the actual subrace information.


DrMcCoy:
Ah, that seems to be a holdover from NWN. There, Subrace is a string that the user can fill out however they want.

PR #214, opened 2018-01-14
KOTOR: 3D Sounds

Nostritius:
This PR adds the 3d sound objects in an area and some of its corresponding nwscript functions.


src/engines/aurora/camera.cpp
```diff
@@ -22,6 +22,7 @@
  *  Engine utility functions for camera handling.
  */
 
+#include <src/sound/sound.h>
```
DrMcCoy: Please use `#include ""` instead of `#include <>` for xoreos' own files.

src/engines/aurora/camera.cpp
```diff
@@ -77,6 +78,8 @@ bool handleCameraKeyboardInput(const Events::Event &e) {
 	} else
 		return false;
 
+	//TODO: Move this somewhere else
```
DrMcCoy: Then move this somewhere else? :P

(Also, there needs to be space after `//`)

src/engines/kotor/sound.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace KotOR {
+
+SoundObject::SoundObject(const Aurora::GFF3Struct &sound) {
+	Common::UString temp = sound.getString("TemplateResRef");
+
+	Common::ScopedPtr<Aurora::GFF3File> uts;
+	if (!temp.empty())
+		uts.reset(loadOptionalGFF3(temp, Aurora::kFileTypeUTS, MKTAG('U', 'T', 'S', ' ')));
+
+	const Aurora::GFF3Struct &gff = uts->getTopLevel();
```
DrMcCoy: If you're using `loadOptionalGFF3()`, the return value might be 0! You can't just dereference `uts` here without checking for 0.

Also, `uts` might be 0 if `temp` is empty, of course.

`Common::ScopedPtr::operator->()` does *not* check for 0 itself.

In fact, can anything in this function work if temp is empty at all?

src/engines/kotor/sound.h
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  An ingame sound in kotor.
+ */
+
+#ifndef XOREOS_SOUND_H
```
DrMcCoy: Wrong format for the include guard name

src/engines/kotor/sound.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+	SoundObject(const Aurora::GFF3Struct &sound);
+
+	void setPosition(float x, float y, float z);
+
+	void play();
+	void stop();
+
+private:
+	uint32 intervalCallback(uint32);
```
DrMcCoy: Unused.

src/engines/aurora/camera.cpp
Nostritius: Because the player is currently not moving, i attached the listener to the camera. This should be changed of course, when the player can move.

DrMcCoy: Then please state something of this effect in the comment, so that everybody knows what this is about :)

Nostritius:
The issues should now be fixed.

DrMcCoy:
There really needs to be a `SoundObject` destructor with `stop()` call in it, to be sure the sound is always stopped when the object is deleted.

Nostritius:
The listener position will now be modified by the player character and a TODO for the random sounds is added.

src/engines/kotor/module.cpp
```diff
@@ -22,6 +22,7 @@
  *  The context needed to run a Star Wars: Knights of the Old Republic module.
  */
 
+#include <src/sound/sound.h>
```
DrMcCoy: Please, you `#include ""`, not `#include <>` for our own includes

DrMcCoy:
Hmm, I rather think the 3D listener placement should follow the camera, not the PC? At least in KotOR. Walking back and forth (with W and S) beside a sound, for example the broken pipe thing at the start, makes it switch between the left/right speaker.

Does the original does that as well?

(I *know* that in NWN, the listener is the camera, at least partially. Zooming into a torch, for example, makes the torch be louder.)

Also, opening the chest at the start is now very quiet.

Nostritius:
I set the listener movement tback to the pc position. Also i added openal relative mode turned on per default for all sounds, which fixes the chest to quiet problem. I also added a distance model, which is the closest i could find to the original game.

DrMcCoy:
Looks (sounds) good now :)

Merged it as 121ec93ab...6e31f35ef, thanks! :)

PR #216, opened 2018-01-15
KOTOR: Fix that border dimension misplaced widgets bug

Nostritius:
The one widget which was misplaced, had weird dimension values, while not having an actual border. This commit prevents a widget from rearranging itself with border dimensions if it actually has no border.


DrMcCoy:
Merged as 2a9506ba, thanks! :)

PR #219, opened 2018-01-19
KOTOR: Return button string

Nostritius:
This PR makes it possible to set the string of the "Return to Ebon hawk"  button and it's confirm dialog, and adds the possiblity to disable it and adds the specific nwscript functions.

### DIFF
--- a/src/engines/kotor/gui/ingame/hud.cpp
+++ b/src/engines/kotor/gui/ingame/hud.cpp
@@ -105,6 +105,18 @@ HUD::HUD(Engines::Console *console) : GUI(console) {
 	getWidget("PB_HEALTH")->setInvisible(true);
 }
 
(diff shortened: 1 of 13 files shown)


DrMcCoy:
Is there a module where we can see the string set to "Return to ebon hawk" work already? I tried loading a few random modules via the debug console, but it's always "Transit disabled"

Nostritius:
Good Question, I assume, that not every module is calling this function. I tested it only against the first two endar spire modules and only the first called this function.

DrMcCoy:
Merged as af924c29...aa12dd1d, thanks! :)

It's just a bit unfortunate that we can't really test it working correctly. I prefer it if we have some visible test cases for engine function, before we implement them.

PR #221, opened 2018-02-10
KOTOR: Load screen

Nostritius:
This PR introduces a loading screen for kotor, which at the moment will be between the mainmenu and the chargen menu and between entering different modules.

### DIFF
--- a/src/engines/kotor/module.cpp
+++ b/src/engines/kotor/module.cpp
@@ -58,6 +58,8 @@
 
 #include "src/engines/kotor/gui/ingame/ingame.h"
 
+#include "src/engines/kotor/gui/loadscreen/loadscreen.h"
+
 namespace Engines {
 
 namespace KotOR {
@@ -118,6 +120,10 @@ void Module::load(const Common::UString &module, const Common::UString &entryLoc
 
 void Module::loadModule(const Common::UString &module, const Common::UString &entryLocation,
                         ObjectType entryLocationType) {
+	_ingame->hide();
+
+	LoadScreen loadScreen(module, _console);
+	loadScreen.show();
 
 	unload(false);
 
@@ -159,6 +165,10 @@ void Module::loadModule(const Common::UString &module, const Common::UString &en
 	_newModule.clear();
 
 	_hasModule = true;
+
+	loadScreen.hide();
+
+	_ingame->show();
 }
 
 void Module::usePC(Creature *pc) {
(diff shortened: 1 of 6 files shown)


src/engines/kotor/module.cpp
```diff
@@ -22,6 +22,8 @@
  *  The context needed to run a Star Wars: Knights of the Old Republic module.
  */
 
+#include <src/engines/kotor/gui/loadscreen/loadscreen.h>
```
DrMcCoy: Please use `#include ""` instead of `#include <>` for our own includes. This one should also be way down in the list, and of course below the Boost.Bind one.

DrMcCoy:
When the user quits xoreos (with Ctrl+Q) during a loading screen, xoreos segfaults.

valgrind log is here: https://gist.github.com/DrMcCoy/23b4d032baf19f4d128c478012d44dcd

DrMcCoy:
That was the main menu one. Here's a valgrind log for the module one: https://gist.github.com/DrMcCoy/d9e44383aec04e271866ba0e41bd827d

Nostritius:
The issues should now be solved

DrMcCoy:
The segfaults are still there.

Nostritius:
Now the two segfaults should be solved.

DrMcCoy:
Maybe. However, no xoreos deadlocks when quitting during a loading screen.

For the main menu one, the issue seems to be with the `while (!_classSelection);` in src/engines/kotor/gui/main/main.cpp:156. For the module one, it's the `while (!_area);` src/engines/kotor/module.cpp:299.

When quitting, these variables are never set, so the thread just locks.

Also, this is *not* how to synchronize between thread. You need to use proper thread synchronization primitives, like a mutex or a semaphore. Just busy-waiting on a normal variable is bad: due to caching and compiler optimization, this can break very easily.

Nostritius:
Hopefully this should fix the issue now. I have removed the while loop and used mutex objects.

Nostritius:
I have now completely removed the seperate thread for loading. I also rebased the code because of recent changes in the module class.

DrMcCoy:
It's not actually modifying the progress bar while loading yet, is it?

Nostritius:
Yes, i decided that i try to get the loading screen running in this commit while offering the possibility to modify the load process. Since it will become quite complex, to measure the current loading progress, especially, if we want a higher granularity, i wanted to leave this for another PR.

DrMcCoy:
Merged as d9ed93d0...934e0f04, thanks! :)

PR #223, opened 2018-03-10
KOTOR: Global numbers

Nostritius:
This PR implements the global numbers functionality for kotor


src/engines/kotor/module.h
```diff
@@ -194,6 +201,10 @@ class Module : public KotOR::Object, public KotOR::ObjectContainer {
 
 	Common::ScopedPtr<IngameGUI> _ingame; ///< The ingame ui.
 
+	// .--- Global values
+	std::map<Common::UString, int> _globalNumbers;
```
DrMcCoy: This map needs to be cleared when we leave the game (to go back to the main menu or load a saved game), i.e. in `Module::unload()` with completeUnload == true.

src/engines/kotor/module.cpp
```diff
@@ -519,6 +519,14 @@ void Module::setReturnEnabled(bool enabled) {
 	_ingame->setReturnEnabled(enabled);
 }
 
+void Module::setGlobalNumber(const Common::UString &id, int value) {
+	_globalNumbers[id] = value;
+}
+
+int Module::getGlobalNumber(const Common::UString &id) {
+	return _globalNumbers[id];
+}
```
DrMcCoy: Make that method const, by using `_globalNumbers.find()` instead of `operator[]`. This also means that you need to explictly state the default value (which would be 0, I suppose?) when the id key is not within the map.

src/engines/kotor/module.cpp
```diff
@@ -519,6 +521,17 @@ void Module::setReturnEnabled(bool enabled) {
 	_ingame->setReturnEnabled(enabled);
 }
 
+void Module::setGlobalNumber(const Common::UString &id, int value) {
+	_globalNumbers[id] = value;
+}
+
+int Module::getGlobalNumber(const Common::UString &id) const {
+	if (_globalNumbers.find(id) != _globalNumbers.end())
+		return _globalNumbers[id];
```
DrMcCoy: That doesn't work that way, you're still using `operator[]`.

`_globalNumbers.find(id)` returns an iterator you can use to get the value (if it's not the end iterator). See for example https://github.com/xoreos/xoreos/blob/master/src/engines/kotor/area.cpp#L402 .

Nostritius:
The issue should now be fixed.

DrMcCoy:
Merged as b47c618b...16a71581, thanks! :)

PR #224, opened 2018-03-10
KOTOR: Add barebone resolution submenu to graphics menu

Nostritius:
This PR implements the barebone resolution menu for the kotor options menu


DrMcCoy:
...Why does the border around the "Advanced Options" button vanish when I click on Cancel in the resolution menu? O_o

DrMcCoy:
Apart from this oddity (which is probably not the fault of this PR as such), this looks good to me. Merged as 55ecfd009cf2c18c4c4929158daa325a67d16fae, thanks! :)

Nostritius:
Could you file a bug for this? So I won't forget to look after this.

DrMcCoy:
Right, good idea. #231 it is, now.

PR #225, opened 2018-03-25
KOTOR: Container menu

Nostritius:
This commit tries to bridge between the selection of objects and the handling of module specific stuff, in this case, when the player clicks on a chest, the ingame ui will open a window for managing the contents of the container.

### DIFF
--- a/src/engines/kotor/gui/ingame/ingame.h
+++ b/src/engines/kotor/gui/ingame/ingame.h
@@ -52,6 +52,9 @@ public:
 	void setReturnQueryStrref(uint32 id);
 	void setReturnEnabled(bool enabled);
 
+	// Container inventory handling
+	void showContainer();
+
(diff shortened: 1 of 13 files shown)


src/engines/kotor/gui/ingame/ingame.h
```diff
@@ -44,6 +44,9 @@ class IngameGUI {
 	void setReturnStrref(uint32 id);
 	void setReturnQueryStrref(uint32 id);
 	void setReturnEnabled(bool enabled);
+
```
DrMcCoy: Trailing whitespace

Nostritius:
The issues should now be fixed. I also added another commit which fixes the flickering and that the gui takes a long time problem.

DrMcCoy:
Merged, thanks! :)

PR #227, opened 2018-05-02
NETWORK: Initial basic network support (WiP)

Nostritius:
I started to look at the nwn network protocol which is used for multiplayer sessions. It was surprisingly easy to find a basic structure and gain enough information to reproduce the protocol used by the nwn lobby which i want to use as basic proof of concept demo. I added a network part for xoreos using boost.asio. With a simple synchronous Api (I want to add asynchronous sending and receiving later). I paused my work on this temporarily, so you can view the Api and make suggestions or critic on it's design.

### DIFF
--- a/src/engines/nwn/gui/multiplayer/rules.mk
+++ b/src/engines/nwn/gui/multiplayer/rules.mk
@@ -20,7 +20,9 @@
 # Multiplayer specific GUI in Neverwinter Nights.
 
 src_engines_nwn_libnwn_la_SOURCES += \
+    src/engines/nwn/gui/multiplayer/multiplayer.h \
     $(EMPTY)
 
 src_engines_nwn_libnwn_la_SOURCES += \
+    src/engines/nwn/gui/multiplayer/multiplayer.cpp \
     $(EMPTY)
--- a/src/engines/nwn/gui/multiplayer/rules.mk
+++ b/src/engines/nwn/gui/multiplayer/rules.mk
@@ -17,23 +17,10 @@
 # You should have received a copy of the GNU General Public License
 # along with xoreos. If not, see <http://www.gnu.org/licenses/>.
 
-# GUI system in Neverwinter Nights.
+# Multiplayer specific GUI in Neverwinter Nights.
 
 src_engines_nwn_libnwn_la_SOURCES += \
-    src/engines/nwn/gui/gui.h \
-    src/engines/nwn/gui/legal.h \
     $(EMPTY)
 
 src_engines_nwn_libnwn_la_SOURCES += \
-    src/engines/nwn/gui/gui.cpp \
-    src/engines/nwn/gui/legal.cpp \
     $(EMPTY)
-
-include src/engines/nwn/gui/widgets/rules.mk
-include src/engines/nwn/gui/dialogs/rules.mk
-include src/engines/nwn/gui/main/rules.mk
-include src/engines/nwn/gui/chargen/rules.mk
-include src/engines/nwn/gui/options/rules.mk
-include src/engines/nwn/gui/ingame/rules.mk
-include src/engines/nwn/gui/multiplayer/rules.mk
-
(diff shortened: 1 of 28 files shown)


src/network/udpsocket.cpp
```diff
@@ -0,0 +1,92 @@
...
+UDPSocket::~UDPSocket() {
+	_socket.close();
+}
+
+void UDPSocket::bind(UDPEndpoint endpoint) {
+	_socket.bind(endpoint);
+}
+
+void UDPSocket::send(const Packet &packet) {
+	char message[packet.getSize()];
```
DrMcCoy: While variable-length arrays do exist in C99, they do not exist in any current C++ standard. Last I heard, they're not even proposed for C++20.

You need to use a Common::ScopedArray<char> for that.

src/network/socket.h
```diff
@@ -0,0 +1,46 @@
...
+#ifndef NETWORK_SOCKET_H
+#define NETWORK_SOCKET_H
+
+#include <boost/asio.hpp>
+
+#include "src/network/packet.h"
+
+namespace Network {
+
+template<typename T> class Socket : boost::noncopyable {
```
DrMcCoy: Socket needs a virtual destructor.

src/network/packet.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifndef NETWORK_PACKET_H
+#define NETWORK_PACKET_H
+
+#include "src/common/writestream.h"
+#include "src/common/readstream.h"
+
+namespace Network {
+
+class Packet {
```
DrMcCoy: Packet needs a virtual constructor

src/engines/nwn/gui/multiplayer/rules.mk
```diff
@@ -0,0 +1,26 @@
...
+# You should have received a copy of the GNU General Public License
+# along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+
+# Multiplayer specific GUI in Neverwinter Nights.
+
+src_engines_nwn_libnwn_la_SOURCES += \
+    $(EMPTY)
+
+src_engines_nwn_libnwn_la_SOURCES += \
+    $(EMPTY)
```
DrMcCoy: Missing end-of-line at the end of the file.

src/engines/nwn/gui/multiplayer/serverlist.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "src/engines/aurora/console.h"
+#include "src/engines/aurora/widget.h"
+
+#include "src/engines/nwn/gui/multiplayer/serverlist.h"
+
+namespace Engines {
+
+namespace NWN {
+
+ServerList::ServerList(::Engines::Console *console) {
```
DrMcCoy: console needs to be marked with the UNUSED() macro.

src/network/packet.h
```diff
@@ -0,0 +1,48 @@
...
+public:
+	Packet();
+
+	/** Calculate the size a buffer for this packet would consume. */
+	virtual unsigned int getSize() const;
```
DrMcCoy: Should be `size_t` instead of `unsigned int`.

src/network/aurora/bnxr_packet.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+void BNXRPacket::serialize(Common::WriteStream *stream) const {
+	LobbyPacket::serialize(stream);
+}
+
+void BNXRPacket::deserialize(Common::SeekableReadStream *stream) {
+	LobbyPacket::deserialize(stream);
+
+	stream->seek(2, Common::SeekableReadStream::kOriginCurrent);
```
DrMcCoy: `skip()`, not `seek()` is what you want there :P.

src/network/aurora/lobbypacket.cpp
```diff
@@ -0,0 +1,77 @@
...
+	else
+		_serverMessage = stream->readByte();
+
+	_protocolVersion = stream->readByte();
+}
+
+Common::UString LobbyPacket::readLengthPrefixedString(Common::SeekableReadStream *stream) {
+	byte length = stream->readByte();
+	char string[length];
+	stream->read(string, length);
```
DrMcCoy: You need to check that return value of read, see if it correctly read the number of bytes.

(Also, no variable-length arrays in C++)

PR #228, opened 2018-05-04
AURORA: Add ltr file reader

Nostritius:
The ltr files are used in the random name generation for generating random names. This PR adds support for reading ltr files and generating names from it.


src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
+//
+// Created by patrick on 02.05.18.
+//
```
DrMcCoy: That should be our standard copyright header instead

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include "src/common/readstream.h"
+#include "src/common/strutil.h"
+
+#include "src/aurora/ltrfile.h"
+#include "src/aurora/resman.h"
+
+static const uint32 kLTRID  = MKTAG('L', 'T', 'R', ' ');
+static const uint32 kVersion10 = MKTAG('V', '1', '.', '0');
```
DrMcCoy: Align those properly, please, or take out the second space between `kLTRID` and `=`

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+static const char* kLetters = "abcdefghijklmnopqrstuvwxyz'-";
```
DrMcCoy: `static const char *kLetters`

Also, I wonder if this works correctly with the Russian version of the game, and also the CJK ones. But that's something for the future. Add a TODO comment about checking this with non-latin games, above the letters, please.

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+namespace Aurora {
+
+LTRFile::LTRFile(Common::UString name) {
```
DrMcCoy: `name` can be a `const Common::UString &` here

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+	readHeader(*stream);
+
+	if (_id != kLTRID)
+		throw Common::Exception("Not a LTR file (%s)", Common::debugTag(_id).c_str());
+
+	if (_version != kVersion10)
+		throw Common::Exception("Unsupported LTR file version %s", Common::debugTag(_version).c_str());
+
+	byte letterCount = stream->readByte();
+	if (letterCount != 28)
```
DrMcCoy: There are files in the games data files that have only 26 letters allocated, `names.ltr`. Are those completely unused?

Otherwise, it might be useful adding support for other letter counts than 28. Maybe pop in a TODO comment here to.

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+		throw Common::Exception("Unsupported Letter count %i", letterCount);
+
+	stream->read(&_singleLetters, sizeof(LetterSet));
```
DrMcCoy: Don't do that, please. Due to struct alignment, endianness and potentially float format differences, this is not portable.

Instead, please read the individual floats in loops with readIEEEFloatLE().

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+	_tripleLetters.resize(28);
+	for (int i = 0; i < 28; ++i) {
+		_tripleLetters[i].resize(28);
+		for (int j = 0; j < 28; ++j) {
+			stream->read(&_tripleLetters[i][j], sizeof(LetterSet));
+		}
+	}
+}
+
+Common::UString LTRFile::generateRandomName(unsigned int maxLetters) {
```
DrMcCoy: Might be worth changing that `unsigned int` to a `size_t`.

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+	// TODO: There is a lot of data yet unused in the letter files.
+	Common::UString name;
+	float probability;
+	int firstLetterIndex = 0;
+	int secondLetterIndex = 0;
+
+	if (maxLetters < 4)
+		throw Common::Exception("Invalid maximal letters. A name has to be at least 4 letters, %i given", maxLetters);
+
+	//get the first three letters.
```
DrMcCoy: A space after the `//` and then upper-case first letter, please.

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,117 @@
...
+		if (_tripleLetters[firstLetterIndex][secondLetterIndex].start[i] > probability) {
+			name += kLetters[i];
+			firstLetterIndex = secondLetterIndex;
+			secondLetterIndex = i;
+			break;
+		}
+	}
+
+	// Generate as much letters as max letters minus start letters minus end letters are given.
+	int length = std::rand() % (maxLetters - 4);
```
DrMcCoy: If `maxLetters` is exactly 4 here, this will produce a division by 0.

src/aurora/ltrfile.h
```diff
@@ -0,0 +1,60 @@
...
+
+private:
+	struct LetterSet {
+		float start[28];
+		float mid[28];
+		float end[28];
+	};
+
+	LetterSet                           _singleLetters;
+	std::vector<LetterSet>              _doubleLetters;
```
DrMcCoy: You need to `#include <vector>` then. `ustring.h` also does, so this compiles without, but still, include what you use.

src/aurora/ltrfile.h
```diff
@@ -0,0 +1,60 @@
...
+	std::vector<std::vector<LetterSet>> _tripleLetters;
```
DrMcCoy: This needs to be `std::vector< std::vector<LetterSet> >`, with a space separating the two `>`. This has been loosened in C++11, I think, but C++03 still needs that space.

src/aurora/ltrfile.cpp
Nostritius: Which game exactly uses the names.ltr? I would assume that this simply means, that it is using the standard alphabet without ' and -.

DrMcCoy: Both NWN and NWN2 have a names.ltr with 26 letters in their data files. Dunno whether they're used, though

Nostritius: I looked into nwn, and it seems that the name generation is already using species dependent ltr files. But i have also added for the sake of completeness a second mode, where the standard alphabet without ' and - is used.

DrMcCoy: A second mode? Can't you just make the number of letters completely variable? As far as I understood, the letter set is fixed, and the less letters, the more letters from the end are left out, no?

Nostritius: Well, I cannot say for sure, that they do not use other special characters in more than 26 letter alphabets.

Nostritius:
I added a unit test and fixed the issues.

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,186 @@
...
+static const char *kLetters26 = "abcdefghijklmnopqrstuvwxyz";
+
+namespace Aurora {
+
+LTRFile::LTRFile(const Common::UString &ltr) {
+	Common::ScopedPtr<Common::SeekableReadStream> stream(ResMan.getResource(ltr, kFileTypeLTR));
+	load(stream.get());
+}
+
+LTRFile::LTRFile(Common::SeekableReadStream *stream) {
```
DrMcCoy: LTRFile doesn't use the stream after loading anymore, right? So it should take a `Common::SeekableReadStream &stream`, to indicate that the LTRFile doesn't take over the stream.

src/aurora/ltrfile.cpp
```diff
@@ -0,0 +1,186 @@
...
+	int secondLetterIndex = 0;
+
+	if (maxLetters < 4)
+		throw Common::Exception(
+				"Invalid maximal letters. A name has to be at least 4 letters, %i given",
+				static_cast<int>(maxLetters));
+
+	// Get the first three letters.
+	probability = static_cast<float>(std::rand())/RAND_MAX;
+	for (int i = 0; i < 28; ++i) {
```
DrMcCoy: These `28`s should be `_letterCount`.

Nostritius:

DrMcCoy:
The LTRFile constructor taking a SeekableReadStream should also take a reference, not a pointer, please. In fact, that's the important part. :P

Also, don't use ScopedPtr in the unit tests, but a manual delete. Unit tests are supposed to only test the one specific thing, so they should ideally not depend on anything else (i.e. if ScopedPtr would break, we don't want the LTRFile unit test to also break because of that). Now, for xoreos, this isn't so easy, so I didn't, for example, mock out UString everywhere, or the stream readers. But the ScopedPtr is easy to not use.

Nostritius:
The issues should now be solved.

DrMcCoy:
Merged as 5a411f86...09a54cd1, thanks! :)

PR #229, opened 2018-05-05
KOTOR: Add background for movie menu

Nostritius:
I recognized, that the movie menu was missing a background, so I added one in this PR.


DrMcCoy:
Merged as 5a411f86ddb44790e090fc6fc5cb5d2a9e17652d, thanks! :)

PR #230, opened 2018-05-05
KOTOR: Initial satellite camera implementation w\ PC movement

seedhartha:
Alters camera behavior in KotOR to match the original game. Press A/D to rotate the camera around PC, W/S to move PC forward or backwards. Use console command "tfc" to toggle free roam camera mode.

TODO: Use FOV, pitch and distance values from camerastyle.2da; smooth out camera rotation and PC movement

### DIFF
--- a/src/engines/aurora/satellitecamera.cpp
+++ b/src/engines/aurora/satellitecamera.cpp
@@ -1,0 +1,88 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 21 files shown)


DrMcCoy:
Could you describe a bit what this does and how, where, etc.? :)

seedhartha:
Updated description.

src/engines/aurora/satellitecamera.cpp
```diff
@@ -0,0 +1,82 @@
...
+	_target._z = z;
+}
+
+void SatelliteCamera::setDistance(float value) {
+	_distance = value;
+}
+
+void SatelliteCamera::setPitch(float value) {
+	_pitch = value;
+	auto pitchRad = Common::deg2rad(_pitch);
```
DrMcCoy: Don't use `auto`, please. We're still C++03 at the moment.

DrMcCoy:
When toggling the free roam camera on, and moving the mouse while holding the middle mouse button (i.e. "free look"), it now lags quite a bit.

The issue is that you pulled the `CameraMan.update()` into the event queue loop in `Module::handleEvents()`. The update resorts all objects, so it is pretty taxing, and mouse move events fire in quick successions. My hacky resolution was to let the input handler process a whole batch before "finalizing" it by resorting the objects.

seedhartha:
Fixed free roam camera lag. Also implemented idle and running animations for PC.

DrMcCoy:
Very nice! Merged as b47c618b6cce739a8402dee0b99c9627a7814d37 (I word-wrapped the commit message to max 70 characters per line), thanks! :D

PR #232, opened 2018-05-06
KOTOR: Implement frame based camera rotation and PC movement

seedhartha:
Camera rotation and PC movement in KotOR are now frame based as opposed to being event based. Also improved transitions between animations.


seedhartha:
Fixed.

DrMcCoy:
Merged as bf27b66e07390d9c07c76975134c7618fcf9f31a, thanks! :)

PR #234, opened 2018-05-07
KOTOR: Implement walkmesh load and display

seedhartha:
Partially ported Supermanu's walkmesh loader. Use console command
"tw" to toggle walkmesh display.


DrMcCoy:
There seems to be a segfault when switching modules while the walkmesh display is enabled.

```
#0  0x00007ffff6cb0389 in glDeleteLists () from /usr/lib64/opengl/nvidia/lib/libGL.so.1
#1  0x0000555555780cba in Engines::RenderableWalkmesh::~RenderableWalkmesh (this=0x7fffd28fd850, 
    __in_chrg=<optimized out>) at src/engines/aurora/renderwalkmesh.cpp:35
#2  0x00005555556c59c6 in Engines::KotOR::Area::~Area (this=0x7fffd28fcc60, __in_chrg=<optimized out>)
    at src/engines/kotor/area.cpp:72
#3  0x00005555556c5c49 in Engines::KotOR::Area::~Area (this=0x7fffd28fcc60, __in_chrg=<optimized out>)
    at src/engines/kotor/area.cpp:80
#4  0x00005555556c96e0 in Common::DeallocatorDefault::destroy<Engines::KotOR::Area> (x=<optimized out>)
    at ./src/common/deallocator.h:44
#5  Common::ScopedPtrBase<Engines::KotOR::Area, Common::DeallocatorDefault>::reset (o=0x0, 
    this=0x7fffdcb618c0) at ./src/common/scopedptr.h:88
#6  Engines::KotOR::Module::unloadArea (this=0x7fffdcb60c50) at src/engines/kotor/module.cpp:267
#7  Engines::KotOR::Module::unload (this=this@entry=0x7fffdcb60c50, 
    completeUnload=completeUnload@entry=false) at src/engines/kotor/module.cpp:230
#8  0x00005555556c9b0f in Engines::KotOR::Module::replaceModule (this=this@entry=0x7fffdcb60c50)
    at src/engines/kotor/module.cpp:298
#9  0x00005555556c9b98 in Engines::KotOR::Module::processEventQueue (this=0x7fffdcb60c50)
    at src/engines/kotor/module.cpp:428
#10 0x00005555556c3015 in Engines::KotOR::Game::runModule (this=0x7fffdcc52570)
    at src/engines/kotor/game.cpp:94
#11 0x00005555556c396c in Engines::KotOR::Game::run (this=this@entry=0x7fffdcc52570)
    at src/engines/kotor/game.cpp:74
#12 0x00005555556c15e8 in Engines::KotOR::KotOREngine::run (this=0x7fffdc000b20)
    at src/engines/kotor/kotor.cpp:127
#13 0x0000555555771538 in Engines::GameInstanceEngine::run (this=0x5555562d7010)
    at src/engines/enginemanager.cpp:222
#14 0x00005555557719f2 in Engines::EngineManager::run (this=0x55555638ada0, game=...)
    at src/engines/enginemanager.cpp:249
#15 0x0000555555771d97 in Engines::GameThread::threadMethod (this=0x555555ea42e0)
    at src/engines/gamethread.cpp:87
#16 0x000055555587e9ce in Common::Thread::threadHelper (obj=obj@entry=0x555555ea42e0)
    at src/common/thread.cpp:92
#17 0x00007ffff6f637bc in SDL_RunThread (data=0x55555638d0c0)
    at /var/tmp/portage/media-libs/libsdl2-2.0.8-r1/work/SDL2-2.0.8/src/thread/SDL_thread.c:283
#18 0x00007ffff6fca229 in RunThread (data=<optimized out>)
    at /var/tmp/portage/media-libs/libsdl2-2.0.8-r1/work/SDL2-2.0.8/src/thread/pthread/SDL_systhread.c:74
#19 0x00007ffff7bbf92a in start_thread (arg=0x7fffe97c9700) at pthread_create.c:465
#20 0x00007ffff3f0aa3f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

seedhartha:
I'm unable to reproduce the error. Could you recompile from the latest sources and give it another try?

DrMcCoy:
Still happens. A valgrind log is here: https://gist.github.com/DrMcCoy/8454e8e84927073b383f5dfb0dc68fce

Since the crash is within the GL, you're probably doing something with the GL resource you're not supposed to.

Looks to me like you're calling glDeleteLists() from within the game thread, not the main thread. This is not allowed. The GL context is only set up in the main thread, the one doing all the rendering and SDL event handling.

I had some IPC stuff set up for lists, so that you can tell the main thread that it should delete the list. Not sure if that still works, because we've since changed to use VBO/IBO instead, via our VertexBuffer and IndexBuffer classes. See for example the model stuff on that, or src/engines/nwn2/trxfile.cpp for something less complex.

Otherwise, src/graphics/aurora/cube.cpp still has lists, but again, not sure if that still works.

seedhartha:
Replaced display lists with glDrawElements and friends.

Supermanu:
Awesome work! Just one comment about the walkmesh. The end goal about pathfinding is to use two different walkmeshes, a global like the one found in the `wok` files and a local one for dynamic objects which will be a grid and generated on the fly. So it would be nice, if your walkmesh renderer could be generic enough to handle rectangle faces and not only triangles.

src/engines/kotor/area.cpp
```diff
@@ -53,9 +55,9 @@ namespace Engines {
 
 namespace KotOR {
 
-Area::Area(Module &module, const Common::UString &resRef) : Object(kObjectTypeArea),
-	_module(&module), _resRef(resRef), _visible(false), _activeObject(0), _highlightAll(false) {
-
+Area::Area(Module &module, const Common::UString &resRef)
+		: Object(kObjectTypeArea), _module(&module), _resRef(resRef),
+		  _visible(false), _activeObject(0), _highlightAll(false) {
```
DrMcCoy: Why are you rewrapping that?

src/engines/aurora/renderwalkmesh.cpp
```diff
@@ -0,0 +1,89 @@
...
+/** @file
+ *  Renders wrapped walkmesh for debug purposes.
+ */
+
+#include "src/engines/aurora/renderwalkmesh.h"
+
+namespace Engines {
+
+RenderableWalkmesh::RenderableWalkmesh()
+		: Renderable(Graphics::kRenderableTypeObject), _dirty(true) {
```
DrMcCoy: I'd prefer it if the parent class initialization was in the first line.

I.e
```
RenderableWalkmesh::RenderableWalkmesh() : Renderable(Graphics::kRenderableTypeObject),
	_dirty(true) {
````

That's how most of the existing code does it, especially when the line isn't too long.

File: src/engines/kotor/module.cpp [hunk 3]
```diff
@@ -515,27 +515,33 @@ void Module::handlePCMovement() {
...
 		float yaw = SatelliteCam.getYaw();
+		float newX, newY;
 
 		if (_forwardBtnPressed && !_backwardsBtnPressed) {
 			_pc->setOrientation(0, 0, 1, Common::rad2deg(yaw));
-			movePC(x - PC_MOVEMENT_SPEED * sin(yaw) * _frameTime,
-			       y + PC_MOVEMENT_SPEED * cos(yaw) * _frameTime,
-			       z);
+			newX = x - PC_MOVEMENT_SPEED * sin(yaw) * _frameTime;
+			newY = y + PC_MOVEMENT_SPEED * cos(yaw) * _frameTime;
```
DrMcCoy: I missed that for the previous PR, but in xoreos, we prefer `kPCMovementSpeed` over `PC_MOVEMENT_SPEED`. It should also be a `static const`, not just a `const`. Same goes for `ROTATION_SPEED` in satellitecamera.cpp.

src/engines/aurora/walkeleveval.cpp
```diff
@@ -0,0 +1,153 @@
...
+};
+
+struct Triangle {
+	Point v0, v1, v2;
+};
+
+inline Point add(const Point &u, const Point &v) {
+	return Point { u.x + v.x,
+	               u.y + v.y,
+	               u.z + v.z };
```
DrMcCoy: clang on the Travis CI VMs doesn't like that, as you can see here: https://travis-ci.org/xoreos/xoreos/jobs/375982631#L1954 . That's C++11 syntax anyway, right?

DrMcCoy:
I also don't see why the WalkmeshElevationEvaluator has to be a single-use class (with state, even). I'd just put that into functionality the Walkmesh class.

seedhartha:
Okay, that should do it. Korriban issues seem to be fixed now. As for the WalkmeshElevationEvaluator, what can I say, I'm a fan of SOLID :)

src/engines/kotor/walkmeshloader.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+void WalkmeshLoader::loadFromStream(Common::SeekableReadStream &stream) {
+	stream.seek(0);
+
+	// Header format:
+	//
+	// uint8[8] - WOK version
+	// uint32 - walkmesh type
+	// uint8[48] - reserved
+	// float[3] - position
```
DrMcCoy: You should check these, then, and throw if they're nonsensical. You should also probably catch throws (which can happen due to reading errors) where you're calling this from, too, and provide a `warning()` message.

If that WOK version is something like "WOK V1.0", you can use readUint32BE() and compare it with MKTAG('W', 'O', 'K', ' ').

Alternatively, the Aurora::AuroraFile base class provides functionality for that. My advice would have been putting this all in the KotOR walkmesh class, have it inherit from Walkmesh and Aurora::AuroraFile.

If you want to keep this as a separate class for some reason, there's no reason to create all this state. You immediately throw it away anyway. Like with the elevation evaluator, just have a `static WalkmeshData loadFromStream(Common::SeekableReadStream &stream)` method.

As is, your whole class design clashes with the current codebase. Please adapt your code to be more in line with how xoreos currently works, thanks.

src/engines/aurora/walkeleveval.h
```diff
@@ -0,0 +1,45 @@
...
+#include "src/engines/aurora/walkmesh.h"
+
+namespace Engines {
+
+class WalkmeshElevationEvaluator {
+public:
+	void setWalkmesh(Walkmesh *w);
+
+	/** Return elevation at given coordinates or FLT_MIN if can't walk there. */
+	float getElevationAt(float x, float y) const;
```
DrMcCoy: If you want to keep this as a separate class, at least make it a utility class without state. Have a `static float getElevationAt(Walkmesh &w, float x, float y)` method.

DrMcCoy:
Btw, @seedhartha, can you transfer all this to KotOR2 as well? :)

KotOR2 is still missing the character generator GUI stuff, too, though. Unless @Nostritius has something in progress there?

Nostritius:
Yes I already have the kotor2 chargen in progress, it's even almost usable except for some minor things.

seedhartha:
Sure. Had been thinking about it myself.

PR #236, opened 2018-05-08
KOTOR: Minimap

Nostritius:
Since walking is now working, i implemented a basic version of the ingame minimap.


DrMcCoy:
Neat! Merged as 67de5440...0db1800c, thanks! :)

Can you make the map arrow rotate too, or would that be more involved?

Nostritius:
Unfortunately this is quite more complex, than it looks. The problem is that guiquads are positioned quite different than regular scene objects, which makes rotating around their own center quite difficult. I have a solution in mind, but his will involve more groundwork. Additionally, I recognized that the arrow rotates with the camera, not with the character. A callback for this like "movePC" would be perfect.

DrMcCoy:
Ah, okay.

PR #237, opened 2018-05-08
KOTOR: Initial save/load game menu implementation

seedhartha:
Accesible from both main menu and ingame options menu.

### DIFF
--- a/src/common/filelist.h
+++ b/src/common/filelist.h
@@ -77,6 +77,11 @@ public:
 	 */
 	bool addDirectory(const UString &directory, int recurseDepth = 0);
 
+	/** Add subdirectories of a directory to the list.
+	 *  @param directory directory to scan
+	 */
+	bool addSubDirectories(const UString &directory);
+
 	/** Add files ending with the given string into another FileList.
 	 *
 	 *  @param  str A file ending to match file names against.
--- a/src/engines/aurora/gui.h
+++ b/src/engines/aurora/gui.h
@@ -75,8 +75,10 @@ protected:
 	GUI *_sub; ///< The currently running sub GUI.
 
 
-	/** Add a widget. */
-	void addWidget(Widget *widget);
+	/** Add a widget.
+	 *  @param show show it immediately
+	 */
+	void addWidget(Widget *widget, bool show = false);
 	/** Remove a widget. */
 	void removeWidget(Widget *widget);
 
--- a/src/engines/jade/gui/main/main.cpp
+++ b/src/engines/jade/gui/main/main.cpp
@@ -57,12 +57,12 @@ MainMenu::MainMenu(Module &module, ::Engines::Console *console) : ::Engines::Kot
 	 */
 	getListBox("ListBoxButtons")->setFill("");
 
-	addWidget(getListBox("ListBoxButtons")->createItem("NEW_GAME"));
-	addWidget(getListBox("ListBoxButtons")->createItem("LOAD_GAME"));
-	addWidget(getListBox("ListBoxButtons")->createItem("MINIGAMES"));
-	addWidget(getListBox("ListBoxButtons")->createItem("OPTIONS"));
-	addWidget(getListBox("ListBoxButtons")->createItem("CREDITS"));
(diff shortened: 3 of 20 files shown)


src/engines/kotor/gui/saveload.h
```diff
@@ -0,0 +1,55 @@
...
+#define ENGINES_KOTOR_GUI_SAVELOAD_H
+
+#include "src/engines/aurora/console.h"
+#include "src/engines/kotor/gui/gui.h"
+
+namespace Engines {
+
+namespace KotOR {
+
+enum : uint8 {
```
DrMcCoy: The typed enum is another C++11 thing we don't want, sorry. :P

seedhartha:
Fixed compilation error. Also implemented saved games list loading. There is a bug with listbox item size which I am not yet sure how to workaround. What does DIMENSION field in GUI GFF file stand for?

src/common/filelist.h
```diff
@@ -77,6 +77,11 @@ class FileList {
 	 */
 	bool addDirectory(const UString &directory, int recurseDepth = 0);
 
+	/** Add subdirectories of a directory to the list.
+	 *  @param directory directory to scan
+	 */
+	bool addSubDirectories(const UString &directory);
```
DrMcCoy: Could you separate adding this method into its own commit, please?

src/engines/aurora/gui.h
```diff
@@ -75,8 +75,11 @@ class GUI : boost::noncopyable {
 	GUI *_sub; ///< The currently running sub GUI.
 
 
-	/** Add a widget. */
-	void addWidget(Widget *widget);
+	/**
+	 * Add a widget.
+	 * @param show show it immediately
+	 */
+	void addWidget(Widget *widget, bool show = false);
```
DrMcCoy: Changing this method should also be its own commit.

src/engines/jade/gui/main/main.cpp
```diff
@@ -57,12 +57,12 @@ MainMenu::MainMenu(Module &module, ::Engines::Console *console) : ::Engines::Kot
...
-	addWidget(getListBox("ListBoxButtons")->createItem("MINIGAMES"));
-	addWidget(getListBox("ListBoxButtons")->createItem("OPTIONS"));
-	addWidget(getListBox("ListBoxButtons")->createItem("CREDITS"));
-	addWidget(getListBox("ListBoxButtons")->createItem("EXIT"));
+	addWidget(getListBox("ListBoxButtons")->createItem("NEW_GAME"), true);
+	addWidget(getListBox("ListBoxButtons")->createItem("LOAD_GAME"), true);
+	addWidget(getListBox("ListBoxButtons")->createItem("MINIGAMES"), true);
+	addWidget(getListBox("ListBoxButtons")->createItem("OPTIONS"), true);
+	addWidget(getListBox("ListBoxButtons")->createItem("CREDITS"), true);
+	addWidget(getListBox("ListBoxButtons")->createItem("EXIT"), true);
```
DrMcCoy: Why? At that point, the GUI itself isn't shown yet.

seedhartha: No idea, that how it worked before :)

DrMcCoy: Huh, am I missing something? You added the parameter `show`, and added `if (show) widget->show()`. So the old behaviour would be the same as `addWidget(..., false);`, no?

seedhartha: In *listbox.cpp*, function *createItem*, there was a non optional call to *show* on newly created item. It didn't work right for me, hence I moved that into *addWidget* function.

DrMcCoy: Ah, okay :)

seedhartha:
All done.

seedhartha:
Done.

DrMcCoy:
Merged as 2c50f7cc...b4ebfdf6, thanks! :)

farmboy0:
What problem do you have with the DIMENSION?

PR #239, opened 2018-05-10
ENGINES: Fix the loading of minimaps without a _

Nostritius:
fix the loading of maps which have no '_'Ä character, like the dantooine and manaan maps.


DrMcCoy:
Thanks! :)

PR #241, opened 2018-05-10
KOTOR2: Fix #240 for kotor2 chargen

### LINKED ISSUE #240: KOTOR/KOTOR2: Missing heads on several creature models
Several NPCs are missing heads. For example here in the module tar_m02ab.
20180510t111446

One draw of the rng in the character generator of KotOR2 even created a headless PC for me, though that might be a bug in the generator.

Nostritius:
This PR should fix #240 regarding the kotor2 character generation


DrMcCoy:
Thanks! :)

PR #242, opened 2018-05-10
TESTS: Undefine SDL main macro for Google Test

seedhartha:
Otherwise main gets replaced with SDL_main which leads to tests
compilation errors on Windows

### DIFF
--- a/tests/googletest/src/gtest_main.cc
+++ b/tests/googletest/src/gtest_main.cc
@@ -31,6 +31,8 @@
 
 #include "gtest/gtest.h"
 
+#undef main // negate -Dmain=SDL_main
+
 GTEST_API_ int main(int argc, char **argv) {
   printf("Running main() from gtest_main.cc\n");
   testing::InitGoogleTest(&argc, argv);


DrMcCoy:
Oh, right, yes, that might be a problem on Windows, I think? I never really got why SDL does that.

But shouldn't `#define SDL_MAIN_HANDLED` be the correct way for that?

seedhartha:
Yep, tests were not compilable on Windows using MinGW. Changed to SDL_MAIN_HANDLED macro.

seedhartha:
Not ready to merge yet.

seedhartha:
Neither defining SDL_MAIN_HANDLED in source files nor appending it to Automake rules helped. And SDL.h is not even explicitly included anywhere in Google Test. The issue is on Windows using MinGW.

DrMcCoy:
Does it help if you do include SDL.h (after defining SDL_MAIN_HANDLED, of course)?

seedhartha:
Does not work that way either.

DrMcCoy:
xoreos itself compiles correctly, though? What do the tests do differently? :/

seedhartha:
xoreos compiles just fine. I was unable to find the culprint but it has something to do with automake. When compiled from the command line (no automake) *libgtest_main.a* produced contains main entry point. When compiled using automake it has SDL_main instead.

DrMcCoy:
What do you mean, from the command line without automake?

seedhartha:
If instead of "make check" I invoke plain g++ and ar commands on same source files.

DrMcCoy:
If you add "V=1" to the make check command (and likewise the normal make for xoreos), it spits outs the actual compiler and linker invocations. That should help find out what it does wrong

seedhartha:
Turns out -Dmain=SDL_main was passed along with AM_CXXFLAGS macro. Is the following a better solution for this issue?

tests_googletest_libgtest_main_la_CXXFLAGS = $(GTEST_FLAGS) $(filter-out -Dmain=SDL_main,$(AM_CXXFLAGS))

DrMcCoy:
That would be a possible way, yes.

I'm curious though where it comes from (sdl2-config?) and why xoreos itself compiles fine.

seedhartha:
Done.

DrMcCoy:
As for why xoreos compiles but the tests do not: could be connected to how gtest_main declares main as GTEST_API_, which adds declspecs/attributes for library export visibility.

And yes, the -Dmain=SDL_main comes directly from the sdl2-config. My cross-compiling mingw has that as well. Not something we can change, unfortunately.

seedhartha:
Seems like too much trouble compared to a one-liner solution :P

DrMcCoy:
True, yeah :)

PR #243, opened 2018-05-12
KOTOR: Add selection circle

Nostritius:
This PR adds the selection circle, utilized by kotor to show, that an object is selected

### DIFF
--- a/src/engines/kotor/module.cpp
+++ b/src/engines/kotor/module.cpp
@@ -69,7 +69,7 @@ Module::Module(::Engines::Console &console) : Object(kObjectTypeModule),
 	_console(&console), _hasModule(false), _running(false),
 	_currentTexturePack(-1), _exit(false), _entryLocationType(kObjectTypeAll),
 	_fade(new Graphics::Aurora::FadeQuad()), _ingame(new IngameGUI(*this)),
-	_freeCamEnabled(false), _prevTimestamp(0), _frameTime(0),
+	_selected(0), _freeCamEnabled(false), _prevTimestamp(0), _frameTime(0),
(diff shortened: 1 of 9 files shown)


DrMcCoy:
To document how the selection circles work in KotOR:

There's three different selection circles in KotOR (names mine, dunno how the engine itself calls them):
1) Inactive inactive
2) Active active
3) Active and ready activeready

- Hovering over any object produces the inactive ones. Additional to the active marker
- Selecting objects via the next/previous keys (Q and E by default) set the active marker on items
- Objects that have been clicked on previously get an active marker. It's the same one moved by Q and E, not an additional one
- Hovering over the object with an active marker changes it into the active & ready one

---

Your PR here only handles the inactive marker. Seems to work correctly for placeables, at least.

But it should also do it for doors and creatures.

src/engines/kotor/module.cpp
```diff
@@ -448,6 +448,37 @@ void Module::clickObject(Object *object) {
...
+
+void Module::updateSelection() {
+	if (_selected) {
+		float x, y, z;
+		Placeable *placeable = ObjectContainer::toPlaceable(_selected);
+		if (placeable) {
+			placeable->getTooltipAnchor(x, y, z);
+
+			float sX, sY, sZ;
+			assert(GfxMan.project(x, y, z, sX, sY, sZ));
```
DrMcCoy: `assert()`s shouldn't have any side effects. They're macros that will be disabled completely when compiling with -DNDEBUG.

I.e. you need to save the result of the function call in a variables and assert on that.

```
const bool result = GfxMan.project(x, y, z, sX, sY, sZ);
assert(result);
```

Nostritius:
Closing this in favor of #409.

PR #245, opened 2018-05-12
COMMON: Deflate without knowing the output size

Nostritius:
This PR adds the possibility to decompress zlib data wihout knowing the exact size.

### DIFF
--- a/src/common/deflate.cpp
+++ b/src/common/deflate.cpp
@@ -29,43 +29,48 @@
 #include "src/common/deflate.h"
 #include "src/common/error.h"
 #include "src/common/scopedptr.h"
+#include "src/common/ptrvector.h"
 #include "src/common/memreadstream.h"
 
 namespace Common {
 
-byte *decompressDeflate(const byte *data, size_t inputSize,
-                        size_t outputSize, int windowBits) {
-
-	ScopedArray<byte> decompressedData(new byte[outputSize]);
-
+static void initZStream(z_stream &strm, int windowBits, size_t size, const byte *data) {
 	/* Initialize the zlib data stream for decompression with our input data.
 	 *
 	 * This ugly const cast is necessary because the zlib API wants a non-const
 	 * next_in pointer by default. Unless we define ZLIB_CONST, but that only
 	 * appeared in zlib 1.2.5.3. Not really worth bumping our required zlib
 	 * version for, IMHO. */
-
-	z_stream strm;
 	strm.zalloc   = Z_NULL;
 	strm.zfree    = Z_NULL;
 	strm.opaque   = Z_NULL;
-	strm.avail_in = inputSize;
+
+	strm.avail_in = size;
 	strm.next_in  = const_cast<byte *>(data);
 
 	int zResult = inflateInit2(&strm, windowBits);
-	BOOST_SCOPE_EXIT( (&strm) ) {
-		inflateEnd(&strm);
-	} BOOST_SCOPE_EXIT_END
-
 	if (zResult != Z_OK)
 		throw Exception("Could not initialize zlib inflate: %s (%d)", zError(zResult), zResult);
+}
+
+byte *decompressDeflate(const byte *data, size_t inputSize,
+                        size_t outputSize, int windowBits) {
+
+	ScopedArray<byte> decompressedData(new byte[outputSize]);
+
+	z_stream strm;
+	BOOST_SCOPE_EXIT( (&strm) ) {
+			inflateEnd(&strm);
+	} BOOST_SCOPE_EXIT_END
+
+	initZStream(strm, windowBits, inputSize, data);
 
 	// Set the output data pointer and size
 	strm.avail_out = outputSize;
 	strm.next_out  = decompressedData.get();
 
 	// Decompress. Z_FINISH, because we want to decompress the whole thing in one go.
-	zResult = inflate(&strm, Z_FINISH);
+	int zResult = inflate(&strm, Z_FINISH);
 
 	// Was the end of the input stream correctly reached?
 	if ((zResult != Z_STREAM_END) || (strm.avail_out != 0)) {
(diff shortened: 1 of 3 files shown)


src/common/deflate.cpp
```diff
@@ -81,6 +81,63 @@ byte *decompressDeflate(const byte *data, size_t inputSize,
...
+
+	do {
+		// Set the output data pointer and size
+		strm.avail_out = frameSize;
+		strm.next_out = decompressedDataBuffer.get();
+
+		// Decompress. Z_FULL_FLUSH, because we want to decompress partwise.
+		zResult = inflate(&strm, Z_FULL_FLUSH);
+		if (zResult != Z_STREAM_END && zResult != Z_OK)
+			throw Exception("Failed to inflate: %s (%d)", zError(zResult), zResult);
```
DrMcCoy: The throwing here (and below) will leak the contents of `buffers`

src/common/deflate.cpp
```diff
@@ -81,6 +81,63 @@ byte *decompressDeflate(const byte *data, size_t inputSize,
...
+
+		buffers.push_back(new byte[frameSize]);
+		if (strm.total_out % frameSize)
+			std::memcpy(buffers[buffers.size()-1], decompressedDataBuffer.get(), strm.total_out % frameSize);
+		else
+			std::memcpy(buffers[buffers.size()-1], decompressedDataBuffer.get(), frameSize);
```
DrMcCoy: Is that necessary? Can't you decompress directly into the `buffers` memory?

src/common/deflate.h
```diff
@@ -57,6 +57,20 @@ static const int kWindowBitsMaxRaw = -kWindowBitsMax;
...
+ *  @param  data       The compressed input data.
+ *  @param  inputSize  The size of the input data in bytes.
+ *  @param  outputSize The size of the decompressed output data.
+ *  @param windowBits  The base two logarithm of the window size (the size of
+ *                     the history buffer). See the zlib documentation on
+ *                     inflateInit2() for details.
+ *  @param frameSize   The size of the extracted frames, defaults to 4096.
+ *  @return The decompressed data
+ */
+byte *decompressDeflate2(const byte *data, size_t inputSize, size_t &outputSize,
```
DrMcCoy: Frankly, I'm not really happy about the name `decompressDeflate2`

src/common/deflate.cpp
```diff
@@ -81,6 +81,63 @@ byte *decompressDeflate(const byte *data, size_t inputSize,
...
+	 * next_in pointer by default. Unless we define ZLIB_CONST, but that only
+	 * appeared in zlib 1.2.5.3. Not really worth bumping our required zlib
+	 * version for, IMHO. */
+
+	z_stream strm;
+	strm.zalloc   = Z_NULL;
+	strm.zfree    = Z_NULL;
+	strm.opaque   = Z_NULL;
+	strm.avail_in = inputSize;
+	strm.next_in  = const_cast<byte *>(data);
```
DrMcCoy: Is it possible to pull all the inflate() stuff into its own common function, to minimize code duplication?

DrMcCoy: Or at least the initializing...

src/common/deflate.h
```diff
@@ -73,6 +87,19 @@ byte *decompressDeflate(const byte *data, size_t inputSize,
 SeekableReadStream *decompressDeflate(ReadStream &input, size_t inputSize,
                                       size_t outputSize, int windowBits);
 
+/** Decompress (inflate) using zlib's DEFLATE algorithm.
```
DrMcCoy: That should probably say what the difference to the function above is.

src/common/deflate.cpp
Nostritius: The problem is, that i don't know at the beginning how big the buffer should be.

DrMcCoy: But you know the max size, frameSize. So you decompress directly into frameSize and keep a separate vector how much of each buffer is filled

Nostritius: Oh, i recognized, this is another  piece of code, i initially thought it is. Simply ignore my last comment.

src/common/deflate.h
```diff
@@ -57,6 +57,20 @@ static const int kWindowBitsMaxRaw = -kWindowBitsMax;
...
+ *  @param  inputSize  The size of the input data in bytes.
+ *  @param  outputSize The size of the decompressed output data.
+ *  @param windowBits  The base two logarithm of the window size (the size of
+ *                     the history buffer). See the zlib documentation on
+ *                     inflateInit2() for details.
+ *  @param frameSize   The size of the extracted frames, defaults to 4096.
+ *  @return The decompressed data
+ */
+byte *decompressDeflateWithoutOutputSize(const byte *data, size_t inputSize, size_t &outputSize,
+						 int windowBits, unsigned int frameSize = 4096);
```
DrMcCoy: The alignment is off here now

src/common/deflate.h
```diff
@@ -73,6 +87,19 @@ byte *decompressDeflate(const byte *data, size_t inputSize,
...
+ *  @param  input      The compressed input data.
+ *  @param  inputSize  The size of the input data to read in bytes.
+ *  @param windowBits  The base two logarithm of the window size (the size of
+ *                     the history buffer). See the zlib documentation on
+ *                     inflateInit2() for details.
+ *  @param frameSize   The size of the extracted frames, defaults to 4096.
+ *  @return A stream of the decompressed data.
+ */
+SeekableReadStream *decompressDeflateWithoutOutputSize(ReadStream &input, size_t inputSize,
+                                       int windowBits, unsigned int frameSize = 4096);
```
DrMcCoy: The alignment is off here now

src/common/deflate.h
DrMcCoy: The name is more descriptive now, yeah. It's a bit clunky, but it's better than the first one :)

tests/common/deflate.cpp
```diff
@@ -87,6 +87,22 @@ GTEST_TEST(DEFLATE, decompressBuf) {
 	delete[] decompressed;
 }
 
+GTEST_TEST(DEFLATE, decompressBufWithoutDecompressedSize) {
+	static const size_t kSizeCompressed   = sizeof(kDataCompressed);
+	static const size_t kSizeDecompressed = strlen(kDataUncompressed);
+
+	size_t size = 0;
+	const byte *decompressed =
+		Common::decompressDeflateWithoutOutputSize(kDataCompressed, kSizeCompressed, size, Common::kWindowBitsMaxRaw);
```
DrMcCoy: Could you give it a low frameSize, so that it does at least do two iterations of that extract loop? Otherwise, we're not testing that this is actually working correctly.

src/common/deflate.cpp
```diff
@@ -33,39 +33,45 @@
 
 namespace Common {
 
-byte *decompressDeflate(const byte *data, size_t inputSize,
-                        size_t outputSize, int windowBits) {
-
-	ScopedArray<byte> decompressedData(new byte[outputSize]);
+void initZStream(z_stream &strm, int windowBits, size_t size, const byte* data);
```
DrMcCoy: You don't need a declaration here if the function itself follow immediately.

DrMcCoy: Also, that function should static.

DrMcCoy: Also also, it's `const byte *data`. :P

src/common/deflate.cpp
```diff
@@ -81,6 +87,55 @@ byte *decompressDeflate(const byte *data, size_t inputSize,
...
+
+	std::vector<byte *> buffers;
+
+	int zResult = 0;
+	do {
+		buffers.push_back(new byte[frameSize]);
+
+		// Set the output data pointer and size
+		strm.avail_out = frameSize;
+		strm.next_out = buffers[buffers.size()-1];
```
DrMcCoy: You can use [back()](http://en.cppreference.com/w/cpp/container/vector/back) instead.

DrMcCoy: On the other hand, isn't that rather a list you have here?

Also, isn't that something for our PtrList class, with a DeallocatorArray?

I.e. it could be a `Common::PtrList<byte, DeallocatorArray>`.

DrMcCoy: (In case you don't know, the PtrList provides a managed list with automatic deletion of elements. Something like a ScopedArray, but for a list of things. With the DeallocatorArray, you can put in arrays of things, i.e. things that need to be deallocated with delete[].)

Nostritius:

DrMcCoy:
Merged, thanks! :)

PR #246, opened 2018-05-12
DRAGONAGE2: Add a new probe for recognizing newer origin versions

Nostritius:
I recognized, that my version of Dragonage2 which I acquired from origin is not recognized, so i added another probe test which covers my version.


DrMcCoy:
Might be worth making it two separate probes. One for the retail version, one for the Origin version?

Nostritius:
There is an actual retail version of Dragon Age 2 without origin?

DrMcCoy:
Back in the day, the physical copies didn't require Origin, yeah. Possibly that these are not available anymore?

Nostritius:
Yes, it was very hard for me to find a used copy of DragonAge2. Nevertheless you're right, i will make a seperate probe for the origin version.

Nostritius:
I have now divided the probe

PR #247, opened 2018-05-13
Port my work from KotOR to KotOR 2

seedhartha:
Ported satellite camera, PC movement (w/ animations) and collision detection

### DIFF
--- a/src/engines/kotor2/module.cpp
+++ b/src/engines/kotor2/module.cpp
@@ -37,6 +37,7 @@
 #include "src/aurora/gff3file.h"
 
 #include "src/graphics/camera.h"
+#include "src/graphics/graphics.h"
 
 #include "src/graphics/aurora/textureman.h"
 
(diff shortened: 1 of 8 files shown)


DrMcCoy:
Merged, thanks! :)

src/engines/kotor2/module.cpp
```diff
@@ -318,6 +319,12 @@ void Module::enter() {
 		_pc->show();
 	}
 
+	Graphics::Aurora::Model *model = _pc->getModel().get();
+	model->clearDefaultAnimations();
+	model->addDefaultAnimation(Common::UString("pause3"), 25);
+	model->addDefaultAnimation(Common::UString("pause2"), 25);
+	model->addDefaultAnimation(Common::UString("pause1"), 50);
```
DrMcCoy: Btw, I'm not really a fan of having this getModel() method, here and in KotOR1. The Creature itself should add these default animations when a PC is created.

Likewise, there should be a playAnimation() method (see NWN does it, for example), and maybe even run()/stop() methods for running/stopping.

PR #248, opened 2018-05-15
KOTOR: Border color

Nostritius:
This PR implements border colors for kotor and kotor2 widgets

### DIFF
--- a/src/engines/kotor/gui/widgets/kotorwidget.cpp
+++ b/src/engines/kotor/gui/widgets/kotorwidget.cpp
@@ -48,7 +48,7 @@ KotORWidget::Extend::Extend() : x(0.0f), y(0.0f), w(0.0f), h(0.0f) {
 }
 
 
-KotORWidget::Border::Border() : fillStyle(0), dimension(0), innerOffset(0),
+KotORWidget::Border::Border() : fillStyle(0), dimension(0), innerOffset(0), hasColor(false),
 	r(0.0f), g(0.0f), b(0.0f), pulsing(false) {
 
 }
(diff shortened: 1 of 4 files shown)


DrMcCoy:
Some before/after pictures would have been nice.

From what I can see, in KotOR2:

20180515t121756
20180515t122005

20180515t121805
20180515t122013

And KotOR:

20180515t121731
20180515t122144

Not sure the tab coloring on the unselected tabs is correct, though?

There's also this, which happens if you close the ingame menu and then reopen the same tab you closed on:

20180515t122245

DrMcCoy:
(The PR looks okay otherwise to me, though)

DrMcCoy:
Oh, right, "kolors" in the commit message, though, should be "colors" :P

DrMcCoy:
Also, could you start capitalising class names in the commit message correctly as well? I.e. "BorderQuad" instead of "borderquad". So that it's clearer what you're talking about

Nostritius:
All issues should now be fixed. This includes the tab problem, i added handling if the border defines no color.

src/engines/kotor/gui/widgets/kotorwidget.cpp
```diff
@@ -328,6 +330,10 @@ void KotORWidget::load(const Aurora::GFF3Struct &gff) {
 
 	if (border.fill.empty())
 		_quad->setColor(0.0f, 0.0f, 0.0f, 0.0f);
+	else if (border.r == 0 && border.g == 0 && border.b == 0)
+		_quad->setColor(1.0f, 1.0f, 1.0f, 1.0f);
```
DrMcCoy: That's a bit...weird. So if the file specifies black, it's actually not?

Nostritius: Would it be better if i add a border.hasColor Flag?

DrMcCoy: Hmm, we need a better way to distinguish no between "no color defined" and "color defined, but it's black" here.

`KotORWidget::createBorder()` (which is the method filling in the border struct, yes?) should be extended. `struct Border` should get a `bool hasColor`, set to the return value of `b.hasField("COLOR")`.

Nostritius:
I have now changed the code to use a hasColor Flag.

DrMcCoy:
Merged as c19522df...b99404cc, thanks! :)

PR #251, opened 2018-05-16
COMMON: Change misleading method comment in Matrix4x4

seedhartha:


DrMcCoy:
Not sure this comment is better.

How about instead adding a line to the original comment with
"vout = vin * this"?

seedhartha:
Point is, that function multiplies a XYZ vector by a matrix, not a matrix by another matrix.

DrMcCoy:
Huh. You're right, I didn't even notice that. Weird

DrMcCoy:
Merged, thanks! :)

Btw, I've long since wanted xoreos to move to glm instead of having our own matrix/vector stuff.

Dunno if that sounds interesting to you, but that might be a task you could do, if you want:
- Pull in glm completely into xoreos, similar to how we pull in glew and utf8cpp, in a glm/ directory
- Convert all code that uses such maths to use glm
- Remove matrix4x4 and vector3 from xoreos
- Try to modify as little as possible from the glm files themselves, but fix compiler warnings (variable shadowing is a popular one)

No pressure, though, if you feel that's too annoying to do. :P

PR #252, opened 2018-05-16
COMMON: Replace Vector3 and Matrix4x4 with GLM classes

seedhartha:
Complete w/ Automake rules.


DrMcCoy:
Yeah, sorry, maybe I didn't make that clear, but the whole thing that changes xoreos to use glm should be one PR :P

seedhartha:
Done. Also updated unit tests.

glm/README.xoreos
```diff
@@ -5,3 +5,4 @@ it:
 
 - Modified include files to silence compiler warnings about variable
   shadowing
+- Commented out GLM_STATIC_ASSERT lines to silence Travis CI warnings
```
DrMcCoy: Hmm, a better, less intrusive way would be to undefine and redefine GLM_HAS_STATIC_ASSERT to 0 in setup.hpp.

However, it'd really like to keep these asserts. I'd rather disable `-Wunused-local-typedef`. See how we disable `-Wundefined-var-template` in CMakeLists.txt:104 and configure.ac:135 for an example on how that's done.

src/aurora/gff4file.cpp
```diff
@@ -28,11 +28,12 @@
 
 #include <cassert>
 
+#include <glm/gtc/type_ptr.hpp>
```
DrMcCoy: You need to use `#include ""` here, not `#include <>`.

DrMcCoy:
Also, the PR fails to compile for me with gcc 7.3.0.

```
In file included from ./glm/detail/type_vec4.inl:968:0,
                 from ./glm/detail/type_vec4.hpp:453,
                 from ./glm/detail/type_mat4x4.hpp:7,
                 from ./glm/mat4x4.hpp:6,
                 from ./src/graphics/graphics.h:33,
                 from src/xoreos.cpp:52:
./glm/detail/type_vec4_simd.inl: In static member function ‘static glm::tvec4<float, (glm::precision)5> glm::detail::compute_vec4_div<float, (glm::precision)5, true>::call(const glm::tvec4<float, (glm::precision)5>&, const glm::tvec4<float, (glm::precision)5>&)’:
./glm/detail/type_vec4_simd.inl:156:11: error: ‘struct glm::tvec4<float, (glm::precision)5>’ has no member named ‘data’
    Result.data = _mm_mul_ps(a.data, _mm_rcp_ps(b.data));
           ^~~~
./glm/detail/type_vec4_simd.inl:156:31: error: ‘const struct glm::tvec4<float, (glm::precision)5>’ has no member named ‘data’
    Result.data = _mm_mul_ps(a.data, _mm_rcp_ps(b.data));
                               ^~~~
./glm/detail/type_vec4_simd.inl:156:50: error: ‘const struct glm::tvec4<float, (glm::precision)5>’ has no member named ‘data’
    Result.data = _mm_mul_ps(a.data, _mm_rcp_ps(b.data));
                                                  ^~~~
./glm/detail/type_vec4_simd.inl: In constructor ‘glm::tvec4<T, P>::tvec4(T) [with T = float; glm::precision P = (glm::precision)5]’:
./glm/detail/type_vec4_simd.inl:367:3: error: class ‘glm::tvec4<float, (glm::precision)5>’ does not have any field named ‘data’
   data(_mm_set1_ps(_s))
   ^~~~
./glm/detail/type_vec4_simd.inl: In constructor ‘glm::tvec4<T, P>::tvec4(T) [with T = float; glm::precision P = (glm::precision)4]’:
./glm/detail/type_vec4_simd.inl:372:3: error: class ‘glm::tvec4<float, (glm::precision)4>’ does not have any field named ‘data’
   data(_mm_set1_ps(_s))
   ^~~~
./glm/detail/type_vec4_simd.inl: In constructor ‘glm::tvec4<T, P>::tvec4(T) [with T = float; glm::precision P = (glm::precision)3]’:
./glm/detail/type_vec4_simd.inl:377:3: error: class ‘glm::tvec4<float, (glm::precision)3>’ does not have any field named ‘data’
   data(_mm_set1_ps(_s))
```

150+ lines of that.

Also, it says
```
In file included from ./glm/detail/../detail/type_int.hpp:6:0,
                 from ./glm/detail/../fwd.hpp:6,
                 from ./glm/detail/type_mat4x4.hpp:6,
                 from ./glm/mat4x4.hpp:6,
                 from ./src/graphics/graphics.h:33,
                 from src/xoreos.cpp:52:
./glm/detail/../detail/setup.hpp:456:100: note: #pragma message: GLM: GCC older than 4.6 has a bug presenting the use of rgba and stpq components
 # pragma message("GLM: GCC older than 4.6 has a bug presenting the use of rgba and stpq components")
```

So I guess the gcc version check there is broken as well. Could be that it doesn't know what to do with gcc >= 7?

DrMcCoy:
I assume the message should read "preventing", not "presenting".

Do we use or need to use these component? Because my MacOS X cross-compiler is still stuck at 4.2.1 at the moment...

DrMcCoy:
I guess the compilation failure is related to https://github.com/g-truc/glm/issues/726 , fixed with https://github.com/g-truc/glm/commit/dd48b56e44d699a022c69155c8672caacafd9e8a .

Is that the most recent release of glm you have there? Could be that you need to pull in a git checkout instead of a stable release, then.

If you can, could you check if it behaves with gcc 8.1 as well afterwards? (If you can't, I can do it. I have gcc 8.1 ready on my machine)

seedhartha:
All issues should now be fixed.

1. No, we don't use rgba and stpq components atm.
2. Can't test for 8.1 but it should work from what I've seen.

seedhartha:
Should be fixed now.

DrMcCoy:
Can you squash that commit into the "COMMON: Replace Vector3 and Matrix4x4..." one? So as if you did it correctly in the first place.

There's another issue with Sonic, it seems, though: mouse/object interactions seem to be broken. I.e. mousing over an object in the master branch is detected (bounding box is shown, object name is printed on stderr, but that doesn't work in your PR.

Possibly something with the GraphicsManager::getWorldObjectAt() or GraphicsManager::unproject() function for the Orthogonal case?

seedhartha:
Fixed. Actually the problem was with how inverse matrix was calculated before (there was a case for which it returned identity matrix). Duplicated that behavior.

DrMcCoy:
Merged as d6ee0365...de5e634e, thanks! :)

I did modify 1c938690 to also include a small snipped in AUTHORS that mentions glm, and to add the revelant section in dists/debian/copyright.

PR #254, opened 2018-05-17
KOTOR2: Fix a graphical glitch in kotor2 borders

Nostritius:
This PR fixes a bug for small borders, where upper and lower elements overlap.


DrMcCoy:
Where exactly does this appear, so I can see for myself?

Nostritius:
bildschirmfoto zu 2018-05-18 12-54-14
For example you can see this in the kotor 2 class selection, in the upper both boxes as the bright stripe inside them.

PR #255, opened 2018-05-17
Fix appveyor builds, use vcpkg to install dependencies on Windows

berenm:
Hi, as the title says, this should fix the builds on Appveyor.

The dependencies are now pulled using `vcpkg` because it will probably do a better job at building these third party libraries on Windows and because it's available out of the box on Appveyor.

Also `libxvid` and `libfaad` aren't provided by `vcpkg` but if anyone wants to send https://github.com/Microsoft/vcpkg a PR with the corresponding port, they will probably be happy to integrate them.

Then there's a link error because of some global const variable in `libmad` not declared extern in `libmad` sources, so it's treated as a static. I have no clue why this was working before, maybe a different `libmad` version or compilation flags between what `xoreos` was doing and what `vcpkg` does. Anyway, i didn't really see the point of requiring this external global zero constant.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -19,6 +19,8 @@ environment:
     CONFIGURE_CMD: CC=cl CXX=cl ./configure --prefix=/
     INSTALL_CMD: make install
 
+    APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
   matrix:
     - CMAKE_GENERATOR: Visual Studio 14 2015
       MSVC_ARCH: amd64_x86
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,8 +10,8 @@ configuration: Debug
 
 environment:
   global:
-    BOOST_ROOT: C:\Libraries\boost_1_59_0
-    BOOST_INCLUDEDIR: C:\Libraries\boost_1_59_0
+    BOOST_ROOT: C:\Libraries\boost_1_66_0
+    BOOST_INCLUDEDIR: C:\Libraries\boost_1_66_0
     FREETYPE_DIR: C:\projects\xoreos\libs\freetype-2.3.5-1-lib
     LIBSDL2: SDL2-2.0.4
     LIBFT2: freetype-2.3.5-1
@@ -32,7 +32,7 @@ environment:
       MSBUILD_PLATFORM: Win32
       PROGRAM_FILES_DIR: C:\Program Files (x86)
 
-      BOOST_LIBRARYDIR: C:/Libraries/boost_1_59_0/lib32-msvc-14.0
+      BOOST_LIBRARYDIR: C:/Libraries/boost_1_66_0/lib32-msvc-14.0
 
       LIBOPENAL_DESTDIR: C:/Programs/OpenAL
       LIBSDL2_DESTDIR: C:/Programs/SDL2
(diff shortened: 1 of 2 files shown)


DrMcCoy:
Thanks so much! :D

One thing though: could you shorten the first line of the commit messages to fit within 70 characters? Namely, c2ebfe7d8ee77bd2fd85b2d0e7ce4333c65f0cc5 and 72159cdabe9d6a83c37bab90c45893cb6e78cfbb are too long. You can see that on GitHub as well, it wraps them if they're too long. Cheers.

berenm:
Done!

DrMcCoy:
Merged, thanks! :)

DrMcCoy:
@berenm, could it be that the liblzma this pulls in (using vcpkg, I guess, it is now?) links with a different libc than xoreos itself?

The issue is this: lzma_properties_decode() allocates memory (using malloc(), if no custom allocator was specified) and it expects the user to free it again.

Until now, I didn't free that memory at all, because the docs didn't really say that. I did talk to the liblzma developer Larhzu, though, and they confirmed the user is supposed to free that memory. Adding a free() on that memory works fine for me locally (and the LeakSanitizer complaints that original lead me to investigate this disappear), but in AppVeyor the unit test simple crashes then (see https://ci.appveyor.com/project/DrMcCoy/xoreos/build/0.0.4+AppVeyor.664#L5734 , for example).

Larhzu suggested that it might be a libc conflict, with liblzma using different malloc() structures from xoreos, so free()'ing something in xoreos that was malloc()'ed by liblzma then breaks.

liblzma doesn't provide a wrapper to free(), unfortunately, so we can't just call that. A workaround is to use a custom allocator, a struct filled with a function pointer to a function using xoreos' malloc(). And yes, that seem to work in AppVeyor. That's a bit hacky, though.

Is there otherwise a way to specify in AppVeyor to pull in a liblzma built with the same libc as xoreos is?

(I guess just compiling liblzma ourselves in the VM for each build might work, with the usual drawbacks that made you choose vcpkg instead to begin with, I assume. Changing it back isn't really a good option then either, yeah.)

Another possible other workaround would be to use lzma_raw_buffer_decode() in xoreos' decompressLZMA1() instead, which, I think, rolls properties decoding and decompressing into one function and, I assume, frees the filter options itself. I haven't tested that yet, though. I think I'd prefer that over using the custom allocator.

What do you think?

berenm:
Hello, it's possible, but I have no idea if this is the case. This is probably something the people behind vcpkg would know better.

IMHO your workaround is totally fine, I think that providing custom allocators to third-party library is a pretty common thing, and probably even a good thing to do in order to keep track of what's going on.

It's still a little bit worrisome if vcpkg introduces such issues, because it means that this may not be the only thing that's broken.

DrMcCoy:
Well, at least we might now know what the problem is should something similar to this break? AppVeyor just seems relatively fragile to me, in general. I really like having the AppVeyor builds in xoreos, but I wish it was more stable. Travis CI, as much as I complain about the outdated Ubuntu on their VMs, just seems to need far less maintaince, or at least so it seems to me.

There's also some intermittent "SSL connect error" when downloading the libraries. See for example here: https://ci.appveyor.com/project/DrMcCoy/xoreos/build/0.0.4+AppVeyor.668#L151

Triggering a rebuild makes it go away, so I guess that's just a temporary fluke that can't be fixed from our end?

berenm:
Yes, it didn't seem as polished as Travis to me either. But I believe it's more related to automation of Windows systems versus POSIX systems. Or maybe we don't do the right things... I'm no Windows expert either.

PR #256, opened 2018-05-18
AURORA: Texture atlas file

Nostritius:
This PR adds a class for loading texture atlas files found in dragonage and dragonage2.


src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+void TextureAtlasFile::getAtlasTexture(const Common::UString &texture,
+									   Common::UString &textureFile, float &x, float &y, float &w, float &h) const {
+	std::map<Common::UString, AtlasTexture>::const_iterator iter = _atlasTextures.find(texture);
+	if (iter == _atlasTextures.end()) {
+		throw Common::Exception("Atlas Texture %s not found", texture);
+	}
+
+	AtlasTexture atlasTexture = iter->second;
```
DrMcCoy: `const AtlasTexture &atlasTexture`. You don't need to copy the whole structure again.

tests/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,83 @@
...
+	EXPECT_EQ(y, 0.5f);
+	EXPECT_EQ(w, 0.5f);
+	EXPECT_EQ(h, 0.5f);
+
+	atl.getAtlasTexture("image1_2.dds", textureFile, x, y, w, h);
+	EXPECT_EQ(textureFile, "realimage2");
+	EXPECT_EQ(x, 0.0f);
+	EXPECT_EQ(y, 0.5f);
+	EXPECT_EQ(w, 0.25f);
+	EXPECT_EQ(h, 0.25f);
```
DrMcCoy: You can add another test with EXPECT_THROW for an invalid texture here.

tests/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  Unit tests for our AtlasTextureFile class.
+ */
+
+#include <src/common/memreadstream.h>
```
DrMcCoy: `#include ""`, please. And put that below the gtest include.

tests/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,83 @@
...
+		"OffsetAndScale_V2=\"0.0 0.5 0.5 0.5\" AtlasSize=\"2048 2048\" OriginalImageSize=\"100 100\"/>\n"
+		"\t</AtlasTexture>\n"
+		"\t<AtlasTexture Name=\"realimage2.dds\">\n"
+		"\t\t<SourceTexture Name=\"image1_2.dds\" OffsetAndScale=\"0.0 0.5 0.25 0.25\" "
+		"OffsetAndScale_V2=\"0.0 0.5 0.25 0.25\" AtlasSize=\"2048 2048\" OriginalImageSize=\"200 200\"/>\n"
+		"\t</AtlasTexture>\n"
+		"</AtlasData>\n";
+
+GTEST_TEST(TextureAtlasFile, isAtlasTexture) {
+	Common::MemoryReadStream *stream = new Common::MemoryReadStream(kTextureAtlasFile);
```
DrMcCoy: That's being to be leaked.

tests/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+	EXPECT_TRUE(atl.isAtlasTexture("image1_1.dds"));
+	EXPECT_TRUE(atl.isAtlasTexture("image2_1.dds"));
+	EXPECT_TRUE(atl.isAtlasTexture("image1_2.dds"));
+
+	EXPECT_FALSE(atl.isAtlasTexture("image2_2.dds"));
+}
+
+GTEST_TEST(TextureAtlasFile, getAtlasTexture) {
+	Common::MemoryReadStream *stream = new Common::MemoryReadStream(kTextureAtlasFile);
```
DrMcCoy: Leak here as well

src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,109 @@
...
+	const Common::XMLNode::Children &atlasTextures = node.getChildren();
+	for (Common::XMLNode::Children::const_iterator c = atlasTextures.begin(); c != atlasTextures.end(); ++c) {
+		if ((*c)->getName() != "AtlasTexture")
+			throw Common::Exception("Invalid tag, <AtlasTexture> expected, <%s> found", (*c)->getName());
+
+		Common::UString atlasTexture = (*c)->getProperty("Name");
+
+		const Common::XMLNode::Children &sourceTextures = (*c)->getChildren();
+		for (Common::XMLNode::Children::const_iterator c = sourceTextures.begin();
+			 c != sourceTextures.end(); ++c) {
```
DrMcCoy: Indent/alignment off

src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,109 @@
...
+			if ((*c)->getName() != "SourceTexture")
+				throw Common::Exception("Invalid tag, <AtlasTexture> expected, <%s> found", (*c)->getName());
```
DrMcCoy: Not AtlasTexture, SourceTexture

Nostritius:

DrMcCoy:
```
src/aurora/textureatlasfile.cpp: In member function ‘void Aurora::TextureAtlasFile::load(Common::SeekableReadStream&)’:
src/aurora/textureatlasfile.cpp:81:50: warning: declaration of ‘c’ shadows a previous local [-Wshadow]
   for (Common::XMLNode::Children::const_iterator c = sourceTextures.begin();
                                                  ^
src/aurora/textureatlasfile.cpp:74:49: note: shadowed declaration is here
  for (Common::XMLNode::Children::const_iterator c = atlasTextures.begin(); c != atlasTextures.end(); ++c) {
```

src/aurora/textureatlasfile.h
```diff
@@ -0,0 +1,55 @@
...
+namespace Aurora {
+
+class TextureAtlasFile {
+public:
+	TextureAtlasFile(Common::SeekableReadStream &stream);
+	TextureAtlasFile(Common::UString name);
+
+	bool isAtlasTexture(const Common::UString &texture) const;
+	void getAtlasTexture(const Common::UString &texture,
+                         Common::UString &textureFile, float &x, float &y, float &w, float &h) const;
```
DrMcCoy: No. You need to indent with one tab, then use spaces to align the rest

src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,109 @@
...
+	const Common::XMLNode::Children &atlasTextures = node.getChildren();
+	for (Common::XMLNode::Children::const_iterator c = atlasTextures.begin(); c != atlasTextures.end(); ++c) {
+		if ((*c)->getName() != "AtlasTexture")
+			throw Common::Exception("Invalid tag, <AtlasTexture> expected, <%s> found", (*c)->getName().c_str());
+
+		Common::UString atlasTexture = (*c)->getProperty("Name");
+
+		const Common::XMLNode::Children &sourceTextures = (*c)->getChildren();
+		for (Common::XMLNode::Children::const_iterator c = sourceTextures.begin();
+             c != sourceTextures.end(); ++c) {
```
DrMcCoy: Here too. Use tabs to indent to the correct level of the for (i.e. two tabs here), then spaces to align the rest

Nostritius:
This should now fix the last problems

src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,109 @@
...
+			Common::UString::split(offsetAndScale, ' ', values);
+
+			atlasTexture.erase(atlasTexture.findFirst(".dds"), atlasTexture.end());
+
+			AtlasTexture texture;
+			texture.textureFile = atlasTexture;
+			Common::parseString(values[0], texture.x);
+			Common::parseString(values[1], texture.y);
+			Common::parseString(values[2], texture.w);
+			Common::parseString(values[3], texture.h);
```
DrMcCoy: You need to check here that values does contain at least (or even exactly) 4 elements. If the atlas file is malformed and `Common::UString::split()` creates less than four entries, accessing the missing elements will result in Undefined Behaviour.

Nostritius:
I have now added a check for the values

src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,111 @@
...
+	w = atlasTexture.w;
+	h = atlasTexture.h;
+}
+
+void TextureAtlasFile::load(Common::SeekableReadStream &stream) {
+	Common::XMLParser parser(stream);
+
+	const Common::XMLNode &node = parser.getRoot();
+
+	if(node.getName() != "AtlasData")
```
DrMcCoy: Since you're comparing strings here, you should probably do these checks case-insensitively. Just so that "ATLASDATA" won't mess you up.

In fact, you can just give the `Common::XMLParser` a true for the second parameter (`makeLower`, and yes, that should probably be better documented), and then just do all string comparisons with the lowercase version (i.e. "atlasdata", in this case).

src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,111 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file
+ * A loader class for texture atlas xml files for dragonage and dragonage2
```
DrMcCoy: Can you add a period at the end of the Doxygen comment here (and in the .h)?

Also, please capitalize XML and write the full game names, i.e. "Dragon Age: Origins" and "Dragon Age II".

src/aurora/textureatlasfile.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+		const Common::XMLNode::Children &sourceTextures = (*a)->getChildren();
+		for (Common::XMLNode::Children::const_iterator s = sourceTextures.begin();
+		     s != sourceTextures.end(); ++s) {
+			if ((*s)->getName() != "SourceTexture")
+				throw Common::Exception("Invalid tag, <SourceTexture> expected, <%s> found", (*s)->getName().c_str());
+
+			Common::UString sourceTexture = (*s)->getProperty("Name");
+			Common::UString offsetAndScale = (*s)->getProperty("OffsetAndScale");
+			Common::UString offsetAndScaleV2 = (*s)->getProperty("OffsetAndScale_V2");
```
DrMcCoy: Can you make these const? I like it being explicit which variables are read-only and which are written to.

src/aurora/textureatlasfile.cpp
Nostritius: But every possible atlas is writing them as "AtlasData". So I do not see the point in making this check less strict when it already covers all possible atlas files.

DrMcCoy: Eh, fair enough. We'll see if there's mods that depends on the game being case-insensitive there. And yeah, I'm not even sure the game is. :P

Nostritius: At least i can assure you, that the game does it. I iterated over all atlas xml files with this code and there never was a problem.

Nostritius:
The issues should now be corrected.

PR #257, opened 2018-05-20
ENGINES: Implement conversations/cutscenes for KotOR games

seedhartha:
Conversations are started from script or by clicking on characters. Implemented GUI and voice over for both games.

### DIFF
--- a/src/aurora/dlgfile.cpp
+++ b/src/aurora/dlgfile.cpp
@@ -226,6 +226,8 @@ void DLGFile::readEntry(const GFF3Struct &gff, Entry &entry) {
 
 	entry.line.sound = gff.getString("Sound");
 
+	entry.line.voice = gff.getString("VO_ResRef");
+
 	entry.line.animation = gff.getUint("Animation", 0);
 
 	entry.line.quest      = gff.getString("Quest");
--- a/src/engines/kotor/gui/widgets/kotorwidget.cpp
+++ b/src/engines/kotor/gui/widgets/kotorwidget.cpp
@@ -253,6 +253,10 @@ float KotORWidget::getHeight() const {
 	return _height;
 }
 
+float KotORWidget::getTextHeight(const Common::UString &text) const {
+	return _text ? _text->getHeight(text) : 0.0f;
+}
+
 void KotORWidget::setFont(const Common::UString &fnt) {
 	_text->setFont(fnt);
 }
--- a/src/engines/kotor/gui/widgets/kotorwidget.cpp
+++ b/src/engines/kotor/gui/widgets/kotorwidget.cpp
@@ -224,6 +224,17 @@ bool KotORWidget::isHighlight() {
 	return _highlighted;
 }
 
+void KotORWidget::setSize(float width, float height) {
+	if (_quad) {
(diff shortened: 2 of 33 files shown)


DrMcCoy:
Heh, seems like there is directing information in the KotOR2 conversations, marked with {} (we should probably remove those, but it might be valuable having a command line switch and config option (command line switches automatically map to config keys) to enable them).

20180520t173712
20180520t173719
20180520t173831
20180520t173853
20180520t174047

seedhartha:
Hidden developer notes in KotOR 2 dialogs. Command line option could wait, I think.

seedhartha:
All issues should now be fixed except for animal and droid sounds in KotOR 2. Apparently the original game ignores voice ResRef on those and instead selects a random sound from StreamVoice/AVO folder based on animal race and emotion type. I believe that is an issue for another PR.

src/engines/kotor/gui/widgets/kotorwidget.cpp
```diff
@@ -365,6 +365,14 @@ void KotORWidget::setText(const Common::UString &text) {
 	_text->setText(text);
 }
 
+void KotORWidget::setHorizontalTextAlign(float halign) {
+	_text->setHorizontalAlign(halign);
```
DrMcCoy: You should wrap these in `if (_text)`.

And yes, all those other places here that access _text (and _quad) should check them as well, like setText() and setFont(). That slipped me by, it seems.

Feel free to do these with commits in this PR or a new PR, otherwise I'll do that once this PR is merged.

src/engines/kotor/gui/widgets/label.cpp
```diff
@@ -39,10 +39,26 @@ WidgetLabel::WidgetLabel(::Engines::GUI &gui, const Common::UString &tag) :
 WidgetLabel::~WidgetLabel() {
 }
 
+void WidgetLabel::enableHighlight() {
+	Graphics::Aurora::Highlightable *highlightable = getTextHighlightableComponent();
```
DrMcCoy: For safety's sake, add a check here if this is 0.

src/aurora/dlgfile.cpp
```diff
@@ -226,6 +226,9 @@ void DLGFile::readEntry(const GFF3Struct &gff, Entry &entry) {
 
 	entry.line.sound = gff.getString("Sound");
 
+	if (gff.hasField("VO_ResRef"))
+		entry.line.voice = gff.getString("VO_ResRef");
```
DrMcCoy: You don't need that `if` here. `GFF3Struct::getString()` returns its second parameter (defaulting to "") when the field doesn't exist.

You only need to check beforehand on `GFF3Struct::getStruct()` and `GFF3Struct::getList()`. These throw when no such field exists.

DrMcCoy:
Can you squash the bits of the last fix-up commits that change something you modify in this PR as well into the original commits? As if that was your original intent in the first place.

seedhartha:
Is that sufficient?

DrMcCoy:
Yes. Merged as a975a627...bdcfbbba, thanks! :)

PR #259, opened 2018-05-22
ENGINES: Implement triggers for KotOR games

seedhartha:
Triggers can be displayed using console command "tt".


DrMcCoy:
Hmm, can you make the console command longer, less ambigous? Something like "showtriggers" would make more sense.

The same for the one showing the walkmesh ("showwalkmesh") and the free camera ("freelook", maybe?). If you think that's too much typing, we do have tab-completion in the console.

You might even spit out something in the console displaying the current status. I.e. "showtriggers" says "Now showing triggers." / "Now hiding triggers.", or something to that effect.

seedhartha:
Renamed new console commands.

DrMcCoy:
Merged, thanks! :)

PR #263, opened 2018-05-26
BUILD: Add support for osx to travis file

Nostritius:
This PR extends the ability of the .travis.yml file to build and test xoreos under osx too. I also fixed a small cmake issue, to get it to build with cmake.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,16 @@ compiler:
   - gcc
   - clang
 
+os:
+  - linux
+  - osx
+
 env:
   global:
     # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
     #   via the "travis encrypt" command using the project repo's public key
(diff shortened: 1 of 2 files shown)


DrMcCoy:
You need to extend the coverity_scan_run_condition too, so only Linux builds evaluate to true (is there an env variable to check there?). Otherwise, pushing to the coverity_scan branch will upload the Coverity Scan build twice.

Also, their gcc version on OS X is ancient, 4.2.1. That's the same version I have for my OS X cross-compiler. If you move over the config to Phaethon (and I would like you to do that, also to xoreos-tools), you need to disable gcc on OS X, because that won't support the C++14 features needed for Phaethon.

DrMcCoy:
Also, what is
```
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.7.0
```

What is that in real clang version numbers?

DrMcCoy:
Ugh. https://stackoverflow.com/questions/33603027/get-apple-clang-version-and-corresponding-upstream-llvm-version

Seems like these version numbers are useless. Thanks Apple.

Will the clang there do C++14, as required by Phaethon? Who knows.

Nostritius:
If this does not work with c++14, it is possible to increase the xcode version to get a newer compiler than 8.1.0

Nostritius:
I have added a condition which excludes building on osx and added a condition for the coverity scan condition for osx.

Nostritius:
I have now changed the coverity scan condition

DrMcCoy:
Merged as 5afa29d4...f718f5eb, thanks! :)

PR #264, opened 2018-05-26
KotOR update

seedhartha:
This PR:
1. Fixes collision detection bug in module 262tel by using GLM for intersection tests.
2. Fixes Bastila model animations bug visible in module unk_m44ac.
3. Adds initial scroll bar implementation in dialogs and save load menu (display only).
4. Implements local variables functions and isObjectPartyMember function.
5. Shows intersected face on the walkmesh.
6. Fixes text colors in dialogs and save load menu.
7. Fixes container issues.


DrMcCoy:
Merged, thanks! :)

PR #266, opened 2018-05-27
ENGINES: Load area camera style in KotOR games

seedhartha:


DrMcCoy:
Hej, could you please add a comment when you change your PR, so that I know what you changed and when you did? Thanks.

And please don't update the PR just to rebase it to the current master. I can do that myself, so it's not necessary. In fact, it even confuses me when it's happening while I'm reviewing the PR.

In general, when you submit a PR, you should be confident that it's ready to be merged. When I look through it and I find it acceptable, I merge it. Otherwise, I add comments and request changes. When you then change your PR, add a comment too so I know I can review it again.

DrMcCoy:
Hmm, as for the rendering of only the adjacent rooms. Good idea, that, yes.

However:
- You probably should also hide/show objects within the rooms (and animations should pause for them, I guess)
- Could you add a debug console switch to disable this? And/or connecting that to flycam on/off, too

seedhartha:
Okay.

1. I already pause animations and hide objects.
2. Will do.

DrMcCoy:
Okay, then there's a bug somewhere :P

20180528t095130

seedhartha:
Added command line option "flycamallrooms" which will show all rooms when in free roam camera mode (it is enabled by default). As for the bug, I have a feeling that this particular room is the one with the skybox which makes it adjacent to all others.

DrMcCoy:
It only shows the objects in that room, not the room model itself.

20180528t103603
20180528t103634

seedhartha:
Nope. The room those objects are in and that curvy hallway are two separate rooms :P

DrMcCoy:
Huh, okay. Weird.

seedhartha:
Please don't merge just yet. I think I found another bug related to skybox.

seedhartha:
I now take not the first, but the smallest room that contains the player. This fixes the Endar Spire bug and many modules on KotOR II.

DrMcCoy:
Hmm, just the adjacent rooms might not be enough?

20180528t124331
20180528t124355

They pop in a bit further down the corridor, but still within the visible range.

DrMcCoy:
Also very bad in the upper and lower Taris cantinas tar_m02ae and tar_m03ae.

seedhartha:
Scratch that, came up with a better idea :D Animation smoothness now depends on how far a model is from the camera (number of skipped animation frames = distance / 4). Doesn't improve FPS, but does improve the overall feel.

DrMcCoy:
Hmm, wait, I didn't check, did you use the VIS file at all? That one has a visibility matrix which rooms are visible from each room. That's what the original game, I think, uses for that. It's in `Area::_vis`.

You only spoke about neighbouring rooms. Is that what you meant, or did you try something else?

I think I remember also experimenting with hiding certain rooms. Depending on just where the camera is, because that was before I even loaded objects. The lower Taris cantina was my main testing area, and somehow it never worked right. Might be I messed up back in the day, or maybe the data is just wonky.

seedhartha:
I only rendered direct neighbors in that implementation. It did occur to me that visibility might be precalculated but I couldn't find such files in Kotor Tool. Will try this approach as well.

farmboy0:
Using the vis files is the correct approach. It will tell you exactly which rooms need to be rendered based on which room the PC is in.

seedhartha:
Got it working with VIS files. Added console commands "getpcroom", "listroomsvisiblefrom" and a command line option "flycamallrooms".

PR #267, opened 2018-05-27
KOTOR: Replaces pc appearence with the actual armorless outfit

Nostritius:
After I found out, how the actual skin textures are called, i replaced the pc standard outfit body with the standard no-outfit body.


DrMcCoy:
Merged as 88005a8d, thanks! :)

PR #269, opened 2018-05-27
KOTOR2: Ingame hud

Nostritius:
This PR implements the basic hud for the kotor 2 ingame gui.


DrMcCoy:
Hmm, 800x600 looks okay:

20180528t085851

But the other resolutions are very white:

20180528t085808

Nostritius:
I have now removed bigger resolutions than 800x600. I believe now, that the hud is dynamically scaled up as the othe guis did. The bigger hud gui files seem to be some leftover from kotor 1.

PR #270, opened 2018-05-27
KOTOR: Replace static checkbox images

### LINKED ISSUE #265: KOTOR: Checkbox textures are hard-coded
I believe this was my doing here:
https://github.com/xoreos/xoreos/blob/master/src/engines/kotor/gui/widgets/checkbox.cpp#L97

Before clicking

After clicking

Nostritius:
This PR replaces the static checkbox images with the ones defines in the gff file. This should fix #265.


DrMcCoy:
Merged as 88005a8d...98ca935c, thanks! :)

PR #272, opened 2018-05-28
JADE: Play background music on main menu

dressupgeekout:
I mean, it works, I dunno :}

### DIFF
--- a/src/engines/jade/game.cpp
+++ b/src/engines/jade/game.cpp
@@ -28,6 +28,8 @@
 
 #include "src/events/events.h"
 
+#include "src/sound/sound.h"
+
 #include "src/engines/aurora/util.h"
 
 #include "src/engines/jade/game.h"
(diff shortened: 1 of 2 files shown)


DrMcCoy:
Hmm, could you rather save that ChannelHandle and use `SoundMan.stopChannel()`? Put this into their own methods, `playMenuMusic()` and `stopMenuMusic` too, please.

I.e. similar to how KotOR does it: https://github.com/xoreos/xoreos/blob/master/src/engines/kotor/game.cpp#L105

dressupgeekout:
Oh heh that's definitely cleaner.

dressupgeekout:
OK let's try this. I decided that `Jade::Game::playMenuMusic()` shouldn't accept a parameter to  let the caller decide _which_ menu music to play, because, unlike KOTOR, there's only one menu music (even during character creation and stuff like that)

src/engines/jade/game.cpp
```diff
@@ -100,6 +102,17 @@ void Game::runModule() {
 	_module->clear();
 }
 
+void Game::playMenuMusic()
+{
```
DrMcCoy: That opening brace needs to go into the previous line, please.

src/engines/jade/game.cpp
```diff
@@ -100,6 +102,17 @@ void Game::runModule() {
...
+	stopMenuMusic();
+	_menuMusic = playSound("background", Sound::kSoundTypeMusic, true);
+}
+
+void Game::stopMenuMusic()
+{
```
DrMcCoy: Here too

dressupgeekout:
OK, here we go

DrMcCoy:
Merged as b9555a11844a5e0cde6641b4887c90961f01ecf9, thanks! :)

PR #275, opened 2018-05-28
BUILD: CMake installation

Nostritius:
This PR is my second try on adding the possibiltiy to install xoreos over cmake. I modified CMakeAM.cmake to read additionally docs and manpages. The programs, man pages and docs aree then installed to the gnu directories.


DrMcCoy:
Is there a way we can provide a "make uninstall" as well?

Nostritius:
Hmm, we could add a way to read the installed files from install_manifest.txt, generate a script from it which can be called over a custom target.

Nostritius:
I have now added a target and a script to execute an uninstallation based on the source of my previous message, but slightly modified to also remove also empty directories, like doc/xoreos.

doc/rules.mk
```diff
@@ -0,0 +1,46 @@
...
+    README.md \
+    FAQ.md \
+    doc/xoreos.conf.example \
+    $(EMPTY)
+
+# Documents for contributors
+dist_doc_DATA += \
+    CONTRIBUTING.md \
+    CODE_OF_CONDUCT.md \
+    $(EMPTY)
```
DrMcCoy: You're not including this file anywhere in the autotools build.

Please make sure that "make dist" and "make install" for automake still work the same with your changes.

Nostritius:
I added the mising include.

cmake/CMakeAM.cmake
```diff
@@ -249,12 +249,24 @@ function(parse_automake AM_FILE_NAME)
     list(APPEND AM_PROGRAMS ${AM_TARGET})
   endforeach()
 
+  set(AM_MANS)
+  foreach(AM_MAN ${dist_man1_MANS})
```
DrMcCoy: My bad, that should have been man6, not man1. But you should probably evaluate both, because of xoreos-tools.

But also, the man pages aren't installed in the correct directory. CMake installs them into `man/xoreos.6` now, but it should be `man/man6/xoreos.6`. Compare make install from automake with make install from cmake.

cmake/CMakeUninstall.cmake
```diff
@@ -0,0 +1,57 @@
+# xoreos - A reimplementation of BioWare's Aurora engine
```
DrMcCoy: You need to add that file into EXTRA_DIST in the top-most rules.mk (where the other CMake files are listed).

Nostritius:
man1 and man6 pages will now be imported and the CMakeUninstall.cmake file will now be added to the EXTRA_DIST_FILES

DrMcCoy:
Did you remove the "include doc/rules.mk" again? O_o

Also, when I install, using CMake, xoreos with `make DESTDIR=/home/drmccoy/fooc/ install`, the files do get correctly put into /home/drmccoy/fooc/ (though inside the usr/local/ subdirectory), but make uninstall doesn't correctly work.

1) Straight-up `make uninstall` tries to delete /usr/local/bin/xoreos
2) `make DESTDIR=/home/drmccoy/fooc/ uninstall` deletes /home/drmccoy/fooc/usr/local/bin/xoreos and the correct xoreos.6 manpage, but want's to remove the directory /usr/local/share/man/man6

```
-- Uninstalling /home/drmccoy/fooc//usr/local/bin/xoreos
-- Uninstalling /home/drmccoy/fooc//usr/local/share/man/man6/xoreos.6
-- Removing Directory "/usr/local/share/man/man6"
CMake Error at CMakeUninstall.cmake:54 (message):
  Problem when removing /usr/local/share/man/man6


make[3]: *** [CMakeFiles/uninstall.dir/build.make:57: CMakeFiles/uninstall] Error 1
make[2]: *** [CMakeFiles/Makefile2:2266: CMakeFiles/uninstall.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2273: CMakeFiles/uninstall.dir/rule] Error 2
make: *** [Makefile:877: uninstall] Error 2
```

Nostritius:
However I managed to delete the include, i readded it. The cause of the other problem was, that  the DESTDIR Variable was handled for the deletion of files, but not for the deletion of folders. I fixed this.

DrMcCoy:
Merged, thanks! :)

If you could apply this to xoreos-tools and phaethon as well, that would be great.

PR #281, opened 2018-06-02
KOTOR2: Add GetObjectByTag nwscript method

Nostritius:
Adds the nwscript function GetObjectByTag


DrMcCoy:
Merged as d79266c8, thanks! :)

PR #282, opened 2018-06-03
KotOR facial animations

seedhartha:
This PR introduces facial animations for KotOR games. Console command "playanim" could be used to test those, e.g. "playanim tlknorm talk" when a character is highlighted.

### DIFF
--- a/src/graphics/aurora/model_kotor.cpp
+++ b/src/graphics/aurora/model_kotor.cpp
@@ -447,12 +447,16 @@ void ModelNode_KotOR::load(Model_KotOR::ParserContext &ctx) {
 	uint32 controllerDataOffset, controllerDataCount;
 	Model::readArrayDef(*ctx.mdl, controllerDataOffset, controllerDataCount);
 
-	std::vector<float> controllerData;
+	std::vector<float> controllerDataFloat;
 	Model::readArray(*ctx.mdl, ctx.offModelData + controllerDataOffset,
-	                 controllerDataCount, controllerData);
+	                 controllerDataCount, controllerDataFloat);
+
+	std::vector<uint32> controllerDataInt;
+	Model::readArray(*ctx.mdl, ctx.offModelData + controllerDataOffset,
+	                 controllerDataCount, controllerDataInt);
 
 	readNodeControllers(ctx, ctx.offModelData + controllerKeyOffset,
-	                    controllerKeyCount, controllerData);
+	                    controllerKeyCount, controllerDataFloat, controllerDataInt);
 
 	if ((flags & 0xFC00) != 0)
 		throw Common::Exception("Unknown node flags %04X", flags);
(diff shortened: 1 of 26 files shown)


DrMcCoy:
You should really stick the animation playing into the Object hierarchy, like NWN does with  `virtual void Object::playAnimation()` (just that it would take two animation names, for the body and the head, in KotOR/KotOR2) and get rid of `Creature::getModel()`.

DrMcCoy:
The talking seems to be a bit too fast?

seedhartha:
Moved animation methods inside the Object hierarchy, slowed facial animation speed.

DrMcCoy:
Oh, and some of the NPCs seem to just shift around in t-pose. Could it be that some models have different default animations than others?

seedhartha:
I'm looking into it. Those models have their own pause animations which are broken apparently (?).

DrMcCoy:
The bot animations are also broken, maybe that's connected?

DrMcCoy:
Seems like protocol droids have cpause1 and cpause2 as idle animations instead. Both mechanical droids and protocol droids play animations correctly now, in principle.

They're just wrongly positioned on the ground. Utility droids float above the ground, while protocol droids sink into the ground when I call playanim on them.

src/graphics/aurora/model_kotor.cpp
```diff
@@ -557,7 +555,28 @@ void ModelNode_KotOR::readOrientationController(uint8 columnCount, uint16 rowCou
 		uint16 timeIndex, uint16 dataIndex, std::vector<float> &data) {
 	switch (columnCount) {
 		case 2:
-			// TODO: 2 column orientation controller
+			for (int r = 0; r < rowCount; r++) {
+				QuaternionKeyFrame q;
+				q.time = data[timeIndex + r];
+
+				uint32 temp = *reinterpret_cast<uint32 *>(&data[dataIndex + r]);
```
DrMcCoy: Yeah, that's not really portable, unfortunately. Potentially, the data might be in a different endianness (float endianness can differ from integer endianness) and even a different format than IEEE (*), so we can't necessarily guarantee that it's the same bit-pattern anymore.

You will probably need to rewrite the loading of the controller data altogether. Read in a byte array, and read the correct data from that here (using the READ_LE_UINT32() etc. macros and convertIEEEFloat()).

(*) If we supported that in readIEEEFloatLE(), which we don't yet, yes

seedhartha:
Is that better? To my understanding once the data in loaded by the processor endianness is completely irrelevant to bitwise operations.

DrMcCoy:
In principle the endianness doesn't matter, yes, most of the time. But you were doing a reinterpret_cast on the data, wanting to reinterpret a bit pattern from a float into a uint32. In general, we want to avoid this, because it's not portable.

It will work on your common x86/x86_64 architecture, but not necessarily on all architectures. There are architectures out there where the endianness of a float is different from the endianness of a uint32. So you need to flip the byte order around then before reading an uint32 out of the float bit pattern.

Also, the float format of the CPU might be different than IEEE, in which case reading the float in could have already changed the bit pattern, making the reinterpret_cast not work either.

Yes, strictly speaking, we don't support that yet anyway, so such systems will fail all over the place. But there's exactly one place we'd need to rewrite/expand for that, because we abstracted the reading of IEEE floats away (`convertIEEEFloat()` in src/common/util.cpp) (well, unless other wrong conversion such as these slipped into the codebase :P). I'd like to keep that at one controllable point, so that we can in the future do that correctly for these systems.

In either case, the new approach is okay now. Merged, thanks! :)

clone2727:
@seedhartha Just for an example, there are some [ARM platforms out there](https://wiki.debian.org/ArmEabiPort) that are mixed-endian between integers and floating point numbers.

PR #285, opened 2018-06-05
KotOR dialog improvements

seedhartha:
Speakers in KotOR I/II dialogs now face the player character and play generic talk animations.


DrMcCoy:
Merged as 6b6954f8...84485718, thanks! :)

PR #286, opened 2018-06-05
Collision detection update for KotOR games

seedhartha:
Got rid of WalkmeshElevationEvaluator, moved Walkmesh to Aurora namespace, split Engines::KotOR::Walkmesh into Aurora::BWMFile and Graphics::Aurora::RenderableWalkmesh. Also implemented walkmeshes for situated objects in KotOR games and fixed some bugs.


DrMcCoy:
I don't like your class design for the Walkmesh, sorry.

That split between WalkmeshData and Walkmesh especially is just needlessly complicated and opaque. Pulling a WalkmeshData from one class into another, why? (Also, the methods in BMWFile should have been static, because they're not operating on any state. BMWFile is nothing more than a utility class here.)

What I'd propose is this:

`class Graphics::Aurora::Walkmesh : public Graphics::Renderable`, a common walkmesh structure. Holds both the walkmesh data and can render said data. Also houses methods `getElevationAt()` and `testCollision`. (Hopefully) useable by all the engines as a common base for a walkmesh.

`class Engines::KotOR::BWMFile : public Graphics::Aurora::Walkmesh` or `class Engines::KotOR::Walkmesh : public Graphics::Aurora::Walkmesh`, the walkmesh for KotOR/KotOR2 (*). As methods for loading a BWMFile into the internal Walkmesh structures, as well as appending while loading.

(*) Or are BWM files used by other engines as well? In that case, I'd put it into `Graphics::Aurora`.

seedhartha:
Changed the structure as you proposed.

DrMcCoy:
Merged, thanks! :)

PR #289, opened 2018-06-10
ENGINES: Fix segfault on module change in KotOR games

seedhartha:
When flycam mode is on.


DrMcCoy:
Merged as 6dfa199f, thanks! :)

PR #290, opened 2018-06-10
GRAPHICS: Remove option to disable animation thread

seedhartha:
Animation thread seems to be stable enough. KotOR isn't really playable w/o it and I don't see it having a negative effect on other games. Therefore there isn't really any reason to have that option anymore.


DrMcCoy:
Merged as 55794e9d, thanks! :)

PR #291, opened 2018-06-10
GRAPHICS: Fix droids and situated objects animations in KotOR games

seedhartha:
All KotOR models use relative position controllers now. Also ported bezier position controller handling from newer version of MDLops.


DrMcCoy:
Merged as dfdddec5, thanks! :)

PR #292, opened 2018-06-10
ENGINES: Fix appearance loading in KotOR games

seedhartha:


DrMcCoy:
Hardcoding these special cases feels a bit hacky. Are you sure it's not us interpreting the data files wrong?

seedhartha:
I changed the way appearance is loaded in KotOR games. It works OK but introduces a side effect of party members being naked. This will be fixed when equipment is implemented.

DrMcCoy:
Merged as 1d8fe48f, thanks! :)

PR #293, opened 2018-06-11
ENGINES: Make PC look at speaker in KotOR games

seedhartha:


DrMcCoy:
Merged as 84d1fae7...7b50deaf, thanks! :)

PR #295, opened 2018-06-12
KotOR animations fix

seedhartha:
This PR fixes various animation bugs in KotOR 1/2 and adds support for multiple animations per model.

### DIFF
--- a/src/graphics/aurora/animation.cpp
+++ b/src/graphics/aurora/animation.cpp
@@ -255,22 +255,8 @@ void Animation::updateSkinnedModel(Model *model) {
 		if (!node->_mesh || !node->_mesh->skin)
 			continue;
 
-		glm::mat4 transform;
-
-		transform = glm::translate(transform, glm::vec3(node->_positionBuffer[0],
-		                                                node->_positionBuffer[1],
-		                                                node->_positionBuffer[2]));
-
-		if (node->_orientationBuffer[0] != 0 ||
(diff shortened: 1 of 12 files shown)


src/graphics/aurora/animation.cpp
```diff
@@ -255,6 +255,10 @@ void Animation::updateSkinnedModel(Model *model) {
 		if (!node->_mesh || !node->_mesh->skin)
 			continue;
 
+		// Special case: handmaiden tongue, e.g., in 262tel
+		if (node->_parent && node->_parent->_name.stricmp("f_jaw_g") == 0)
+			continue;
```
DrMcCoy: Hmm, is there a way to get this to work without hardcoding a fix like that? Are there special properties in the models somewhere, maybe?

seedhartha: This particular model node is different from all the others in that it's parent is a bone. Because of this it has transformations applied twice to it: by the renderer and by the skeletal animation routine. I was not able to come up with a better solution atm.

DrMcCoy: Okay. Might be something to look out for in the future, then, maybe.

Please add your explanation as a comment there, so that we won't forget what this is about.

seedhartha:
Done.

DrMcCoy:
Merged, thanks! :)

PR #296, opened 2018-06-12
JADE: Remove dependency on KotOR GUI

seedhartha:
Make Jade Empire use a copy of KotOR GUI, effectively removing the dependency on KotOR engine. This is the first step of the GUI system update I have in mind.

### DIFF
--- a/src/engines/jade/gui/gui.cpp
+++ b/src/engines/jade/gui/gui.cpp
@@ -1,0 +1,309 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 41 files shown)


Nostritius:
Just for clarification, where is the benefit in making an extra copy of the gui system for jade?

seedhartha:
IMHO, KotOR and Jade are two separate contexts with different requirements and corner cases. This approach is also more in line with #261.

Nostritius:
Then why not simply seperating the gui system from kotor and make it an external dependency, like for example the script engines? Because, from what i know about the jade gui system, it's pretty much the same as in kotor.

DrMcCoy:
Is there a way we can find a common base the GUIs in the KotORs and Jade? No idea where that best fit, but maybe into src/engines/aurora/?

Likewise, I'd really like some common base for most of the two KotORs. Maybe through some kind of "virtual" engine?

seedhartha:
Sure, it makes sense to extract the common base for KotOR games and Jade Empire. As I said, it is only the first step.

DrMcCoy:
Merged as 8cf96010, thanks! :)

PR #297, opened 2018-06-14
ENGINES: Refactor KotOR GUI classes

seedhartha:
This PR moves KotOR widgets to Aurora namespace, fixes bad class design I introduced and does small improvements here and there. Most changes were made to the list box interface:
1. Scrollbar and item widgets were made it's children.
2. Renderable elements specific to scrollbar and inventory items were moved to the corresponding subclasses.
3. WidgetListItem was split into WidgetProtoItem and KotORInventoryItem.


DrMcCoy:
AddressSanitizer shows several leaks with your PR: https://gist.github.com/DrMcCoy/7ee56b9049e2cd2fca35a39b0f482ebe (some of these seem to be duplicates)

seedhartha:
Is that reliable? Some of these are objects passed to the ScopedPtr. They do get destroyed and that code I did not change. Scrollbar is not a leak also since it is later added to the GUI and will be destroyed from one.

seedhartha:
Could you check if the last commit solves those issues?

DrMcCoy:
It's even worse, now it's a use-after-free: https://gist.github.com/DrMcCoy/310bd730569d05411e4e11c630ee0af0

If you have a recent gcc (7+) or clang (5+), you can enable the sanitizers yourself with
```
export CXXFLAGS="-O2 -fno-omit-frame-pointer -fsanitize=address,undefined"
```
before calling ./configure (or cmake). `address` is ASan, the [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html), for many memory-related issues, and `undefined` is UBSan, the [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), for catching Undefined Behaviour. `-fno-omit-frame-pointer` makes sure that the stack frame pointer is set for every function, otherwise the reported stack traces will be meaningless to you.

For even more meaningful strack traces, you can also set
```
export UBSAN_OPTIONS=print_stacktrace=1
export ASAN_OPTIONS=fast_unwind_on_malloc=0
```
before running xoreos. This will make execution slightly slower, though.

Alternatively, you can run xoreos in [valgrind](http://valgrind.org/). This will be *very* slow, though. Also, you can't use a xoreos build with enabled sanitizers for that. They don't work together, unfortunately. However, valgrind's memcheck tool (its default tool) tends to catch certain cases the sanitizers won't.

seedhartha:
Apparently, both AddressSanitizer and valgrind are non-existent on Windows. I did however run xoreos in Dr. Memory and it detected no memory leaks with this PR. Scrollbar does in fact get freed in GUI::~GUI. Could be that I'm missing some steps?

DrMcCoy:
Oh, right, yes, valgrind doesn't do Windows, sorry. I had thought, though, that the sanitizers were available there, but apparently not, then? Unfortunate.

Frankly, I never understood how some people think Windows is so much better to develop on, because I found it very developer-hostile. But YMMV :).

I'll see if I can trace the leak, if it's still there with f6769acf6240f38cdc48af82b83f0312eb715241.

DrMcCoy:
I found the problem. It's in this line: https://github.com/xoreos/xoreos/pull/297/commits/f6769acf6240f38cdc48af82b83f0312eb715241#diff-09d8073cac1cd4f4e32183a34ff65104R318

You're using the just the tag in the GUI file, which is, for all scrollbars, just "SCROLLBAR". `addSub(*_scrollbar);` adds the widget to the GUI, but only if it's not already there. And it checks by looking for the tag. (Because we're sometimes settings widgets as a kind of "virtual" sub widget. They're instantiated by the GUI itself, but belong to another widget.)

If a widget with the same tag is already in the GUI, the widget is not added to the GUI, so will be leaked when the GUI deletes all its widgets in the destructor.

I.e. **tags in a GUI need to be unique**. The usual `GUI::addWidget()` method even throws when it finds a widget with the same tag.

This also necessary when checking if a widget was clicked or moused-over. `GUI::getWidgetAt()` gets a generic Renderable from the `GraphicsManager`, and uses the tag to find the correct widget this Renderable belongs to. This could, possibly, we changed by introducing some other kind of ID, in the future.

I generally use parent tag + "#" + sub tag for this. I.e. changing this line above to
```
_scrollbar = new WidgetScrollbar(*_gui, _tag + "#" + gff.getString("TAG"));
```
would fix the leak.

DrMcCoy:
And yes, this obviously needs to be better documented in the code :)

seedhartha:
Fixed the leak as you proposed and made Jade Empire use shared widget classes.

DrMcCoy:
Merged, thanks! :)

PR #298, opened 2018-06-16
GRAPHICS: Fix MSVC specific crashes

seedhartha:


DrMcCoy:
Oh, damn, good catch! Merged, thanks! :)

DrMcCoy:
(Technically, it's of course not just MSVC-specific, but just using reserve() and then accessing the storage is clearly a violation of the interface.)

PR #300, opened 2018-06-19
AURORA: GFF3Writer

Nostritius:
I thought long about my first attempt on a gff3 writer, but ultimately decided that it would be better to write an external class for exporting gff3, since the internal structure of the GFF3File class makes it very difficult to modify values. Also it is difficult to get the necessary information for exporting in an easy way. This class is only for generating a gff3 tree and export it. I have made many unit tests and also tested it against the original kotor executable by replacing a simple file by one created by the writer, and it behaved as i expected.

### DIFF
--- a/src/aurora/gff3writer.cpp
+++ b/src/aurora/gff3writer.cpp
@@ -1,0 +1,466 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  Writer for writing version 3.x of biowares GFF (general file format).
+ */
+
+#include <algorithm>
+
+#include <boost/make_shared.hpp>
+
+#include "src/common/writestream.h"
+#include "src/common/memreadstream.h"
+
+#include "src/aurora/gff3writer.h"
+
+namespace Aurora {
+
+GFF3Writer::GFF3Writer(uint32 id, uint32 version) : _id(id), _version(version) {
+	_structs.push_back(boost::make_shared<GFF3WriterStruct>(this));
+}
+
+GFF3WriterStructPtr GFF3Writer::getTopLevelStruct() {
+	return _structs[0];
+}
+
+void GFF3Writer::write(Common::WriteStream &stream) {
+	stream.writeUint32BE(_id);
+	stream.writeUint32BE(_version);
+
+	uint32 structOffset = 56; // id + version + header.
+	uint32 structCount = static_cast<uint32>(_structs.size());
+
+	uint32 fieldOffset = structOffset + structCount * 12;
+	uint32 fieldCount = static_cast<uint32>(_fields.size());
+
+	uint32 labelOffset = fieldOffset + fieldCount * 12;
+	uint32 labelCount = static_cast<uint32>(_labels.size());
+
+	uint32 fieldDataOffset = labelOffset + labelCount * 16;
+	uint32 fieldDataCount = 0;
+
+	// Count the total size of field data.
+	for (size_t i = 0; i < _fields.size(); ++i)
+		fieldDataCount += getFieldDataSize(_fields[i]);
+
+	uint32 fieldIndicesOffset = fieldDataOffset + fieldDataCount;
+	uint32 fieldIndicesCount = 0;
+
+	// Count all fields of structs with more than one field.
+	for (size_t i = 0; i < _structs.size(); ++i) {
+		const GFF3WriterStructPtr strct = _structs[i];
+		if (strct->getFieldCount() <= 1)
+			continue;
+
+		fieldIndicesCount += strct->getFieldCount() * 4;
+	}
+
+	uint32 listIndicesOffset = fieldIndicesOffset + fieldIndicesCount;
+	uint32 listIndicesCount = 0;
+
+	// Count all lists elements plus their size as int.
+	for (size_t i = 0; i < _lists.size(); ++i) {
+		listIndicesCount += (_lists[i]->getSize() + 1) * 4;
+	}
+
+	// Write the header.
+	stream.writeUint32LE(structOffset);
+	stream.writeUint32LE(structCount);
+	stream.writeUint32LE(fieldOffset);
+	stream.writeUint32LE(fieldCount);
+	stream.writeUint32LE(labelOffset);
+	stream.writeUint32LE(labelCount);
+	stream.writeUint32LE(fieldDataOffset);
+	stream.writeUint32LE(fieldDataCount);
+	stream.writeUint32LE(fieldIndicesOffset);
+	stream.writeUint32LE(fieldIndicesCount);
+	stream.writeUint32LE(listIndicesOffset);
+	stream.writeUint32LE(listIndicesCount);
+
+	// Write structs data.
+	size_t structFieldIndicesIndex = 0;
+	for (size_t i = 0; i < _structs.size(); ++i) {
+		GFF3WriterStructPtr strct = _structs[i];
+
+		// Struct Id.
+		stream.writeUint32LE(strct->getId());
+
+		// Field Index.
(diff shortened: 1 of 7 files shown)


DrMcCoy:
https://gist.github.com/DrMcCoy/43b2c2fbed18779be7ac2e631c2eacb4

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,610 @@
...
+				case GFF3Struct::kFieldTypeFloat:
+					stream.writeIEEEFloatLE(field.floatValue);
+					break;
+				default:
+					throw Common::Exception("Invalid Field type");
+			}
+		} else {
+			// If the values are complex (greater then 4 bytes) write the index to the field data.
+			stream.writeUint32LE(fieldDataIndex);
+			switch (field.type) {
```
DrMcCoy: Isn't that a repeat of lines 61ff?

Move that into a method `getFieldSize()` or something like that.

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,610 @@
...
+					break;
+				default:
+					throw Common::Exception("Invalid field type");
+			}
+		}
+	}
+
+	// Write labels.
+	for (size_t i = 0; i < _labels.size(); ++i) {
+		Common::UString label = _labels[i];
```
DrMcCoy: Is that copy there necessary?

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,610 @@
...
+
+#include "src/common/writestream.h"
+#include "src/common/memreadstream.h"
+
+#include "src/aurora/gff3writer.h"
+
+namespace Aurora {
+
+GFF3Writer::GFF3Writer(uint32 id, uint32 version) : _id(id), _version(version) {
+	_structs.push_back(GFF3WriterStructPtr(new GFF3WriterStruct(this)));
```
DrMcCoy: Can you use boost::make_shared instead of `new`'ing the struct into the shared_ptr (here and everywhere else as well)?

I.e. `boost::make_shared<GFF3WriterStruct>(this)` should be equivalent to `boost::shared_ptr<GFF3WriterStruct>(new GFF3WriterStruct(this))`, and gets rid of the explicit `new`.

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,610 @@
...
+	for (size_t i = 0; i < _fields.size(); ++i) {
+		const Field &field = _fields[i];
+		stream.writeUint32LE(field.type);
+		stream.writeUint32LE(field.labelIndex);
+
+		/*
+		 * Determine if this field has simple values (less equal 32 bit) which are written in the field
+		 * or complex values, bigger than 32bit like strings written in the field data section.
+		 */
+		bool simple =
```
DrMcCoy: That's a const.

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,610 @@
...
+			case GFF3Struct::kFieldTypeDouble:
+				stream.writeIEEEDoubleLE(field.doubleValue);
+				break;
+			case GFF3Struct::kFieldTypeStrRef:
+				stream.writeUint32LE(4);
+				stream.writeUint32LE(field.uint32Value);
+				break;
+			case GFF3Struct::kFieldTypeResRef:
+				stream.writeByte(MIN(static_cast<byte>(16), static_cast<byte>(field.stringValue.size())));
+				stream.write(field.stringValue.c_str(), field.stringValue.size());
```
DrMcCoy: You still might write a longer string than 16 bytes here

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,610 @@
...
+void GFF3WriterStruct::addByte(const Common::UString &label, byte value) {
+	// Create a field index.
+	_fieldIndices.push_back(_parent->_fields.size());
+
+	// Create field.
+	GFF3Writer::Field field;
+	field.type = GFF3Struct::kFieldTypeByte;
+	field.labelIndex = _parent->addLabel(label);
+	field.uint32Value = value;
+	_parent->_fields.push_back(field);
```
DrMcCoy: Lots of code duplication here in these methods. Creating the field index and adding the field, i.e. everything except setting the actual value, can be abstracted into a method.

Nostritius:

src/aurora/locstring.h
```diff
@@ -91,9 +91,9 @@ class LocString {
 	void readLocString(Common::SeekableReadStream &stream);
 
 	/** Get the size, the string table will consume after being written. */
-	uint32 getWrittenSize(bool withNullTerminate = false);
+	uint32 getWrittenSize(bool withNullTerminate = false) const;
 	/** Write the LocString to a write stream. */
-	void writeLocString(Common::WriteStream &stream, bool withNullTerminate = false);
+	void writeLocString(Common::WriteStream &stream, bool withNullTerminate = false) const ;
```
DrMcCoy: There's now an extra space between the `const` and the `;`

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+	// Add the label.
+	field.labelIndex = _parent->addLabel(label);
+
+	// Insert the newly created struct into the struct vector.
+	_parent->_structs.push_back(strct);
+
+	// Insert the struct to the field vector.
+	_parent->_fields.push_back(field);
+
```
DrMcCoy: Trailing whitespace

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,491 @@
...
+	// Insert the list to the field vector.
+	_parent->_fields.push_back(field);
+
+	return strct;
+}
+
+void GFF3WriterStruct::addByte(const Common::UString &label, byte value) {
+	size_t index = _parent->createField(GFF3Struct::kFieldTypeByte, label);
+	_parent->_fields[index].uint32Value = value;
+	_fieldIndices.push_back(index);
```
DrMcCoy: I'd go even further. `GFF3Writer::Field &GFF3WriterStruct::createField(GFF3Struct::FieldType type, const Common::UString &label)`. Calls `_parent->createField()`, put the index into `_fieldIndices` and returns a direct reference to the field.

Then you only need to do
```
createField(GFF3Struct::kFieldTypeByte, label).uint32Value = value;
```
etc. in each method.

src/aurora/gff3writer.h
```diff
@@ -0,0 +1,224 @@
...
+
+class GFF3WriterStruct;
+class GFF3WriterList;
+
+typedef boost::shared_ptr<GFF3WriterStruct> GFF3WriterStructPtr;
+typedef boost::shared_ptr<GFF3WriterList> GFF3WriterListPtr;
+
+class GFF3Writer {
+public:
+	//TODO: Add a constructor consuming a GFF3File object.
```
DrMcCoy: `// TODO: `

src/aurora/gff3writer.h
```diff
@@ -0,0 +1,224 @@
...
+	GFF3Writer(uint32 id, uint32 version = MKTAG('V', '3', '.', '2'));
+
+	/** Get the toplevel struct. */
+	GFF3WriterStructPtr getTopLevelStruct();
+
+	/** Write the gff3 to stream. */
```
DrMcCoy: Please, capitalize GFF and GFF3.

src/aurora/gff3writer.h
```diff
@@ -0,0 +1,224 @@
...
+			uint64Value = field.uint64Value;
+			int32Value = field.int32Value;
+			int64Value = field.int64Value;
+			floatValue = field.floatValue;
+			doubleValue = field.doubleValue;
+			vectorValue = field.vectorValue;
+			stringValue = field.stringValue;
+			locStringValue = field.locStringValue;
+
+			if (field.voidData.get() != 0) {
```
DrMcCoy: You should still do an unconditional `field.voidData.reset()` before that. Otherwise, "a = b", with a.voidData filled and b.voidData empty will not empty a.voidData.

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,491 @@
...
+			else
+				stream.writeUint32LE(0);
+		}
+
+		// Field Count.
+		stream.writeUint32LE(strct->getFieldCount());
+	}
+
+	// Write fields.
+	unsigned int fieldDataIndex = 0;
```
DrMcCoy: Why the move to unsigned int instead of size_t here?

src/aurora/gff3writer.cpp
```diff
@@ -0,0 +1,491 @@
...
+	std::vector<Common::UString>::iterator iter = std::find(_labels.begin(), _labels.end(), label);
+	if (iter != _labels.end()) {
+		return static_cast<uint32>(std::distance(_labels.begin(), iter));
+	} else {
+		_labels.push_back(label);
+		return static_cast<uint32>(_labels.size() - 1);
+	}
+}
+
+uint32 GFF3Writer::getFieldDataSize(const Aurora::GFF3Writer::Field &field) {
```
DrMcCoy: That can even be a static function in this file's scope. It doesn't need to be a method of GFF3Writer.

Nostritius:

DrMcCoy:
Merged as 4971a265a...da35c58f5, thanks! :)

PR #301, opened 2018-06-20
AURORA: ERFWriter

Nostritius:
This PR implements a writer for the erf archive format version 1.0.


DrMcCoy:
First of all, it's unclear whether this ERFWriter class takes over the streams of the files it adds. You should probably document that. If so, the ERFWriter needs to deallocate them. Otherwise, the caller (like the tests) need to. Right now, neither does, so the ERFWriter is leaking all of its files.

Second, why are you reading the stream data manually into an array instead of using WriteStream::writeStream()?

Third, I'm not so sure about the design of the class itself. It hold all the file data in memory before writing it into a stream? That's not really optimal when the ERF is rather big (which might be the case if we're using the ERFWriter in the tools, and also for NWN with large modules).

Instead, an ideal ERFWriter class should write the file data as soon as it gets them. Since the index comes before the file data, that might be a bit difficult right now, yes.

There's two possible solution in my mind right now:
1) The caller needs to first register all the file meta data (name, size, type) with the ERFWriter instance, before ERFWriter accepts the file data in turn
2) WriteStream will be expanded to allow seeking within the writing stream (possibly even splitting it into a SeekableWriteStream like the ReadStream classes). Then, the ERFWriter (in the constructor, for example) is given the number of files. It can allocate the index in the output stream, by making it all 0x00. Then, for each file, it seeks to the index, writes the metadata and then seeks to the bottom to write the file data.

Number 2 is the more difficult, but more cleaner, in my mind, approach. Should be easy for memory streams. For file streams, I'm unsure about the portability aspects of it (*), but I'd wager it works correctly on all systems we'd ever care about.

(*) Technically, the way we seek in binary streams is already not portable. Officially, seek positions in binary streams are not necessarily equal to the byte offsets. It's only required that you can get a position with ftell() and seek back to the same position with fseek(). That's it. However, but Linux/macOS/Windows behave like we want. In fact, I'm not aware of any system where this doesn't work.

Nostritius:
About the third thing, would temporary files be a solution, so that the ERFWriter writes the streams to a tmp dir and then simply writes one after another from file into the erf and deletes them?

DrMcCoy:
Not really, no

Nostritius:
With the new SeekableWriteStreams, i rewrote the erf writer. It now takes the write stream plus the total number of files and reserves space for the table entries. I then write these entries one after another with the add() method.

src/aurora/erfwriter.cpp
```diff
@@ -0,0 +1,108 @@
...
+void ERFWriter::add(const Common::UString resRef, uint16 resType, unsigned int size, Common::ReadStream &stream) {
+	if (_currentFileCount == _fileCount)
+		throw Common::Exception("More files added than expected");
+
+	// Write the key table entry
+	_stream.seek(_keyTableOffset + _currentFileCount * 24);
+
+	char cname[16];
+	memset(cname, '\0', 16);
+	memcpy(cname, resRef.c_str(), MIN(static_cast<size_t>(16), resRef.size()));
```
DrMcCoy: `MIN<size_t>()`.

Is the copying necessary, though? You could just write `MIN<size_t>(16, resRef.size())` from `resRef.c_str()` and then `writeZeros(16 - MIN<size_t>(16, resRef.size())`.

src/aurora/erfwriter.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+	// Write the resource table entry
+	_stream.seek(_resourceTableOffset + _currentFileCount * 8);
+
+	_stream.writeUint32LE(_offsetToResourceData);
+	_stream.writeUint32LE(size);
+
+	// Write the actual resource data and add the size to the offset
+	_stream.seek(_offsetToResourceData);
+	_stream.writeStream(stream, size);
```
DrMcCoy: You need to check the return value of writeStream().

`WriteStream::writeStream()` will read at most size bytes from the stream, but might read less, if the stream is finished beforehand. It returns the number of bytes actually copied.

You could even put the size argument last in this method, and default it to SIZE_MAX, meaning the whole stream, however much that would be. This way, the caller doesn't need to necessarily know how big the stream is beforehand.

(And you need to move writing the resource table entry after writing the file data.)

src/aurora/erfwriter.cpp
```diff
@@ -0,0 +1,108 @@
...
+	if (_description.getNumStrings())
+		stream.writeUint32LE(_description.getID());
+	else
+		stream.writeUint32LE(0);
+
+	// Write 116 bytes of reserved header data.
+	stream.writeZeros(116);
+
+	// Write the Localized string table.
+	_description.writeLocString(stream);
```
DrMcCoy: Can you add a test for writing a LocString with multiple languages, please?

In fact, you should probably add a test for writing LocStrings to the LocStrings unit tests as well.

src/aurora/erfwriter.cpp
```diff
@@ -0,0 +1,108 @@
...
+	stream.writeUint32BE(_id);
+	stream.writeUint32BE(MKTAG('V', '1', '.', '0'));
+
+	// Write Header.
+	stream.writeUint32LE(_description.getNumStrings()); // Language count.
+	stream.writeUint32LE(_description.getWrittenSize()); // Localized string size.
+
+	stream.writeUint32LE(_fileCount); // Entry Count.
+
+	uint32 locStringTableOffset = 160; // 160 is the size of the erf header.
```
DrMcCoy: That's even a static const right there. `static const kLocStringTableOffset = 160;`

src/aurora/erfwriter.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+/** @file
+ *  Writing BioWare's ERFs (encapsulated resource file).
+ */
+
+#ifndef AURORA_ERFWRITER_H
+#define AURORA_ERFWRITER_H
+
+#include "src/common/scopedptr.h"
```
DrMcCoy: This include is not used.

src/aurora/erfwriter.h
```diff
@@ -0,0 +1,70 @@
...
+private:
+	Common::SeekableWriteStream &_stream;
+
+	uint32 _id;
+	uint32 _currentFileCount;
+	uint32 _fileCount;
+	uint32 _offsetToResourceData;
+	uint32 _keyTableOffset;
+	uint32 _resourceTableOffset;
+	LocString _description;
```
DrMcCoy: Do you need to save the description? It's only used in the constructor, no?

Nostritius:

DrMcCoy:
https://gist.github.com/DrMcCoy/4471284ecc2f29f411c2fbf8fcce4679
https://gist.github.com/DrMcCoy/d5c27fa2f780502f0457692695bea548

src/aurora/erfwriter.cpp
```diff
@@ -0,0 +1,113 @@
...
+ *  Writing BioWare's ERFs (encapsulated resource file).
+ */
+
+#include <ctime>
+
+#include "src/aurora/erfwriter.h"
+
+namespace Aurora {
+
+static const uint32 kLocStringTableOffset = 160; // 160 is the size of the erf header.
```
DrMcCoy: You didn't need to make it a global. It should still be in the scope of the constructor, defined right before it's used.

Nostritius:
The issues should now be fixed.

DrMcCoy:
Merged as f4543bfd...d6ca9544, thanks! :)

PR #302, opened 2018-06-20
COMMON: Seekable write streams

Nostritius:
This PR implements seeking in Write streams as suggested in #301.

### DIFF
--- a/src/common/memwritestream.cpp
+++ b/src/common/memwritestream.cpp
@@ -78,9 +78,23 @@ size_t MemoryWriteStream::size() const {
 	return _bufSize;
 }
 
+size_t MemoryWriteStream::seek(ptrdiff_t offset, SeekableWriteStream::Origin whence) {
+	assert((size_t)_pos <= _bufSize);
+
+	const size_t oldPos = _pos;
+	const size_t newPos = evalSeek(offset, whence, _pos, 0, size());
+	if (newPos > _bufSize)
+		throw Exception(kSeekError);
+
+	_pos = newPos;
+	_ptr = _ptr - oldPos + newPos;
+
+	return oldPos;
+}
+
 
 MemoryWriteStreamDynamic::MemoryWriteStreamDynamic(bool disposeMemory, size_t capacity) :
-	_data(0, disposeMemory), _ptr(0), _capacity(0), _size(0) {
+		_data(0, disposeMemory), _ptr(0), _capacity(0), _size(0) {
 
 	reserve(capacity);
 }
@@ -99,9 +113,10 @@ void MemoryWriteStreamDynamic::reserve(size_t s) {
 	if (_data)
 		memcpy(newData, _data.get(), _size);
 
+	size_t oldPos = pos();
 	_data.dispose();
 	_data.reset(newData);
-	_ptr = _data.get() + _size;
+	_ptr = _data.get() + oldPos;
 }
 
 void MemoryWriteStreamDynamic::ensureCapacity(size_t newLen) {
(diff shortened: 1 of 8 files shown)


src/common/memwritestream.cpp
```diff
@@ -114,12 +128,14 @@ void MemoryWriteStreamDynamic::ensureCapacity(size_t newLen) {
 size_t MemoryWriteStreamDynamic::write(const void *dataPtr, size_t dataSize) {
 	assert(dataPtr);
 
-	ensureCapacity(_size + dataSize);
+	ensureCapacity(pos() + dataSize);
 
 	std::memcpy(_ptr, dataPtr, dataSize);
 
+	if (pos() + dataSize > _size)
+		_size += dataSize;
```
DrMcCoy: Hmm, are you sure that's correct? Unless I'm not seeing something, shouldn't that be `_size += dataSize - pos();`? I.e. `_size` should grow by the difference/overhang.

If you could also add a unit test for that case then, that would be great. Something like write 6 bytes, make sure that the size of the stream is 6, seek back two bytes, write another 6 bytes and make sure that the size of the stream is then 10 (instead of 12), and that the contents is also still correct.

And the same test for the file writer, too.

Likewise, please add a test to the non-dynamic write stream that wants to write more. Something like I said above, write 6 bytes, seek back 2, write another 6 and make sure that the method returns 2 (for 2 bytes written, because there's only place in the buffer for 2). And content matching.

src/common/writefile.cpp
```diff
@@ -94,8 +94,9 @@ size_t WriteFile::write(const void *dataPtr, size_t dataSize) {
 
 	assert(dataPtr);
 
+	const size_t oldPos = pos();
 	const size_t written = std::fwrite(dataPtr, 1, dataSize, _handle);
-	_size += written;
+	_size += MAX(static_cast<size_t>(0), -size() + oldPos + written);
```
DrMcCoy: You could use `MAX<size_t>()` instead, if you want to specifiy that it should use the `size_t` version of the MAX template.

Nostritius:
That was indeed a problem. I fixed this and added more unit tests, to check this.

DrMcCoy:
I'd really feel better if these new tests, with the overhang, also checked that the expected contents was written (that's why I explictly said "and that the contents is also still correct" and "And content matching", after all).

It would be bad if, for example, the truncating case for the non-dynamic MemoryWriteStream accidentally wrote the last two bytes instead of the first two, or something.

Nostritius:
You're right, i forgot this. I have now added a content check to the unit tests.

tests/common/memwritestream.cpp
```diff
@@ -311,6 +311,70 @@ GTEST_TEST(MemoryWriteStream, writeString) {
...
+	EXPECT_EQ(data[4], 0x89);
+	EXPECT_EQ(data[5], 0xAB);
+	EXPECT_EQ(data[6], 0x01);
+	EXPECT_EQ(data[7], 0x23);
+	EXPECT_EQ(data[8], 0x45);
+	EXPECT_EQ(data[9], 0x67);
+	EXPECT_EQ(data[10], 0x89);
+	EXPECT_EQ(data[11], 0xAB);
+	EXPECT_EQ(data[12], 0xCD);
+	EXPECT_EQ(data[13], 0xEF);
```
DrMcCoy: I'm sorry, but I'm still not completely happy with this test.

It now doesn't test the case where the non-dynamic MemoryWriteStream has only space for less than the requested amount.

I.e. add another writing after this, in this test here:
```
static const byte kData[4] = { 0x12, 0x34, 0x56, 0x78 };
stream.seek(-2, Common::SeekableWriteStream::kOriginCurrent);
const size_t writeCount = stream.write(kData, 4);
EXPECT_EQ(writeCount, 2);
EXPECT_EQ(data[0], 0x01);
[...]
```

I.e. seek back, ask to write more than there's space for it, make sure only x bytes are written and compare to make sure the correct bytes are written.

Nostritius:
I added the test for the none-dynamic write stream.

DrMcCoy:
Thanks, now all the weird corner-cases that someone might accidentally break are covered, I think.

Merged as 10d2910c...29e7f1ef, thanks! :)

I also added "const" to three variable definitions. I like variables to be as explictly typed as possible. If the variable is only meant to be read from, because it caches a calculation or something, I like the const to be explictly stated.

PR #303, opened 2018-06-22
COMMON: Writing multiple bytes at once

Nostritius:
An idea, ihad while writing #302, was to have more the possibility to write a sequence of same bytes to the stream. This is for example useful for the ERFWriter, which has a 116 byte zeros section.


Nostritius:
Done :)

DrMcCoy:
Merged as 29e7f1ef...f4543bfd, thanks! :)

PR #305, opened 2018-06-24
KOTOR: Fix equipment bug for non-caucasian characters

seedhartha:


DrMcCoy:
Ah. I was just about to say that I remember seeing it working. Didn't realize that it was skin color dependant. By chance, I didn't hit upon that before.

If you haven't already, could you go through all options in both games, to check that all of them work now? Thanks.

PR #306, opened 2018-06-25
COMMON: Fix bug write completely in written data in WriteFile

Nostritius:
I recognized a bug in the WriteFile class, where the size was calculated wrong, when writing completely in already written data. This PR fixes this problem and extends the unit tests to check for this corner case.


DrMcCoy:
The correct signed version of size_t is ptrdiff_t, not int (well, in POSIX, there's also ssize_t).

Merged as 8197d16e...584b9c07, with me slightly fixing your fix there, thanks! :)

PR #307, opened 2018-07-03
KOTOR: User data

Nostritius:
This PR aims to create a base for the user data, the data which the game generates, mostly save games.

### DIFF
--- a/src/engines/aurora/userdata.cpp
+++ b/src/engines/aurora/userdata.cpp
@@ -1,0 +1,47 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  Utility class for managing user data, for example save games.
+ */
+
+#include "src/common/filepath.h"
+#include "src/common/configman.h"
+
+#include "src/engines/aurora/userdata.h"
+
+namespace Engines {
+
+UserData::UserData(const Common::UString &id) {
+	// If a directory for the user data is given over command line, we use it.
+	if (ConfigMan.hasKey("userdata"))
+		_rootPath = ConfigMan.getString("userdata");
+	else
+		_rootPath = Common::FilePath::getUserDataDirectory() + "/saves/" + id;
+
+	// If the user data directory was not created yet, create it.
+	if (!Common::FilePath::isDirectory(_rootPath))
+		Common::FilePath::createDirectories(_rootPath);
+}
+
+UserData::~UserData() {
+}
+
+} // End of namespace Engines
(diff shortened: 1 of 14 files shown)


DrMcCoy:
Does the texture from stream thing takes into account:

1) that we may need to recreate all textures? When the resolution is changed, certain parameters are changed, the GL context needs to be recreated, including textures. Texture::doRebuild() takes care of that, by recreating the texture from the image instance
2) that textures can reload themselves? This is needed for reloading textures of a different quality when a different texture pack is loaded. Texture::reload() takes care of that, by reloading the texture when it knows its filename.

src/engines/aurora/userdata.cpp
```diff
@@ -0,0 +1,47 @@
...
+UserData::UserData(const Common::UString &id) {
+	// If a directory for the user data is given over command line, we use it.
+	if (ConfigMan.hasKey("userdata"))
+		_rootPath = ConfigMan.getString("userdata");
+	else
+		_rootPath = Common::FilePath::getUserDataDirectory() + "/saves/" + id;
+
+	// If the user data directory was not created yet, create it.
+	if (!Common::FilePath::isDirectory(_rootPath))
+		Common::FilePath::createDirectories(_rootPath);
```
DrMcCoy: `Common::FilePath::createDirectories()` might throw when the directories can't be created (because of, for example, missing permissions on the filesystem).

I don't think you want your whole UserData constructor to fail in that case, do you?

src/engines/kotor/userdata.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+		save = Common::FilePath::getFile(save);
+		Common::UString id = save.substr(save.begin(), save.getPosition(6));
+		unsigned int value = 0;
+		Common::parseString(id, value);
+		_maxSaveId = MAX(_maxSaveId, value);
+
+		if (!save.contains("AUTOSAVE") && !save.contains("QUICKSAVE")) {
+			id = save.substr(save.getPosition(13), save.end());
+			Common::parseString(id, value);
```
DrMcCoy: When the save name is malformed, `id` might be empty, in which case `Common::parseString()` throws. It might also also contain non-numerical data, in which case it also throws.

DrMcCoy: The glob *might* already take care of it, but it's not obvious. Some comments might help.

src/engines/aurora/userdata.cpp
Nostritius: What would you suggest as standard behaviour in such a case?

Nostritius:
I've gone over the texture thing and it seems, that it does not make any problem.

1. The rebuilding works fine, since the image decoder saves the raw data of the image and it will be simply applied as texture

2. The reloading works also, since the image has no name applied, and so won't be reloaded. Since there is only one version of the image, would this be enough or should this be reloaded anyways?

Nostritius:
Ah damned, I accidentally closed this PR while cleaning up my branches

DrMcCoy:
Hmm, okay, I completely lost track what this PR is supposed to solve, sorry.

You say that this is supposed to define a place for user data like screenshots and save games, but what are these quad and texture commits?

Right now, this would just create a directory `saves/$id` in the user data directory, i.e. `~/.local/share/xoreos/saves/foobar`. I don't think this is the right approach.

I image we might want to roll this the other way round, i.e. have a `~/.local/share/xoreos/$id` directory for each target (and the ID is the target name from the config/command line, I guess). And that directory then has subdirectories for saves, screenshots, etc.

So probably a method to get the whole encompassing game user data directory (`~/.local/share/xoreos/$id`), and a method to get a use-directory by enum, like kUserDirectorySaves would give you `~/.local/share/xoreos/$id/saves`.

We might even pull the target name directly from the ConfigManager instead of passing it as an argument.

What use cases do we need?

- Saves
- Screenshots
- Portraits
- Override (override directory for placing resources in, taking precedence over all other sources)

NWN / NWN2 has a few more

- Modules (community modules)
- HAKs (community resource hak-packs)
- Movies (community movie resources)
- Music (community music resources)
- Talktables (community tlk resources)
- LocalVault (local single-player characters)
- ServerVault (multi-player characters)
- Database (campaign database to transfer information from one module to the next)

More?

PR #308, opened 2018-07-07
KOTOR: Customize listbox sounds in saveload menu, dialogs and equipment screen

seedhartha:
All three have different combinations of item selection sounds.


DrMcCoy:
Merged, thanks! :)

PR #309, opened 2018-07-11
KOTOR: Make resolution changeable

Nostritius:
This PR adds the possibility to change a resolution in kotor and adds an enhancement for the list structure to enable mouse wheel scrolling.

### DIFF
--- a/src/engines/aurora/gui.cpp
+++ b/src/engines/aurora/gui.cpp
@@ -292,6 +292,21 @@ void GUI::removeWidget(Widget *widget) {
 	delete widget;
 }
 
+void GUI::clearWidgets() {
+	// Delete all widgets
+	for (WidgetList::iterator widget = _widgets.begin(); widget != _widgets.end(); ++widget) {
+		delete *widget;
+		*widget = 0;
+	}
+
+	_widgets.clear();
+	_widgetMap.clear();
+}
+
+bool GUI::empty() {
+	return _widgets.empty();
+}
+
 bool GUI::hasWidget(const Common::UString &tag) const {
 	return getWidget(tag) != 0;
 }
--- a/src/engines/aurora/kotorjadegui/listbox.cpp
+++ b/src/engines/aurora/kotorjadegui/listbox.cpp
@@ -382,13 +382,13 @@ void WidgetListBox::applyChangesToItemWidgets() {
 	}
 }
 
-void WidgetListBox::mouseWheel(uint8 state, int x, int y) {
+void WidgetListBox::mouseWheel(uint8 UNUSED(state), int UNUSED(x), int y) {
 	if (y == 0 || !_adjustHeight)
 		return;
 
 	_startIndex = MIN(
 		MAX(_startIndex - y, 0),
-		MIN(static_cast<int>(_itemWidgets.size()) - _numVisibleItems, 0)
+		MAX<int>(_itemWidgets.size() - _numVisibleItems, 0)
 	);
 	refreshItemWidgets();
 }
--- a/src/engines/aurora/kotorjadegui/listbox.cpp
+++ b/src/engines/aurora/kotorjadegui/listbox.cpp
@@ -142,7 +142,7 @@ void WidgetListBox::createItemWidgets(uint32 count) {
 				break;
 			case kLBItemTypeDefault:
 			default:
-				item = new WidgetProtoItem(*_gui, tag);
+				item = new WidgetProtoItem(*_gui, tag, this);
 				break;
 		}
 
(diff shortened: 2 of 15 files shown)


DrMcCoy:
1) The mouse scrolls the wrong way round, at least as I'm accustomed to it. When I move the scroll wheel down, the list *should* scroll down, not up.
2) Remove the different refresh rates. We can't enforce them with SDL, or in fact at all on many modern systems. Just have width x height, without refresh rates
3) Support for 1280x1024 was added with path 1.02. So in an unpatched game, it's not available. That's why `Engines::KotOR::HUD::HUD()` in src/engines/kotor/gui/ingame/hud.cpp checks to see if the relevant mipc* GUI files (and mi8x6 for the Xbox version) are available before adding each resolution to the set of supported resolutions. You should probably do that as well here. (In the future, when we have GUI scaling, we might relax that)
4) When switching resolutions, you need to reload the HUD and recalculate the conversation black-bar/cutoff coordinates. Maybe more. As is, switching resolution and then going in-game breaks the in-game GUI

DrMcCoy:
I see you force-pushed this, did you already fix what I suggested or are you still working on this? :)

Nostritius:
No, not yet. I think i fixed something smaller and wanted to save this.

src/engines/aurora/kotorjadegui/listbox.cpp
```diff
@@ -248,6 +248,13 @@ void WidgetListBox::selectItemByWidgetTag(const Common::UString &tag) {
 	}
 }
 
+void WidgetListBox::selectItemByIndex(int index) {
+	if (index < 0 && index >= _items.size())
```
DrMcCoy:
```
src/engines/aurora/kotorjadegui/listbox.cpp: In member function ‘void Engines::WidgetListBox::selectItemByIndex(int)’:
src/engines/aurora/kotorjadegui/listbox.cpp:252:25: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<Common::UString>::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
  if (index < 0 && index >= _items.size())
                   ~~~~~~^~~~~~~~~~~~~~~~
```

I.e. add an `(uint)` to the second `index`.

Also, that should be a `||`, not an `&&`.

src/engines/kotor/gui/options/resolution.cpp
```diff
@@ -22,23 +22,64 @@
...
 
-OptionsResolutionMenu::OptionsResolutionMenu(Console *console) : GUI(console) {
+OptionsResolutionMenu::OptionsResolutionMenu(Console *console) : GUI(console), _newWidth(0), _newHeight(0) {
 	load("optresolution");
 
 	WidgetPanel *guiPanel = getPanel("TGuiPanel");
 	guiPanel->setPosition(-guiPanel->getWidth()/2, -guiPanel->getHeight()/2, 0);
+
+	int currentIndex = -1;
+	std::vector<Graphics::DisplayMode> modes = WindowMan.getDisplayModes();
```
DrMcCoy: This only accounts for resolutions the display says it can natively do, so only, I guess, fullscreen modes the display advertises. This ignores wanting to run it in windowed mode in a non-native resolution.

So I'd add more resolutions to that list. For example, the resolutions the original KotOR binary supports. And just a list of general common resolutions. But you might want only add resolutions the native display could fit, i.e. get the max width and height of all display modes, and only add resolutions <= that max width and height.

src/engines/aurora/gui.cpp
```diff
@@ -292,6 +292,21 @@ void GUI::removeWidget(Widget *widget) {
 	delete widget;
 }
 
+void GUI::clearWidgets() {
```
DrMcCoy: Remove the period at the end of the commit message, please.

src/engines/kotor/gui/options/resolution.cpp
```diff
@@ -22,23 +22,64 @@
...
+	}
+
+	listBox->setAdjustHeight(true);
+	listBox->setItemSelectionEnabled(true);
+	listBox->refreshItemWidgets();
 }
 
 void OptionsResolutionMenu::callbackActive(Widget &widget) {
-	// TODO: Add changing of resolutions
+	if (widget.getTag() == "BTN_OK") {
```
DrMcCoy: You also need to update the "width" and "height" config variables here, so that a resolution change is saved in the config file and sticks with the next restart of xoreos.

Nostritius:
The issues should now be fixed and i added support for common resolutions if windowed.

farmboy0:
Is there a reason not using the static resolution list already in the code?

Nostritius:
Which static resolution list do you mean?

DrMcCoy:
Do you mean the static list of native resolutions the original game supported? Because we now theoretically support more resolutions than that, due to scaling (well, pulling apart the elements, not resizing the individual widgets themselves (yet?)).

farmboy0:
/src/graphics/resolution.h

DrMcCoy:
Oh, I forgot that we have those :)

Nostritius:
I have now changed the last commit to use the kResolution array.

DrMcCoy:
Merged as d57111291...46b9be158, thanks! :)

There's still one thing, though, the GUIBackground. For resolutions smaller than the biggest original resolution, it should probably get the next biggest (that can fit both width and height). If it's centered correctly, that just means something is cut off on the sides.

For resolutions greater than that, it should probably display the largest background, and add a black border behind it, so that the game is not seen behind it, I guess.

PR #311, opened 2018-07-14
KotOR actions update

seedhartha:
This PR:
1. Extracts action execution to a dedicated class.
2. Implements collision detection for non-PC creatures.
3. Adds support for action "follow leader".
4. Loads creature speed based on appearance.

### DIFF
--- a/src/engines/kotor/actionexecutor.cpp
+++ b/src/engines/kotor/actionexecutor.cpp
@@ -1,0 +1,111 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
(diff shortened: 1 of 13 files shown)


DrMcCoy:
With this PR, Trask's movement is now framerate dependant. His animation is "jittering"/stuttering  with vsync turned on, walking smooth when vsync is turned off.

776b98cca seems to introduce this.

seedhartha:
Fixed. There's also an issue with walk/run animation switch, though I believe it should be fixed with proper animation interpolation later.

DrMcCoy:
Thanks :)

I'll recheck this later in the evening today. I'm currently not at home on my desktop, but on my laptop, and vsync doesn't work on my laptop [*], so I can't actually check this here.

[*] Nvidia with Optimus on Linux, which funnels the rendered image from the Nvidia memory into the in-CPU Intel GPU memory (and only the Intel GPU has access to the monitor). There is a way to make vsync work for that, but I'd need to rewrite my X config, and so far I haven't had time for that.

Though I might try getting the framerate down differently? Hmm...

DrMcCoy:
Okay, checked with my desktop machine as well. And yes, the stuttering seems to be gone, nice, thanks! :)

Merged as e3c8c96de...169a42676

PR #312, opened 2018-07-14
KOTOR: HUD Scaling

Nostritius:
This PR adds the ability to scale the hud of no proper resolution is found.

### DIFF
--- a/src/engines/kotor/gui/ingame/hud.cpp
+++ b/src/engines/kotor/gui/ingame/hud.cpp
@@ -39,6 +39,134 @@ namespace Engines {
 
 namespace KotOR {
 
+enum Position {
+	kPositionUpperLeft,
+	kPositionUpperRight,
+	kPositionLowerLeft,
+	kPositionLowerRight,
+	kPositionUpperMid,
+	kPositionLowerMid,
+	kPositionNotSpecified
+};
+
+struct KnownWidget {
+	const char *name;
+	bool visible;
+	Position position;
+};
+
+static const KnownWidget kKnownWidgets[] = {
+	{ "LBL_MAPVIEW", true, kPositionUpperLeft },
+	{ "LBL_MAPBORDER", true, kPositionUpperLeft },
+	{ "LBL_ARROW", true, kPositionUpperLeft },
+	{ "LBL_PLOTXP", false, kPositionUpperLeft },
+	{ "LBL_STEALTHXP", false, kPositionUpperLeft },
+	{ "LBL_CASH", false, kPositionUpperLeft },
+	{ "LBL_JOURNAL", false, kPositionUpperLeft },
+	{ "LBL_LIGHTSHIFT", false, kPositionUpperLeft },
+	{ "LBL_DARKSHIFT", false, kPositionUpperLeft },
+	{ "LBL_ITEMRCVD", false, kPositionUpperLeft },
+	{ "LBL_ITEMLOST", false, kPositionUpperLeft },
+
+	{ "LBL_MENUBG", true, kPositionUpperRight },
+	{ "BTN_MSG", true, kPositionUpperRight },
+	{ "BTN_ABI", true, kPositionUpperRight },
+	{ "BTN_JOU", true, kPositionUpperRight },
+	{ "BTN_INV", true, kPositionUpperRight },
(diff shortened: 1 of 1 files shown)


src/engines/kotor/gui/ingame/hud.cpp
```diff
@@ -207,6 +213,138 @@ void HUD::setPartyMember2(Creature *creature) {
 	setPortrait(3, creature != 0, creature ? creature->getPortrait() : "");
 }
 
+void HUD::scaleWidgets(unsigned int wWidth, unsigned int wHeight) {
+	scaleWidgetToUpperLeftEdge(getWidget("LBL_MAPVIEW"), wWidth, wHeight);
```
DrMcCoy: Hmm, a lot of code duplication there. And duplicating the widget names partially.

Maybe instead this big chunk, we should rather have a static array of structs with the names of all widgets, their visibility status from the start (taking over the list on lines 106ff) and their relative position.

Something like

```
enum Position {
	kPositionUpperLeft,
	kPositionMidLeft,
	...
	kPositionUnknown
};

struct KnownWidget {
	const *name;
	bool visible;
	Position position
};

static const KnownWidget kKnownWidgets[] = {
	{ "LBL_MAPBORDER", true, kPositionUpperLeft },
	{ "LBL_LEVELUP1", false, kPositionUpperLeft },
	...
};
```

Then in `HUD::HUD()`, we're iterating over kKnownWidgets, setting them invisible if necessary, and scaling them if necessary and a position is given. Any other property we need there?

DrMcCoy:
That's better, yes. Works great, mostly.

One thing, though: the Xbox GUI is called "mi8x6", not "mipc28x6", so just taking kResolution[0] won't work. You need to take the first available one, not the first of all possible ones.

Nostritius:
The code should now look after mipc28x6 and mi8x6 with a preference to mipc28x6

src/engines/kotor/gui/ingame/hud.cpp
```diff
@@ -81,49 +209,60 @@ HUD::HUD(Module &module, Engines::Console *console)
...
 	if (!foundRes) {
-		Common::UString resString;
-
 		for (std::set<Resolution>::const_iterator it = availableRes.begin(); it != availableRes.end(); ++it) {
-			if (!resString.empty())
-				resString += ", ";
-
-			resString += Common::UString::format("%dx%d", it->width, it->height);
+			// Search for an existing interface with 800x600 resolution with preference to the pc interface.
+			if (Common::UString(it->gui) == "mipc28x6" || (Common::UString(it->gui) == "mi8x6" && !foundRes)) {
```
DrMcCoy: Why are you comparing the name instead width and height directly?

You can also just break when one is found, instead of checking that foundRes is not assigned yet.

I.e. I think that
```
if ((it->width == 800) && (it->height == 600)) {
	foundRes = &*it;
	break;
}
```

would be way clearer.

DrMcCoy: (In C++11, I'd even say that's a std::find() with a lambda checking width and height, but that's future talk :P)

Nostritius: Since the pc version of kotor also contains the gui file for the xbox ui, i wanted to prefer the pc ui and the only way of doing this, is by comparing their names. Maybe later you can add a command line switch for forcing the xbox ui. Also i wanted to make sure, that in the case of both being found and supported, the pc version is used since that would be what a normal pc user would expect first and, as far as i know, the pc ui is not contained in the xbox game but the xbox ui is contained in the pc game.

DrMcCoy: That's why availableRes is a std::set, operator<() only checks the dimensions, and the PC version is checked first. availableRes doesn't contain the xbox GUI file in PC version, even though the file is theoretically available.

DrMcCoy: (Yes, that's a bit of a hack in either case, I know)

Nostritius: you're right, i will correct this

DrMcCoy:
Merged, thanks! :)

Can you move that over to KotOR2, too, please?

PR #314, opened 2018-07-15
KOTOR: Implement party selection in party selection GUI

seedhartha:
Applying changes in-game are for later PR.

### DIFF
--- a/src/engines/aurora/kotorjadegui/checkbox.cpp
+++ b/src/engines/aurora/kotorjadegui/checkbox.cpp
@@ -100,16 +100,23 @@ void WidgetCheckBox::setState(bool state) {
 	_state = state;
 
 	// Preserve the current color after replacing the texture
+
 	float textR, textG, textB, textA;
-	float quadR, quadG, quadB, quadA;
 	_text->getColor(textR, textG, textB, textA);
-	_quad->getColor(quadR, quadG, quadB, quadA);
+
+	float quadR, quadG, quadB, quadA;
+	if (_quad)
+		_quad->getColor(quadR, quadG, quadB, quadA);
+
+	//-
 
 	setFill(_state ? _selected : _unselected);
 	setHighlight(_state ? _selectedHighlighted : _unselectedHighlighted);
 
 	_text->setColor(textR, textG, textB, textA);
-	_quad->setColor(quadR, quadG, quadB, quadA);
+
+	if (_quad)
+		_quad->setColor(quadR, quadG, quadB, quadA);
 
 	setActive(true);
 }
--- a/src/engines/aurora/kotorjadegui/checkbox.cpp
+++ b/src/engines/aurora/kotorjadegui/checkbox.cpp
@@ -172,6 +172,22 @@ void WidgetCheckBox::mouseUp(uint8 UNUSED(state), float UNUSED(x), float UNUSED(
 	setActive(true);
 }
 
+void WidgetCheckBox::setSelected(const Common::UString &selected) {
+	_selected = selected;
+}
+
+void WidgetCheckBox::setUnselected(const Common::UString &unselected) {
+	_unselected = unselected;
+}
+
+void WidgetCheckBox::setSelectedHighlighted(const Common::UString &selectedHighlighted) {
+	_selectedHighlighted = selectedHighlighted;
+}
+
+void WidgetCheckBox::setUnselectedHighlighted(const Common::UString &unselectedHighlighted) {
+	_unselectedHighlighted = unselectedHighlighted;
+}
+
 void WidgetCheckBox::setTextHighlighting(Graphics::Aurora::Highlightable *highlightable) {
 	highlightable->setHighlightable(true);
 	highlightable->setHighlightDelta(0.0f, 0.0f, 0.0f, 0.05f);
--- a/src/engines/aurora/kotorjadegui/checkbox.cpp
+++ b/src/engines/aurora/kotorjadegui/checkbox.cpp
@@ -58,7 +58,10 @@ void WidgetCheckBox::load(const Aurora::GFF3Struct &gff) {
 	KotORJadeWidget::load(gff);
 
 	Border border = createBorder(gff);
-	Graphics::Aurora::TextureHandle texture = TextureMan.get(border.fill);
+	Graphics::Aurora::TextureHandle texture;
+
+	if (!border.fill.empty())
+		texture = TextureMan.get(border.fill);
 
 	_selected = gff.getStruct("SELECTED").getString("FILL");
 	_unselected = gff.getStruct("BORDER").getString("FILL");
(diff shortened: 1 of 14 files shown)


DrMcCoy:
(In case you haven't seen the compile error, you're missing a .c_str() here: https://travis-ci.org/xoreos/xoreos/jobs/404081867#L1859)

seedhartha:
Fixed.

src/engines/aurora/kotorjadegui/checkbox.cpp
```diff
@@ -58,22 +58,29 @@ void WidgetCheckBox::load(const Aurora::GFF3Struct &gff) {
 	KotORJadeWidget::load(gff);
 
 	Border border = createBorder(gff);
-	Graphics::Aurora::TextureHandle texture = TextureMan.get(border.fill);
+	Graphics::Aurora::TextureHandle texture;
+
```
DrMcCoy: Trailing whitespace

src/engines/kotor/object.cpp
```diff
@@ -42,7 +42,8 @@ Object::Object(ObjectType type)
 		  _currentHitPoints(0),
 		  _maxHitPoints(0),
 		  _minOneHitPoint(false),
-		  _room(0) {
+		  _room(0),
+		  _active(true) {
```
DrMcCoy:
```
In file included from /home/drmccoy/projects/xoreos/xoreos/src/engines/kotor/object.cpp:28:
/home/drmccoy/projects/xoreos/xoreos/src/engines/kotor/object.h: In constructor ‘Engines::KotOR::Object::Object(Engines::KotOR::ObjectType)’:
/home/drmccoy/projects/xoreos/xoreos/src/engines/kotor/object.h:167:8: error: ‘Engines::KotOR::Object::_room’ will be initialized after [-Werror=reorder]
  Room *_room; ///< Room the object is in.
        ^~~~~
/home/drmccoy/projects/xoreos/xoreos/src/engines/kotor/object.h:151:7: error:   ‘bool Engines::KotOR::Object::_active’ [-Werror=reorder]
  bool _active;
       ^~~~~~~
/home/drmccoy/projects/xoreos/xoreos/src/engines/kotor/object.cpp:38:1: error:   when initialized here [-Werror=reorder]
 Object::Object(ObjectType type)
 ^~~~~~
  CXX      src/engines/kotor/situated.lo
cc1plus: all warnings being treated as errors
```

src/engines/kotor/gui/ingame/partyselection.cpp
```diff
@@ -22,23 +22,145 @@
...
+		_config.selectedSlots.push_back(_config.forceNPC2);
+		++_numSelectedSlots;
+	}
+
+	setSlotTexture(0, (_config.trask ? "po_ptrask" : "po_pbastila"));
+
+	for (int i = 0; i < 9; ++i) {
+		toggleSlot(i, _config.slotEnabled[i]);
+
+		bool selected = (_config.forceNPC1 == i) ||
```
DrMcCoy: That's a const, even

src/engines/kotor/gui/ingame/partyselection.cpp
```diff
@@ -22,23 +22,145 @@
...
+		_returnCode = 2;
+	else if (tag.beginsWith("BTN_NPC")) {
+		Common::UString tmp(tag);
+		tmp.replaceAll("BTN_NPC", "");
+		int index;
+		Common::parseString(tmp, index);
+		_activeSlot = index;
+		onSelectionChanged();
+	}
+	else if (tag == "BTN_ACCEPT") {
```
DrMcCoy: The else needs to go onto the line with the closing brace

src/engines/kotor/gui/ingame/partyselection.cpp
```diff
@@ -22,23 +22,145 @@
...
+
+	for (int i = 0; i < 9; ++i) {
+		WidgetCheckBox *checkBox = getCheckBox("BTN_NPC" + Common::composeString(i));
+
+		if (std::find(_config.selectedSlots.begin(),
+		              _config.selectedSlots.end(),
+		              i) != _config.selectedSlots.end()) {
+			checkBox->setBorderColor(0.0f, 1.0f, 0.0f, 1.0f);
+		}
+		else if (_activeSlot == i)
```
DrMcCoy: The else needs to go onto the previous line here as well.

If you think this will look too squished, you can add a new line before the closing brace,

src/engines/kotor/gui/ingame/partyselection.cpp
```diff
@@ -22,23 +22,145 @@
...
+			checkBox->setBorderColor(1.0f, 1.0f, 0.0f, 1.0f);
+		else
+			checkBox->setBorderColor(0.0f, 0.0f, 0.0f, 0.0f);
+	}
+
+	bool activeSlotSelected = (_activeSlot != -1) &&
```
DrMcCoy: That's a const

seedhartha:
I tried deleting objects and it had broken execution of delayed actions: there was still a reference (not a pointer) to the object in global/local action variables. How would you handle that?

seedhartha:
Fixed the issues and de-hardcoded party members. Delayed actions problem is solved by manually resetting pointers to the deleted object.

DrMcCoy:
Hmm, you're right, the deleting is a bit complicated.

There's two issues with it still:
1) The debug printing FunctionManager::call(), if the debug level for the scripts is < 5, will try to print the tag of the parameter (the deleted object) after the call, when it's already deleted.
2) The delayed action might already have pushed the pointer to the deleted object onto its stack, if the timing is unfortunate. I had that happen a few times when testing this PR

1 is easily fixed by rewriting the function to format the signature into a string before calling the engine function. I have a commit here locally.

2 is more difficult. Maybe we need another approach here. Maybe we need to move over the object into a temporary "garbage" list? That list is then periodically checked when a delayed script ends. When no currently running delayed script uses this object, it can be reaped?

That will still pose a problem if we have two or more scripts using the object, running offset/shifted from each other, always renewing themselves, so that always one script is running using that object.

Also, PtrList has no way to "release" a member without deleting it right now.

A different approach would be to not use direct object pointers at all in the scripts. But only throw around object IDs (that are unique, i.e. using Common::generateIDNumber() which is backed by an atomic counting up). The scripts then need to query some kind of global (?) instance holding all objects. Deleted objects can then be handled more gracefully instead of derefencing invalid memory. All in-game containers could then also only use the ID.

The object ID thing is, I believe, what the original game does.

That's a way more complex change, though, and should be its own PR regardless. Not sure what to do with this PR in the meantime.

Do you have any thoughts on this matter?

seedhartha:
Using object ID does seem to be the right way to handle this. I may look into it in the future, but for now I've removed this functionality from the PR.

DrMcCoy:
> Using object ID does seem to be the right way to handle this

Yes, agreed.

I've actually been warned about this some time ago by niv (part of the NWN modding community, one of the people behind the community-run neverwintervault.org, IIRC now working for Beamdog on NWN:EE), and had it on my TODO list since. It never seemed so immediately pressing, though.

I guess I have some egg on my face now :)

> I may look into it in the future

That would be great.

> for now I've removed this functionality from the PR

Fair, yes.

PR #315, opened 2018-07-15
DRAGONAGE2: Implement DoAreaTransition nwscript method

Nostritius:
This PR should make sure, that the default area loads the actual start area.

### DIFF
--- a/src/engines/dragonage2/script/function_tables.h
+++ b/src/engines/dragonage2/script/function_tables.h
@@ -870,7 +870,7 @@ const Functions::FunctionPointer Functions::kFunctionPointers[] = {
 	{ 742, "GetPlotResRef"                         , 0                                                  },
 	{ 743, "GetPlotFlagName"                       , 0                                                  },
 	{ 744, "DisplayFloatyMessage"                  , 0                                                  },
-	{ 745, "DoAreaTransition"                      , 0                                                  },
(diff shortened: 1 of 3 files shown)


DrMcCoy:
Huh, second parameter is a script? Are you sure?

My DA2 script.ldf says:
```
/** @brief Moves the entire party into an area
*
* This function will do an area transtion if the target area is within the same area·
* list of the current area OR if the target area is in a different area list then
* it will unload the current area list and load into memory (with a loading bar)
* a new area list. The party will then be jumped to the specified waypoint (if it exists)
*
* @param sArea - tag of the target area
* @param sWaypointTag - tag of the target waypoint
* @returns 1 for success, 0 for failure (area list does not exist or area does not exist)
* @remarks this function will jump ALL party members
* @author Ross
*/
int DoAreaTransition( string sArea, string sWaypointTag ) = 745;
```

I.e. the second parameter should be a waypoint to jump to, not a script.

DrMcCoy:
Also, why are the gallows loaded at all anyways at the start? Is that the intro cutscene with Varric being interrogated?

Nostritius:
I corrected the the waypoint parameter. About the gallows, i'm as confused about this as you are. Maybe the varric cutscene takes place somewhere there? But since it is explicitly marked as entry area, i suppose we should leave it for now.

DrMcCoy:
Merged as d232ae1d0042aa405298263ca8d4fee150776b10, thanks! :)

PR #317, opened 2018-07-16
JADE: Fix a language encoding problem

Nostritius:
This PR should fix the encoding of jade to utf8. Since it was configured with the wrong encodings.


DrMcCoy:
Merged, thanks! :)

PR #320, opened 2018-07-18
AURORA: SACFile

Nostritius:
This PR adds support for sac files. These are essentially gff3 files with an extra header, used in Jade Empire as save files.

### DIFF
--- a/src/aurora/sacfile.cpp
+++ b/src/aurora/sacfile.cpp
@@ -1,0 +1,52 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  Handling GFF3 files with SAC header.
+ */
+
+#include "src/common/encoding.h"
+#include "src/common/memreadstream.h"
+
+#include "sacfile.h"
+
+namespace Aurora {
+
+SACFile::SACFile(Common::SeekableReadStream *stream) : GFF3File(load(stream)), _stream(stream) {
+	if (getType() != MKTAG('S', 'A', 'V', ' '))
+		throw Common::Exception("Invalid GFF ID");
+}
+
+Common::UString SACFile::getLevelFile() const {
+	return _levelFile;
+}
+
+Common::SeekableReadStream *SACFile::load(Common::SeekableReadStream *stream) {
+	stream->skip(4); // Unknown value, probably a version header?
+	const uint32 nameLength = stream->readUint32LE(); // Length of the level identifier.
+
+	_levelFile = Common::readStringFixed(*stream, Common::kEncodingASCII, nameLength); // The Level identifier string.
+
+	stream->skip(4); // Unknown value, probably a checksum?
+
+	return new Common::SeekableSubReadStream(stream, stream->pos(), stream->size());
+}
+
+} // End of namespace Aurora
--- a/src/aurora/sacfile.h
+++ b/src/aurora/sacfile.h
@@ -1,0 +1,54 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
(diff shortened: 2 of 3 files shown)


src/aurora/sacfile.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+Common::SeekableReadStream *SACFile::load(Common::SeekableReadStream &stream) {
+	stream.skip(4); // Unknown value, probably a version header?
+	uint32 nameLength = stream.readUint32LE(); // Length of the level identifier.
+
+	_levelFile = Common::readStringFixed(stream, Common::kEncodingASCII, nameLength); // The Level identifier string.
+
+	stream.skip(4); // Unknown value, probably a checksum?
+
+	return stream.readStream(stream.size() - stream.pos());
```
DrMcCoy: Is there a reason you're doing the extra read/copy here?

You could instead construct a `Common::SeekableSubReadStream` for the range of the "inner" GFF data.

Of course, the parent stream needs to be for as long as the sub stream exists, so you need to change the `SACFile` constructor to take a `Common::SeekableReadStream *`, and take over that stream (and keep it around, and delete it when the `SACFile` is deconstructed, but, for example, stuffing it into a `ScopedPtr`).

This mirrors how `GFF3File` works already, too.

Nostritius:
The issue should be fixed

src/aurora/sacfile.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  Handling GFF3 files with SAC header.
+ */
+
+#ifndef XOREOS_SACFILE_H
```
DrMcCoy: The include guard is wrongly named

src/aurora/sacfile.h
```diff
@@ -0,0 +1,54 @@
...
+private:
+	Common::UString _levelFile;
+	Common::ScopedPtr<Common::SeekableReadStream> _stream;
+
+	Common::SeekableReadStream *load(Common::SeekableReadStream *stream);
+};
+
+} // End of namespace Aurora
+
+#endif //XOREOS_SACFILE_H
```
DrMcCoy: There needs to be a space after the // (also, wrongly named)

src/aurora/sacfile.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+Common::SeekableReadStream *SACFile::load(Common::SeekableReadStream *stream) {
+	stream->skip(4); // Unknown value, probably a version header?
+	uint32 nameLength = stream->readUint32LE(); // Length of the level identifier.
+
+	_levelFile = Common::readStringFixed(*stream, Common::kEncodingASCII, nameLength); // The Level identifier string.
+
+	stream->skip(4); // Unknown value, probably a checksum?
+
+	return new Common::SeekableSubReadStream(stream,stream->pos(), stream->size());
```
DrMcCoy: A space after a parameter comma, please.

src/aurora/sacfile.cpp
```diff
@@ -0,0 +1,52 @@
...
+		throw Common::Exception("Invalid gff id");
+}
+
+Common::UString SACFile::getLevelFile() const {
+	return _levelFile;
+}
+
+Common::SeekableReadStream *SACFile::load(Common::SeekableReadStream *stream) {
+	stream->skip(4); // Unknown value, probably a version header?
+	uint32 nameLength = stream->readUint32LE(); // Length of the level identifier.
```
DrMcCoy: That's const

src/aurora/sacfile.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "src/common/encoding.h"
+#include "src/common/memreadstream.h"
+
+#include "sacfile.h"
+
+namespace Aurora {
+
+SACFile::SACFile(Common::SeekableReadStream *stream) : GFF3File(load(stream)), _stream(stream) {
+	if (getType() != MKTAG('S', 'A', 'V', ' '))
+		throw Common::Exception("Invalid gff id");
```
DrMcCoy: Capitalize GFF, please, and ID as well

src/aurora/sacfile.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "src/common/readstream.h"
+
+#include "src/aurora/gff3file.h"
+
+namespace Aurora {
+
+/**
+ * A SAC file is basically a GFF3 file with some bytes extra header. This
+ * class inherits from the GFF3File class to offer all gff access methods
```
DrMcCoy: Capitalize GFF, please

Nostritius:
The Issues should now be fixed. I don't want to write a unit test yet, since i'm not sure if my assumptions about the data of the header are correct. If someone could provide me saves from as many stages of the game as possible, i could verify this, since i do not have souch a collection.

DrMcCoy:
Okay, just keep it in mind :).

I mean, I do have saves from, IIRC, two playthroughs, but they're "personal" and I always feel kinda weird about distributing saves from my own playthroughs. Does that make sense?

You might want to Google for Jade Empire saves, maybe. Could be somebody already provided them?

farmboy0:
I have been playing KOTOR1 up to the 4th planet and could provide those saves if you want them.

Nostritius:
Sorry, I need Jade Empire saves, at least the header and some hint about the state of the save.

PR #321, opened 2018-07-30
KOTOR: Fix the loading screen in the saveload menu

Nostritius:
I experienced a bug in the loading menu, where the where the loading screen overlapped with the load menu. This should fix this.


DrMcCoy:
Merged as b56b28c9b, thanks!

PR #322, opened 2018-07-31
KOTOR: Some ingame cursors

Nostritius:
This PR adds some ingame cursors, which will appear, when hovering over specific objects. Also i changed the click event from a mouse down to a mouse up event since this matches more the overall behaviour of the game.

### DIFF
--- a/src/engines/kotor/creature.cpp
+++ b/src/engines/kotor/creature.cpp
@@ -35,6 +35,7 @@
 #include "src/graphics/aurora/modelnode.h"
 #include "src/graphics/aurora/model.h"
 #include "src/graphics/aurora/animationchannel.h"
+#include "src/graphics/aurora/cursorman.h"
 
 #include "src/engines/aurora/util.h"
 #include "src/engines/aurora/model.h"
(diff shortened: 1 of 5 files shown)


src/engines/kotor/gui/gui.cpp
```diff
@@ -93,11 +93,11 @@ Common::UString GUI::getName() const {
 }
 
 void GUI::mouseDown() {
-	CursorMan.setState("down");
+	CursorMan.set(CursorMan.getCurrentGroup(), "down");
```
DrMcCoy: Huh...

`setState()` is supposed to do that, setting the cursor to another state in the current group.
Likewise, `setGroup()` is supposed to set a new group, with the current state.

The comments in cursorman.h even say that. However, the code does something else.
I wrote both, so I wonder what happened. :P

In either case, I think the code in cursorman.cpp should be changed to actually do what it claims to do.

Nostritius:
The cursorman should now be fixed

src/engines/kotor/creature.cpp
```diff
@@ -576,10 +577,12 @@ void Creature::createPC(const CharacterGenerationInfo &info) {
 }
 
 void Creature::enter() {
+	CursorMan.set("talk", CursorMan.getCurrentState());
```
DrMcCoy: That should be `CursorMan.setGroup("talk");`.

Similarly for the following two commits.

Nostritius:
CursorMan.set should now be replaced to CursorMan.setGroup.

DrMcCoy:
Merged, thanks! :)

PR #323, opened 2018-07-31
KOTOR: Add spinning map arrow

Nostritius:
This PR makes the arrow in the minimap spin around on the camera angle.

### DIFF
--- a/src/engines/aurora/kotorjadegui/kotorjadewidget.cpp
+++ b/src/engines/aurora/kotorjadegui/kotorjadewidget.cpp
@@ -441,6 +441,11 @@ void KotORJadeWidget::setPosition(float x, float y, float z) {
 	}
 }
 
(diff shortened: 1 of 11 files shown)


DrMcCoy:
You're not resetting the orientation on module change.

That means when you rotate the camera and with it the arrow, and then load another module (for example via loadmodule console command), the arrow starts to point in the wrong direction.

Nostritius:
this problem should now be fixed by setting the current rotation on entering an area.

DrMcCoy:
...It still doesn't work, though. The arrow still moves side-ways and weird upon changing the areas

Nostritius:
The rotation was not applied depending on the north axis. Ichanged this, and now the arrow should rotate different depending on the North Axis.

DrMcCoy:
Nice, this looks good now.

Merged, thanks! :)

PR #324, opened 2018-07-31
KOTOR: Fix transparent borders in minimaps

Nostritius:
This PR should fix the transparent areas on the minimap


DrMcCoy:
Merged as 127f21747, thanks!

PR #328, opened 2018-08-07
KOTOR: Fix the wrong function in setGlobalBoolean

Nostritius:
I recognized a small bug in the setGlobalBoolean nwscript function.


DrMcCoy:
Merged, thanks! :)

PR #330, opened 2018-08-10
KOTOR: Add DestroyObject nwscript method

Nostritius:
This nwscript method should let trask disappear after you equipped and talked to him to later replace him by a version of him placed as your companion.


DrMcCoy:
This won't work. Please see the comments in this PR: https://github.com/xoreos/xoreos/pull/314#issuecomment-405227688

PR #331, opened 2018-08-12
KOTOR: Character generation model change

Nostritius:
This PR recreates the heads of the models if thei portraits are changed in the character generation.

### DIFF
--- a/src/engines/kotor/gui/chargen/chargeninfo.cpp
+++ b/src/engines/kotor/gui/chargen/chargeninfo.cpp
@@ -197,68 +197,27 @@ Creature *CharacterGenerationInfo::getCharacter() const {
 	return creature.release();
 }
 
-Graphics::Aurora::Model *CharacterGenerationInfo::getModel() {
-	// TODO: This duplicates code in Creature::createPC()
+void CharacterGenerationInfo::recreateHead() {
+	Graphics::Aurora::Model *head = _head;
+	_head = loadModelObject(Creature::getHeadMeshString(getGender(), getSkin(), getFace()), "");
+	_body->attachModel("headhook", _head);
+	delete head;
 
+	// Restart animation to apply it to the new head.
+	// TODO: Possibly there is another way of doing this, without restarting the animation?
+	_body->playAnimation("pause1", true, -1);
+}
+
+Graphics::Aurora::Model *CharacterGenerationInfo::getModel() {
 	if (_body)
 		return _body.get();
 
-	Common::UString body, head;
-
-	body = "p";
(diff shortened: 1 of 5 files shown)


src/engines/kotor/gui/chargen/chargeninfo.cpp
```diff
@@ -197,6 +197,48 @@ Creature *CharacterGenerationInfo::getCharacter() const {
 	return creature.release();
 }
 
+void CharacterGenerationInfo::recreateHead() {
```
DrMcCoy: This is like the third or fourth copy of the same code, here in CharacterGenerationInfo and also in class Creature.

Can you abstract this away, so that there's one function that creates the model name for the head and one for the body, used by both class Creature and class CharacterGenerationInfo?

Nostritius:
I have now removed the code duplication in the chargen info class.

DrMcCoy:
Sure, but this still means the code duplication in `class CharacterGenerationInfo` and `class Creature` is still there.

I was more imagining static methods in `class Creature` that create the strings and get called by both `Creature` and `CharacterGenerationInfo`.

I.e., basically, move `getBodyMeshString()` and `getHeadMeshString()` to `Creature`, make them static and make them take gender, class, skin, face, state (where used) as parameters. Likewise, `Creature` needs a `getBodyTexture()` or somesuch static method.

DrMcCoy:
(I might not have communicated what I would have liked clearly enough, though, sorry)

Nostritius:
I have thought about this, and since for the character generation, only the standard uniform is required and generated by the chargen info and the creature class needs a far more complex function with more parameters,  i would consider leaving it this way for now.

DrMcCoy:
It's way less complex than it looks. That's why I'd like to split the different strings into their own methods, unified for the generator and for Creature.

- One for the head. That's the same in Creature and CharacterGenerationInfo
- One for the body. It's just one extra parameter, state, that's pinned to "b" in the character generator
- One for the body texture. Only used in Creature
- One for the portrait. Only used in Creature. Shares a common base with the head, so maybe that can be separated into a function as well (getHeadBase()?)

If you trace it out by hand, it's all really the same. It just looks different because it's a bit of a mess.

DrMcCoy:
Those heads look, though: https://gist.github.com/DrMcCoy/731b590cf58d6c97432fcb11d371476a

Nostritius:
This should now hopefully fix the problem with the head leak.

DrMcCoy:
Merged as 6c7e9c6e2...abcde6d63, thanks! :)

PR #333, opened 2018-08-18
ACTIONSCRIPT: Array class and some minor updates

Nostritius:
This PR introduces the array class to actionscript and fixes several minor bugs in the ActionScript interpreter


src/aurora/actionscript/variable.h
```diff
@@ -66,6 +66,8 @@ class Variable {
 	const Common::UString &asString();
 	bool asBoolean();
 
+	template<typename T> boost::shared_ptr<T> as();
```
DrMcCoy: You're using that functionality in 8b02e03dc761c51efc53bc63f3794899142e779e, which is *before* this commit.

Please order your commits so that every single commit still compiles. I.e. order this commit *before* the "Add basic array class" one.

This is really crucial for looking for regression with "git bisect". When there's a commit in the middle that doesn't compile, that doesn't exactly break the bisect, but it adds annoyances.

src/aurora/actionscript/object.cpp
```diff
@@ -80,6 +80,11 @@ Variable Object::call(Common::UString function, AVM &avm) {
 		counter += 1;
 	}
 
+	for (int i = 0; i < arguments.size(); ++i) {
```
DrMcCoy: `error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<Aurora::ActionScript::Variable>::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]`

Make that `i` a `size_t`, please.

src/aurora/actionscript/object.cpp
```diff
@@ -65,7 +65,7 @@ void Object::setMember(Common::UString id, Function *function) {
 	_members[id] = ObjectPtr(function);
 }
 
-Variable Object::call(Common::UString function, AVM &avm) {
+Variable Object::call(Common::UString function, AVM &avm, std::vector<Variable> arguments) {
```
DrMcCoy: That can be a `const std::vector<Variable> &`, too.

src/aurora/actionscript/variable.cpp
```diff
@@ -25,6 +25,8 @@
 #include "src/aurora/actionscript/variable.h"
 #include "src/aurora/actionscript/object.h"
 #include "src/aurora/actionscript/function.h"
+#include "variable.h"
```
DrMcCoy: Full paths, please.

Also, src/aurora/actionscript/variable.h is already included in line 25.

src/aurora/actionscript/object.cpp
```diff
@@ -48,17 +48,24 @@ bool Object::hasMember(Common::UString id) {
...
 }
 
-Variable Object::getMember(Common::UString id) {
-	if (_members.find(id) != _members.end())
-		return _members[id];
+Variable Object::getMember(Variable id) {
+	if (!id.isString())
+		throw Common::Exception("Object::getMember id is not a string");
+
+	Common::UString idString = id.asString();
```
DrMcCoy: That's a const.

src/aurora/actionscript/object.cpp
```diff
@@ -48,17 +48,24 @@ bool Object::hasMember(Common::UString id) {
 	return _members.find(id) != _members.end();
 }
 
-Variable Object::getMember(Common::UString id) {
-	if (_members.find(id) != _members.end())
-		return _members[id];
+Variable Object::getMember(Variable id) {
```
DrMcCoy: `const Variable &`

src/aurora/actionscript/variable.h
```diff
@@ -95,6 +97,10 @@ class Variable {
 	} _value;
 };
 
+template<typename T> boost::shared_ptr<T> Variable::as() {
```
DrMcCoy: That's a `const` method.

src/aurora/actionscript/object.cpp
```diff
@@ -48,17 +48,24 @@ bool Object::hasMember(Common::UString id) {
...
-		_members.insert(std::make_pair(id, ObjectPtr(new Object)));
-		return _members[id];
+		_members.insert(std::make_pair(idString, ObjectPtr(new Object)));
+		return _members[idString];
 	}
 }
 
-void Object::setMember(Common::UString id, Variable member) {
-	_members[id] = member;
+void Object::setMember(Variable id, Variable value) {
```
DrMcCoy: `const Variable &id`, `const Variable &value`

Nostritius:
I have now fixed (hopefully all issues, and since i replaced the get, set and hasMember Variables through const references in the end, i assume this should also be ok.

DrMcCoy:
Merged as 55879aad0...7577e4e23, thanks! :)

PR #336, opened 2018-08-19
DRAGONAGE2: Add xbox360 probe

Nostritius:
As the name suggests, this PR adds a probe for the xbox360 version of dragonage2.
In the long term it might be worth looking into some of the archive formats utilized by the xbox360, namely the pirs format (mostly used for content acquired from the xbox store), the con format (for dvd content installed to the hard drive) and xbox dvd images.
This will not work, since the xbox image is in some ways different from the pc image but at least it will now recognize the xbox version correctly.

### DIFF
--- a/src/engines/dragonage2/probes.cpp
+++ b/src/engines/dragonage2/probes.cpp
@@ -88,12 +88,32 @@ public:
 	}
 };
 
+class EngineProbeXbox360 : public EngineProbe {
+public:
+	Aurora::Platform getPlatform() const {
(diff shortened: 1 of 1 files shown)


DrMcCoy:
I don't remember exactly what, but I think I remember holding that back because I saw...a copy-paste error? Something along those lines. Dunno why I didn't add a comment then.

I'll recheck this in the evening.

DrMcCoy:
Merged as 833b918d4853681d0324925f6a931749de949169, thanks! :)

PR #337, opened 2018-08-19
DRAGONAGE: Add xbox360 probe

Nostritius:
This PR adds an xbox 360 probe for dragonage.


DrMcCoy:
Merged as 33bef0158, thanks! :)

PR #338, opened 2018-08-25
SOUND: Fix xact ascii wavebank compiling without libvorbis

Nostritius:
There is a bug, when compiling without vorbis in the xact ascii wave banks. This PR should fix this.

### DIFF
--- a/src/sound/xactwavebank_ascii.cpp
+++ b/src/sound/xactwavebank_ascii.cpp
@@ -63,7 +63,11 @@ RewindableAudioStream *XACTWaveBank_ASCII::getWave(size_t index) const {
 	if (!dataStream)
 		throw Common::Exception("XACTWaveBank_ASCII::getWave(): No such resource \"%s\"", wave.name.c_str());
 
+#ifdef ENABLE_VORBIS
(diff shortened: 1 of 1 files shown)


src/sound/xactwavebank_ascii.cpp
```diff
@@ -63,7 +63,11 @@ RewindableAudioStream *XACTWaveBank_ASCII::getWave(size_t index) const {
 	if (!dataStream)
 		throw Common::Exception("XACTWaveBank_ASCII::getWave(): No such resource \"%s\"", wave.name.c_str());
 
+#ifdef ENABLE_VORBIS
 	return makeVorbisStream(dataStream.release(), true);
+#else
+	return 0;
```
DrMcCoy: It should even throw, with a message like https://github.com/xoreos/xoreos/blob/master/src/sound/sound.cpp#L295

Nostritius:
Ahh, i already thought about something like this, but then i thought it might run with the null pointer simply without this sound. Nevertheless, it's changed.

DrMcCoy:
Merged as 61fe0aa0d, thanks! :)

PR #339, opened 2018-08-25
KOTOR: Add getPCSpeaker nwscript function

Nostritius:
This pr adds the getPCSpeaker nwscript function, which should enable trasks class specific lines in the first dialog.


DrMcCoy:
Merged as dcecda7d9, thanks! :)

PR #341, opened 2018-08-26
NWScript object management

seedhartha:
This PR makes NWScript variables store object ids, not object pointers. It introduces two new classes: ObjectReference, which holds the object id, and ObjectManager, which acts as the global registry for objects. This PR also implements destroyObject function in KotOR.


src/engines/sonic/placeable.cpp
```diff
@@ -53,11 +56,15 @@ namespace Sonic {
 
 Placeable::Placeable(const Aurora::GFF4Struct &placeable) : Object(kObjectTypePlaceable),
 	_typeID(0xFFFFFFFF), _appearanceID(0xFFFFFFFF), _scale(1.0f) {
+	_id = Common::generateIDNumber();
```
DrMcCoy: Hmm, Placeable::load() reads the _id out of the GFF, though. There's probably no reason it has to be this _id, though, so you can probably add a new variable, say _placeableID or something, and assign the value from the GFF there.

src/engines/dragonage/object.cpp
```diff
@@ -47,6 +50,9 @@ using ::Aurora::GFF4List;
 using namespace ::Aurora::GFF4FieldNamesEnum;
 
 Object::Object(ObjectType type) : _type(type), _static(true), _usable(false) {
+	_id = Common::generateIDNumber();
```
DrMcCoy: For areas, the _id gets overwritten in Area::loadEnvironment(), with a value from the GFF. You'll need to add a commit adding a, for example _envID value to Area and change the assignment in loadEnvironment to use write into that value instead.

Same applies to Dragon Age II as well.

DrMcCoy:
Otherwise, I like this PR. Looks to me like it goes in the right direction :)

seedhartha:
Fixed the above issues.

DrMcCoy:
Sorry, got another use-after-free here: https://gist.github.com/DrMcCoy/549d6f2ed5ea7532539ee148a02f75a7

The issue here is that Module::_delayedActions contains raw Object pointers (ActionQueue, which is a std::multiset<Action>, struct Action members owner and triggerer)

seedhartha:
Done.

DrMcCoy:
I can't seem to trigger any crashes now, yes. :)

However, could you apply bc7047f948aac8de0f9b54b4fc9437af2c7c6221 to the other engines? All except for Sonic have that `struct Action` with a raw Object pointer in module.h.

seedhartha:
Modified actions/events in other engines as well.

seedhartha:
Have you intentionally not included the last DA commits or is something wrong with GitHub today?

DrMcCoy:
Huh, yeah, I somehow missed them.

GitHub *is* weird today, a bit slow with the updates. But this might be completely on me, I think I forgot to update the remote before merging. I'll add them, sorry

DrMcCoy:
There, added them as 61cd69b57...6c7e9c6e2. And made sure the commits I merged are the correct ones and nothing went missing.

PR #342, opened 2018-08-26
KOTOR: Make the party member symbols immediately appear and disappear

Nostritius:
If adding a new character portrait to the hud, it won't appear unless the hud is hidden and shown again. This PR shows or hides the specific elemts immediately


DrMcCoy:
Is that supposed to show Trask's portrait in the HUD when he joins? Because it doesn't, for me, even in combination with #343.

Nostritius:
Currently, he is not really joining the party, even after the selection screen is shown. You have to add some code to get the id from the selection screen, get the template from it, create a creature from it, and then set this creature as ingame gui party member, then you should see the image of him. Initially I wanted to make one PR for the character joining and switching, but i decided to do this after #341 and push what i have and what won't interfere with #341.

DrMcCoy:
Merged as 0729b4cf4, thanks! :)

PR #343, opened 2018-08-26
KOTOR: Fixes and preparation for character switching

Nostritius:
This PR fixes and prepares some things to enable the creation of and switching between party members, especially intended for the first dialog when you need to switch to trask to actually open the door.


DrMcCoy:
Merged as 0729b4cf4...2c7cdc0a5, thanks! :)

PR #344, opened 2018-08-27
KOTOR: Enable dialogs for situated objects

Nostritius:
Situated objects like doors, can have dialogs too. This PR should enable them.


DrMcCoy:
Merged as 2c7cdc0a5...1bfacf150, thanks! :)

PR #350, opened 2018-08-28
COMMON: Add signed bits bitstream method

Nostritius:
This PR adds the possibility to get a signed bit values from a bitstream. This is important for scaleform gfx loading.

### DIFF
--- a/src/common/bitstream.h
+++ b/src/common/bitstream.h
@@ -63,6 +63,9 @@ public:
 	/** Read a multi-bit value from the bit stream. */
 	virtual uint32 getBits(size_t n) = 0;
 
+	/** Read a multi-bit signed value from the bit stream. */
+	virtual int32 getSignedBits(size_t n) = 0;
+
 	/** Add a bit to the n-bit value x, making it an (n+1)-bit value. */
 	virtual void addBit(uint32 &x, size_t n) = 0;
 
@@ -194,6 +197,19 @@ public:
 
 			v >>= (32 - n);
 		}
+
+		return v;
+	}
+
+	/** Read a multi-bit signed value from the bit stream. */
+	int32 getSignedBits(size_t n) {
+		if (n == 0)
+			return 0;
+
+		int32 v = getBits(n);
+
+		if (v >> (n - 1))
+			v |= 0xFFFFFFFF << n;
 
 		return v;
 	}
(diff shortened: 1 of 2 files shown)


src/common/bitstream.h
```diff
@@ -198,6 +201,19 @@ class BitStreamImpl : boost::noncopyable, public BitStream {
 		return v;
 	}
 
+	/** Read a multi-bit signed value from the bit stream. */
+	int32 getSignedBits(size_t n) {
+		int32 v = getBits(n);
+
+		bool isSigned = v >> (n - 1);
```
DrMcCoy: This will break for n == 0

src/common/bitstream.h
```diff
@@ -198,6 +201,19 @@ class BitStreamImpl : boost::noncopyable, public BitStream {
...
+		if (isSigned) {
+			int h = (v >> (n - 1)) << (n - 1);
+			v = (v ^ h) - h;
```
DrMcCoy: Why so convoluted?

Nostritius:
I made an extra condition for n == 0 and simplified the negating a bit. Shall i simplify it even more?

DrMcCoy:
1) You can just check for n == 0 at the start and return 0, no?
2) Yes, please simplify it further

Nostritius:
I made a if (n == 0) statement and simplified the expression even further.

src/common/bitstream.h
```diff
@@ -198,6 +201,19 @@ class BitStreamImpl : boost::noncopyable, public BitStream {
...
 
+	/** Read a multi-bit signed value from the bit stream. */
+	int32 getSignedBits(size_t n) {
+		if (n == 0)
+			return 0;
+
+		int32 v = getBits(n);
+
+		if (v >> (n - 1))
+			v |= 0xFFFFFFFF << n;
```
DrMcCoy: Technically, that might produce a value that traps on some negative value representations.

What is it you want to accomplish? Isn't it just a masking away of the sign bit and negating the value? 

Why don't you just do a & with the ~ of the bit shifted to the correct position (on the unsigned value), cast to signed and then unary-minus the whole thing? That makes it way clearer what's going on, and is more portable.

DrMcCoy: I mean, all these bit fiddling tricks you'd normally do if you want to get rid of the branch, but you've got an explit branch with the if there anyway...

Nostritius:
We agreed, that this functionality should be part of the scaleform gfx loader.

PR #351, opened 2018-08-29
KOTOR: Fix highlightAll after object deletion

seedhartha:


DrMcCoy:
Merged as 374818af4, thanks! :)

PR #352, opened 2018-08-29
JADE: Register objects with ObjectManager

seedhartha:


DrMcCoy:
Merged as 5516f8f6d, thanks! :)

PR #354, opened 2018-08-29
KOTOR: Fix crash on head change in character generator

seedhartha:


src/engines/kotor/gui/chargen/chargeninfo.cpp
```diff
@@ -201,7 +201,10 @@ void CharacterGenerationInfo::recreateHead() {
 	Graphics::Aurora::Model *head = _head;
 	_head = loadModelObject(Creature::getHeadMeshString(getGender(), getSkin(), getFace()), "");
 	_body->attachModel("headhook", _head);
+
+	GfxMan.lockFrame();
 	delete head;
+	GfxMan.unlockFrame();
```
DrMcCoy: The attachModel() call should probably be moved within the locking as well

clone2727: We should probably have an RAII class for handling the lock/unlock (if we don't already).

seedhartha:
Done.

DrMcCoy:
Merged as 71c97acf8, thanks! :)

PR #356, opened 2018-09-02
ACTIONSCRIPT: Segment violation fix and minor updates

Nostritius:
This PR fixes a segment violation bug and adds some other updates.


DrMcCoy:
Merged as 0555d8b4b...66926436d, thanks! :)

PR #359, opened 2018-09-04
ACTIONSCRIPT: Fix string checks in actionGetURL

Nostritius:
This PR fixes a bug in the actionGetURL opccode of actionscript. Thanks to @darkstar for pointing me to this.


DrMcCoy:
Merged as 0555d8b4b, thanks! :)

PR #360, opened 2018-09-04
KOTOR: Character Switching

Nostritius:
This PR enables switching between characters. After Trask joined the party, you should be able to switch to him by clicking on his portrait.


DrMcCoy:
Merged as 66926436d...94d91e0eb, thanks! :)

PR #361, opened 2018-09-05
COMMON: Add rectangle helper typedef and operators

Nostritius:
This PR adds a Rect helper struct inside the Commmon library. Since Scaleform needs some rectangle hanlding.

### DIFF
--- a/src/common/rect.h
+++ b/src/common/rect.h
@@ -1,0 +1,56 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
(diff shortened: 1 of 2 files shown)


DrMcCoy:
Eh, not really a fan of just further typedef'ing SDL types like that.

Also, I really don't like having these operators in the global namespace like that.

Nostritius:
Well, i tried to move them into the namespace first, but the operators won't be recognized. But nevertheless i will change this. Shall I write something on my own, or do we have a similar class already? I need this for the scaleform code, since it handles a lot of rectangles (It is actually the only thing preventing me from publishing my gfx file loader, the first part of my gfx code).

Nostritius:
Closing this for #363

PR #363, opened 2018-09-11
COMMON: Add rectangle class

Nostritius:
My second try for a rectangle class.

### DIFF
--- a/src/common/rect.cpp
+++ b/src/common/rect.cpp
@@ -1,0 +1,36 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *  A simple rectangle helper class.
+ */
+
+#include "src/common/rect.h"
+
+namespace Common {
+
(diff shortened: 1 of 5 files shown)


src/common/rect.h
```diff
@@ -0,0 +1,46 @@
...
+	Rect();
+
+	operator bool() const;
+	bool operator!() const;
+
+	bool operator==(const Rect &rect) const;
+	bool operator!=(const Rect &rect) const;
+
+	double x, y;
+	double w, h;
```
DrMcCoy: You sure you want to work with floating point values here, instead of integers like SDL_Rect?

Keep in mind that absolute equality of two such rectangles (i.e. `operator==()`) that have been created by different calculations might be difficult. You might need to substract the coordinates and compare to an epsilon instead.

Not saying that you should do that *now*, just something to keep in mind if you really want to keep the doubles and depend on absolute equality.

clone2727: Or, at the very least, make the underlying type a template parameter.

Nostritius:
After thinking about it, i decided to remove the equality operators, since i do not actually need them.

src/common/rect.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "src/common/rect.h"
+
+namespace Common {
+
+Rect::Rect() : x(0.0), y(0.0), w(0.0), h(0.0) {
+}
+
+Rect::operator bool() const {
+	return x != w && y != h;
```
DrMcCoy: Also, are you sure that's correct?

So if you have this rectangle:

```
(5/5)      (10/5)
  .----------.
  |          |
  .__________.
(5/10)     (10/10)  
```

then

```
x = 5.0
y = 5.0
w = 5.0
h = 5.0
```

So this would return false.

This method would work if you'd use x1, y1, x2, y2, and not x, y, w, h. With width and height, you need to check for `w != 0.0 && h != 0.0`, no?

Nostritius:
You're right, i corrected this

DrMcCoy:
Also, thinking about it, why is it `operator bool()` anyway? I'd rather have a `bool empty() const` method (and no notEmpty() or something, just the empty() is enough).

Also also, what about rects with negative weight or height? Is that a valid rect for you? If not, you should probably make the empty check `x <= 0.0 || y <= 0.0`.

Nostritius:
On both points you're right, i will correct this

DrMcCoy:
Ooops, I mean `w <= 0.0 || h <= 0.0`, of course. Now I'm getting all confused, too :P

Nostritius:
I have now fixed this for negative values and added a unit test

DrMcCoy:
Merged as 9d1fb8b03...4421a039e, thanks! :)

PR #367, opened 2018-09-18
AURORA: New FEV loading capabilities

Nostritius:
Add some new values which can now be read from the FEVFile loader

### DIFF
--- a/src/aurora/fevfile.cpp
+++ b/src/aurora/fevfile.cpp
@@ -119,7 +119,8 @@ void FEVFile::readCategory(Common::SeekableReadStream &fev) {
 void FEVFile::readEventCategory(Common::SeekableReadStream &fev) {
 	Common::UString name = readLengthPrefixedString(fev);
 
-	fev.skip(4); // Unknown value
+	// Read user properties
+	readProperties(fev);
 
 	const uint32 numSubEventCategories = fev.readUint32LE();
 	const uint32 numEvents = fev.readUint32LE();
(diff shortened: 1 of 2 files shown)


src/aurora/fevfile.h
```diff
@@ -45,6 +47,14 @@ class FEVFile {
 		kStreamFromDisk
 	};
 
+	/** Some objects in FMOD can have generic properties. */
+	struct Property {
+		uint32 type;
+		int32 integerValue;
+		float floatValue;
+		Common::UString stringValue;
```
DrMcCoy: Hmm, that same pattern, a variable with multiple potential types, comes up again and again in the codebase now. In the ActionScript code, the NWScript code, the Lua code, etc.

Maybe be should see if we can use Boost.Any or Boost.Variant for this instead? This code might be the ideal test environment for that. Could you see if you can do that?

For the differences between Any and Variant, see https://www.boost.org/doc/libs/1_53_0/doc/html/variant/misc.html#variant.versus-any . I'd say Variant sounds like the better candidate, with Any a fall-back if Variant should not work for our use case.

You probably still need variable to store the current type, though. I'd rather that be an enum instead of just an integer without any obvious meaning, too.

Nostritius:
I have now changed the properties to use an enum and boost::variant. As for the other occasions with such constructs (at least the ones I wrote), I will change them in other PRs.

src/aurora/fevfile.cpp
```diff
@@ -174,11 +175,42 @@ void FEVFile::readEvent(Common::SeekableReadStream &fev) {
...
 
+std::map<Common::UString, FEVFile::Property> FEVFile::readProperties(Common::SeekableReadStream &fev) {
+	const uint32 numUserProperties = fev.readUint32LE();
+	std::map<Common::UString, FEVFile::Property> properties;
+	for (uint32 i = 0; i < numUserProperties; ++i) {
+		Common::UString propertyName = readLengthPrefixedString(fev);
+		Property property;
+		property.type = PropertyType(fev.readUint32LE());
+		switch (property.type) {
+			case 0: // Integer value
```
DrMcCoy: Can you replace these with the enum value as well, please?

Nostritius:
Done

DrMcCoy:
Merged as 204cccc73...29ce6338c, thanks! :)

DrMcCoy:
Btw, upon further thought, I think this belongs into sound/, not into aurora/.

Nostritius:
Hmm, but SSFFile is also in aurora?

DrMcCoy:
I consider FEV to be more connected to FMOD wavebanks than SSFs are to
sound files.

In fact, my "vision" for FEV, and also to the analogues files for XACT
and Wwise, is that a FEV would, by itself, be able to start playing a
sound. I.e. the engine would load a FEV, that FEV would maybe even
load the relevant wavebanks, and then the engine would tell the FEV to
play a sound, and the FEV would handle the rest.

XACT and Wwise would have similar interfaces.

PR #374, opened 2018-09-30
KOTOR2: GUI Scaling

Nostritius:
This PR implements the UI Scaling for kotor2, as the original game did. This PR will lead to some ugly looking results regarding the border quads. I want to wait for @mirv-sillyfish until i fix this, since it is heavily opengl related.. Nevertheless, it is not interfering the playability of kotor2.

### DIFF
--- a/src/engines/kotor2/gui/gui.cpp
+++ b/src/engines/kotor2/gui/gui.cpp
@@ -22,25 +22,48 @@
  *  A KotOR2 GUI.
  */
 
-#ifndef ENGINES_KOTOR2_GUI_GUI_H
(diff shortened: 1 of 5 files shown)


farmboy0:
Is there a reason you are not using the GUI scaling that is already in the GraphicsManager?

Nostritius:
Thanks for the hint. I tried this, but the problem is, that it scales everything, which leads to aliasing problems withthe text and borders.
bildschirmfoto vom 2018-10-01 12-38-07

DrMcCoy:
Merged as 4830f8f93...6e0ad44f1, thanks! :)

PR #375, opened 2018-10-03
KOTOR: Reset to default cursor, when dialog occurs

Nostritius:
I recognized, that when in a dialog, the cursor stays the sae, if for example a door is hovered. This should fix this for kotor.


DrMcCoy:
Merged as 204cccc73, thanks! :)

PR #379, opened 2018-10-22
VIDEO: Extend Actimagine header

Nostritius:
I have found another source of information for the Actimage VX format, called MobiclipDecoder,  where the author read the complete header information. This PR extends the actimagine decoder to read almost the entire header data. It is now even possible to read the audio data, although i don't know, which compression is applied there.


DrMcCoy:
Kostya looked at VX a few years ago and said the audio looks like some form of ADPCM. There's of course 1000 ADPCM schemes, but maybe it can be figured out

DrMcCoy:
Btw, this PR doesn't apply anymore, because of @clone2727's video changes. Can you fix this up, please? :)

Nostritius:
The changes should now be applied as far as possible

PR #380, opened 2018-10-28
KOTOR: Fill remaining gui background with black

Nostritius:
This fills the remaining background of guis with background with black.


DrMcCoy:
Merged as 9ce6c6f6f, thanks! :)

PR #383, opened 2018-11-16
COMMON: SerializationStream

Nostritius:
I'm currently rethinking my approach for the networking stuff. Part of this was to solve the problem of seperate serialization and deserialization. To overcome this problem i developed the SerializationStream class which has two child classes for reading and writing and depending on which one is used, it is possible that reading and writing share the same code.

### DIFF
--- a/src/common/rules.mk
+++ b/src/common/rules.mk
@@ -87,6 +87,7 @@ src_common_libcommon_la_SOURCES += \
     src/common/changeid.h \
     src/common/xml.h \
     src/common/rect.h \
+    src/common/serializationstream.h \
     $(EMPTY)
 
 src_common_libcommon_la_SOURCES += \
@@ -137,4 +138,5 @@ src_common_libcommon_la_SOURCES += \
     src/common/changeid.cpp \
     src/common/xml.cpp \
     src/common/rect.cpp \
+    src/common/serializationstream.cpp \
     $(EMPTY)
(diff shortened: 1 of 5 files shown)


DrMcCoy:
Just throwing a few questions at you to recap what the state of this was, sorry about that.

This feels like it duplicates a lot of our streams again? Feels a bit...repeater-y? But probably can't be helped, I guess.

What about pulling in some serialization from Boost? Or, say, protobuf? IIRC, you said they don't offer a solution to specify the endianness?

Is this to be used for our own network protocol? Or to reimplement the original network protocol?

How exactly would this be used? Can you give an example?

Nostritius:
The idea behind this was to offer a way to serialize and deserialize in one chunk of code. I consider it not a repetition of the stream class but rather an abstraction above them to enable writing or reading while using the same code. I have looked over both, boost serialization and protobuf, but the problem with these libraries nowadays is, that they are constructed assuming that the application doesn't have strict requirements, how the data is layed out as binary. They might be worth a look at if we are to create our own networking protocol (and, in the case of protobuf if you would allow precompilation steps with protoc ;)), but pretty much useless, if we want to recreate the original protocol a 100%. Another option, @nelak2 pointed me to, would be to use [kaitai](https://kaitai.io) which would be strict regarding the memory layout, but would also require a precompilation step.
An example for the usage could be:
```
uint32 id

SerializationStream ss = ...

ss.readOrWriteUint32BE(id);

switch (id) {
case kBNER:
    ...
    break;
}
```
Depending on how id an ss are intialized id will be written into a write stream or read from read stream and we don't need to double the code. The idea behind this was inspired by boost serializations & operator, but with more freedom, on how the data is laid out in binary form.

DrMcCoy:
Okay, merged it as 95db9720b...00d1a90d6, thanks! :)

PR #386, opened 2018-11-22
VIDEOS: Fix quicktime compilation without libfaad

Nostritius:
I recognized, that a compilation error was thrown with the latest revision, when building without libfaad. This should fix it.

### DIFF
--- a/src/video/quicktime.cpp
+++ b/src/video/quicktime.cpp
@@ -756,7 +756,11 @@ Sound::PacketizedAudioStream *QuickTimeDecoder::AudioSampleDesc::createAudioStre
 	case MKTAG('m', 'p', '4', 'a'):
 		switch (_objectTypeMP4) {
 		case 0x40:
+#ifdef ENABLE_FAAD
 			return Sound::makeAACStream(*_extraData);
+#else
+			return 0;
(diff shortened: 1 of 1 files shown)


DrMcCoy:
That commit is not really completely correct, though, is it?

I don't know much about Quicktime (that's @clone2727's forte :P), but it looks to me like it should rather go inside the 0x40 case, because the other cases might be something else.

There probably shouldn't be a `warning()` either, because that's not going to be triggered: `isAudioCodecSupported()` presumably already guards against `createAudioStream()` being called in that case. So it's only about removing the call to the removed function.

So I think it should rather be like this:

```c++
case MKTAG('m', 'p', '4', 'a'):
	switch (_objectTypeMP4) {
	case 0x40:
#ifdef ENABLE_FAAD
		return Sound::makeAACStream(*_extraData);
#else
		return 0;
#endif
	}
	break;
}
```

Does that make sense?

clone2727:
@DrMcCoy is correct on all counts.

Nostritius:
It should now be fixed

DrMcCoy:
Okay, merged it as 92f92f896, then, thanks! :)

PR #389, opened 2018-12-02
GRAPHICS: Add new flexible takeImage method

Nostritius:
This Method aims to be a more flexible approach for creating screenshots especially targeted towards the save game screenshot creation. Maybe it might even replace someday the takeScreenshot Method? It contains std::future and therefore needs C++11.

### DIFF
--- a/src/graphics/graphics.cpp
+++ b/src/graphics/graphics.cpp
@@ -106,6 +106,10 @@ GraphicsManager::GraphicsManager() : Events::Notifyable() {
 
 	_takeScreenshot = false;
 
+	_takeImage = false;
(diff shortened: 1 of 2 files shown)


DrMcCoy:
Hmm, what's the reason for using futures here? Does taking a screenshot really take so much time that we need to do that asynchronously?

Nostritius:
The Problem with the current architecture is that the renderer and the game logic are running in different threads. The current approach is good, if you want to take a screenshot and just throw it into a file and don't care from the point of view of the game. But this method was intended for creating screenshots which can be used for the save games, where you actually need the image inside the game logic.

PR #391, opened 2018-12-11
GRAPHICS: Add simple image resizing for Surface class

Nostritius:
This PR adds NN image resizing to the Surface class. It is loosely related #389, since the savegame pictures also need to be resized to a specific size.


Nostritius:
A doxygen comment is now added. And about the memory, I tested this code and also carefully thought about this. As far, as I'm aware, there shouldn't be any memory violations.

DrMcCoy:
Merged as dc5b4aa5e, thanks! :)

PR #392, opened 2018-12-11
GRAPHICS: Add the possibility to dump tga to stream

Nostritius:
Add the possibility to dump a surface to a tga writestream

### DIFF
--- a/src/graphics/images/dumptga.cpp
+++ b/src/graphics/images/dumptga.cpp
@@ -30,6 +30,7 @@
 #include "src/common/writefile.h"
 
 #include "src/graphics/images/decoder.h"
+#include "src/graphics/images/dumptga.h"
 
 namespace Graphics {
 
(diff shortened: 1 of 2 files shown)


src/graphics/images/dumptga.cpp
```diff
@@ -116,4 +116,26 @@ void dumpTGA(const Common::UString &fileName, const ImageDecoder *image) {
 	file->flush();
 }
 
+void dumpTGA(Common::WriteStream &stream, const ImageDecoder *image) {
```
DrMcCoy: That function duplicates an awful lot of code. :/

I'm opening the actual file pretty late in the original `dumpTGA()`, to decrease the possibility of leaving a blank file when an early check throws, but that's not really essential. I'd rather have a function that writes to a stream, and a small wrapper function that opens a file and then calls that other function.

Nostritius:
I have now changed the whole thing to use the stream function as base and the file function calls the stream function.

DrMcCoy:
Merged as 155fbb270, thanks! :)

PR #394, opened 2018-12-23
BUILD: Add cmake vpx support

Nostritius:
After the support for VP8/VP9 videos was added, cmake build support was missing. I added this.


DrMcCoy:
Merged as 97ea507f3...7c829ad46, thanks! :)

PR #398, opened 2018-12-29
AURORA: Enable extraction of bitmaps from pe files

Nostritius:
This PR enables to extract bitmaps from pe exe files. It is thought as base for potentially implementing the spalash screens of dragonage and dragonage2.

### DIFF
--- a/src/aurora/pefile.cpp
+++ b/src/aurora/pefile.cpp
@@ -48,57 +48,86 @@ const Archive::ResourceList &PEFile::getResources() const {
 }
 
 Common::SeekableReadStream *PEFile::getResource(uint32 index, bool UNUSED(tryNoCopy)) const {
-	// Convert from the PE cursor group/cursor format to the standalone
-	// cursor format.
+	ResourceList::const_iterator iter = _resources.begin();
+	if (index >= _resources.size())
+		throw Common::Exception("Resource index out of range (%u/%u)", index, (uint)_resources.size());
 
-	Common::ScopedPtr<Common::SeekableReadStream>
-		cursorGroup(_peFile->getResource(Common::kPEGroupCursor, index));
+	std::advance(iter, index);
 
-	if (!cursorGroup)
-		throw Common::Exception("No such PE resource %u", index);
+	switch (iter->type) {
+		case kFileTypeBMP: {
+			Common::ScopedPtr<Common::SeekableReadStream> stream(_peFile->getResource(Common::kPEBitmap, _peIDs.at(index)));
+
+			Common::MemoryWriteStreamDynamic bmp(true);
+			bmp.writeByte('B');
+			bmp.writeByte('M');
+
+			bmp.writeUint32LE(stream->size() + 14);
+			bmp.writeZeros(4);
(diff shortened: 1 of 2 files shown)


Nostritius:
I have now rewritten the code to use another method for determining the type.

Nostritius:

src/aurora/pefile.cpp
```diff
@@ -48,57 +48,86 @@ const Archive::ResourceList &PEFile::getResources() const {
 }
 
 Common::SeekableReadStream *PEFile::getResource(uint32 index, bool UNUSED(tryNoCopy)) const {
-	// Convert from the PE cursor group/cursor format to the standalone
-	// cursor format.
+	ResourceList::const_iterator iter;
+	for (iter = _resources.begin(); iter != _resources.end(); ++iter) {
+		if (iter->index == index)
+			break;
+	}
```
DrMcCoy: - If index isn't there in _resources, this will leave iter dangling and invoke UB
- Also, index is == the PE resource ID, no? Then you again have the issue that multiple IDs, but different types will clash

src/aurora/pefile.cpp
```diff
@@ -121,6 +150,23 @@ void PEFile::load(const std::vector<Common::UString> &remap) {
 
 		_resources.push_back(res);
 	}
+
+	std::vector<Common::PEResourceID> bitmapList = _peFile->getNameList(Common::kPEBitmap);
+
+	for (std::vector<Common::PEResourceID>::const_iterator it = bitmapList.begin(); it != bitmapList.end(); ++it) {
+		Resource res;
+
+		uint32 id = it->getID() - 1;
```
DrMcCoy: If ID is 0, this will wrap. Why subtract 1 in the first place?

Nostritius:
I replaced now the pe id with a unique id for the archive and added a map which maps the unique id to the pe id. This should hopefully fix the problem of potential multiple ids with different types. I also fixed the other issues.

DrMcCoy:
The IDs you give them are sequential, no? Why does it need to be a map, then; shouldn't a vector be enough?

Nostritius:
You're right, I will change this

Nostritius:
This should now be done

src/aurora/pefile.cpp
```diff
@@ -48,57 +48,86 @@ const Archive::ResourceList &PEFile::getResources() const {
 }
 
 Common::SeekableReadStream *PEFile::getResource(uint32 index, bool UNUSED(tryNoCopy)) const {
-	// Convert from the PE cursor group/cursor format to the standalone
-	// cursor format.
+	ResourceList::const_iterator iter = _resources.begin();
+	if (index >= _resources.size())
+		return 0;
```
DrMcCoy: This should probably even throw.

Compare with what the other archives do in this case: https://github.com/xoreos/xoreos/blob/master/src/aurora/rimfile.cpp#L97 , for example

Nostritius:
I have now replaced the return 0 through a throw.

DrMcCoy:
This looks good now, merged, thanks! :)

PR #400, opened 2018-12-30
TESTS: Add unit tests for Engines::NWN2::Waypoint and Engines::NWN2::Creatures classes

rjshae:
This is a prototype for unit testing the NWN2 classes using data generated in the NWN2 Toolset. The header files were generated from saved .git files that were converted to byte arrays using a Python script.


DrMcCoy:
I think you somehow messed up your branch's history.

The easiest way, I think, would be to create a new branch, from the current, up-to-date xoreos master branch, and then cherry-pick the new commits on top.

rjshae:
Yes, I was thinking of doing just that. Thanks.

rjshae:
I'm not sure how to interpret that AppVeyor failure.

DrMcCoy:
Hmm :/

A quick googling suggests that 0xC0000135 is the WinAPI error code STATUS_DLL_NOT_FOUND, so I'm assuming it can't find the DLLs when executing the tests?

Interestingly enough, tests_engines_test_trigger works, though, and that pulls in all the same libraries on my Linux box. But it's possible that MSVC strips some unneeded DLLs.

@berenm, I don't suppose you know if all the libraries we're pulling in are installed in locations that would be found when executing the compiled binaries?

rjshae:
I tried removing various libraries from the nwn2_LIBS declaration, but I get link errors. They all appear to be needed for the build.

berenm:
I would say that probably the third-party .dll aren't found by default and would need to be copied alongside the test binaries if anything need to run on the AppVeyor servers. Until now, it was only compiling, and for this the .dll were not actually needed.

DrMcCoy:
Okay, I found out that I can just installed the dependency walker on AppVeyor, and use that to pin-point which DLL is missing. Turns out it was xvidcore.dll

Fixed this with a1c1a269b, and merged the PR as a1c1a269b...c313e23c6. Thanks! :)

rjshae:
Thanks!

PR #403, opened 2019-01-14
KOTOR: Fix the carth dialog

Nostritius:
After a lot of research i found out, why the carth dialog was not triggered properly. It was because the trigger scripts need the actual trigger object as owner and not the area. I fixed this in this commit. I verified this behaviour by decompiling the trigger script and adding the following line:
`if (GetTag(OBJECT_SELF) == "end_trig02")
			PlayMovie("01A");`
It behaved as I expected and played the movie.


DrMcCoy:
Oh, nice! Merged as ce8b6b558, thanks! :)

PR #405, opened 2019-01-17
SOUND: Fix vorbis usage when vorbis is not enabled

Nostritius:
A small fix to compile when vorbis is disabled.


DrMcCoy:
Merged, thanks! :)

PR #406, opened 2019-01-19
KOTOR: Skills and Abilities

Nostritius:
This PR introduces Skills and abilities for creatures as preparation for the security check to open the second door on the endar spire.


DrMcCoy:
Merged as 70446dc8c...4b5128c1f, thanks! :)

PR #407, opened 2019-01-19
KOTOR: Fix parameter index for some nws functions

Nostritius:
A small bug i recognized.


DrMcCoy:
Oh, good catch. Merged as 70446dc8c, thanks! :)

PR #408, opened 2019-01-24
SOUND: FEV Update

Nostritius:
This PR adds several new load features for the FEV loading. Also it  moves the FEVFile class into the sound packade.

### DIFF
--- a/src/sound/fevfile.cpp
+++ b/src/sound/fevfile.cpp
@@ -245,7 +245,21 @@ void FEVFile::readEvent(Common::SeekableReadStream &fev) {
 
 	event.positionRandomization3D = fev.readUint32LE();
 
-	fev.skip(8);
+	uint32 numLayers = fev.readUint32LE();
+	event.layers.resize(numLayers);
+	for (uint32 i = 0; i < numLayers; ++i) {
+		EventLayer layer;
+
+		fev.skip(2);
+
+		layer.priority = fev.readSint16LE();
+
+		fev.skip(6);
+
+		event.layers[i] = layer;
+	}
+
+	fev.skip(4);
 
 	event.userProperties = readProperties(fev);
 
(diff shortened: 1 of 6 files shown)


src/xoreos.cpp
```diff
@@ -28,6 +28,7 @@
 #include <cstdio>
 
 #include <vector>
+#include <src/sound/fevfile.h>
```
DrMcCoy: (The include format is wrong here...)

But this whole block is testing code you apparently accidentally left in :P

src/sound/fevfile.cpp
```diff
@@ -182,9 +182,32 @@ void FEVFile::readEvent(Common::SeekableReadStream &fev) {
 	event.volumeRandomization = fev.readIEEEFloatLE();
 	event.priority = fev.readUint32LE();
 	event.maxPlaybacks = fev.readUint32LE();
-	event.mode = fev.readUint32LE();
 
-	fev.skip(12);
+	uint32 mode = fev.readUint32BE();
```
DrMcCoy: (There's duplicate stuff in the commit message here)

Nostritius:

DrMcCoy:
Hmm, can you rename class FEVFile to FMODEvents (I guess), to be more in line with FMODSampleBank (and the rest of the similar sound classes?). Likewise, the files should then be fmodevents.h and fmodevents.cpp.

Otherwise, this looks good :)

Nostritius:
Hmm, it's primary function is (at the moment) still a file loader, so what about FMODEventFile?

DrMcCoy:
Sounds good to me, yes :)

Nostritius:
The file is now renamed

DrMcCoy:
Merged as 70fd28a44...8bda31121, thanks! :)

PR #409, opened 2019-01-26
KOTOR: Selection

Nostritius:
This PR introduces the selection circles and the object information. It tries to mirror the behaviour of the original game and is my second try after #243. It is not yet complete and should only show the name and the health of the object when hovering over it.


Nostritius:
The crash should now be fixed

DrMcCoy:
Merged as 536c65de8...469a23871, thanks! :)

PR #414, opened 2019-02-03
AURORA: Fix PEFile cursor remap

Nostritius:
A small fix which fixes the remapping of cursors, which led to wrong cursors being shown in kotor.

### DIFF
--- a/src/aurora/pefile.cpp
+++ b/src/aurora/pefile.cpp
@@ -140,10 +140,10 @@ void PEFile::load(const std::vector<Common::UString> &remap) {
 			throw Common::Exception("Found non-integer cursor group");
 
 		uint32 id = it->getID();
-		if (id >= remap.size())
+		if (id >= remap.size() || id <= 0)
 			res.name = Common::UString::format("cursor%d", id);
 		else
-			res.name = remap[id];
+			res.name = remap[id - 1];
 
 		res.type  = kFileTypeCUR;
 		res.index = _peIDs.size();


src/aurora/pefile.cpp
```diff
@@ -143,7 +143,7 @@ void PEFile::load(const std::vector<Common::UString> &remap) {
 		if (id >= remap.size())
 			res.name = Common::UString::format("cursor%d", id);
 		else
-			res.name = remap[id];
+			res.name = remap[id - 1];
```
DrMcCoy: Unfortunately, that now means that an ID of 0 would wrap here.

I think it would probably be best not to do any arithmetics on the ID, and simply make the remap vectors map directly by adding a dummy "" value at index 0 in the engine code for KotOR and KotOR2.

Or am I overlooking something here? @clone2727, you and me wrote the remapping in the first place like ages ago, no? Am I forgetting something? :P

clone2727: It broke in bb44be0. It previously always subtracted 1.

Yeah, could just put a dummy value at index 0.

DrMcCoy: Yes, I think it breaking is my fault. I explictly called that out in the review because I didn't even remember.

Nostritius:
I have now added a second condition, which should catch the case id == 0.

DrMcCoy:
I still think changing the mapping to be 1:1 would be cleaner, but this works too. Thanks, merged :)

PR #416, opened 2019-02-05
NWN2: Add GetGameDifficulty() script function

rjshae:
This will remove a warning during startup about a TODO function. The code sets
the default difficulty value to '2' for the GAME_DIFFICULTY_NORMAL mode and
checks that the difficulty value is in the range 0 (GAME_DIFFICULTY_VERY EASY)
to 4 (GAME_DIFFICULTY_DIFFICULT).


DrMcCoy:
Merged, thanks! :)

PR #417, opened 2019-02-05
GRAPHICS: Fix environment map rendering

mirv-sillyfish:
Fix environment maps (under and over) not being rendered.


DrMcCoy:
Merged, thanks! :)

PR #418, opened 2019-02-06
GRAPHICS: force add duplicate mesh names

mirv-sillyfish:
When a duplicate mesh name is detected, force add the mesh by
changing its name to something unique.


DrMcCoy:
Okay, merged as c3f03255f11a0b3914c873fcd0ee8a38cc28fdc1, thanks! :)

PR #420, opened 2019-02-07
DRAGONAGE: Fix startup on Windows

seedhartha:
Caused by [this](https://github.com/xoreos/xoreos/blob/2f8b9a66e4e274b2735d46804b3e64b03bcc5e7e/src/aurora/resman.cpp#L276) line, which behaves differently on Windows and Linux.

### DIFF
--- a/src/engines/dragonage/dragonage.cpp
+++ b/src/engines/dragonage/dragonage.cpp
@@ -217,28 +217,28 @@ void DragonAgeEngine::initResources(LoadProgress &progress) {
 	progress.step("Adding core archive directories");
 
 	progress.step("Indexing core resources files");
-	Game::loadResources ("/packages/core", 0, _resources);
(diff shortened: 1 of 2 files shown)


DrMcCoy:
Hmm, damn. :/

In case it's not obvious (which it probably isn't), the idea behind is that paths starting with a / should only match paths that start in the root of the game data files directory, not all subdirectories.

For example, for a game in /data/game/blah/, when looking for
- "foo/bar", it also matches /data/game/blah/quux/foo/bar in addition to /data/game/blah/foo/bar
- "/foo/bar" only matches /data/game/blah/foo/bar and not /data/game/blah/quux/foo/bar

In the case of Dragon Age, I just don't want DLCs, mods, etc. to accidentally clash with the stock original campaign files.

That API design itself is probably not really all that great, yeah. Not sure what to do about it.

seedhartha:
Okay, created issue #421 instead.

PR #422, opened 2019-02-08
WIP: Meta engine for KotOR games

seedhartha:
This PR introduces foundation of meta engine for KotOR games to reduce code duplication. I also took the liberty to cleanup and reorganize some of the files.


DrMcCoy:
Neat :)

But, hmm, I'm not so sure that, organizationally, this KotOR meta-engine belongs into src/engines/kotor/meta/. I would have put it into src/engines/kotormeta/ (or some other name?), with both src/engines/kotor/ and src/engines/kotor2/ depending on it.

Or maybe name it src/engines/kotor/ for the meta-engine, and src/engines/kotor1/ and src/engines/kotor2/ the games themselves? (or src/engines/kotor0 for the meta-engine?)

Or maybe src/engines/kotor/ the meta-engine, and src/engines/kotor/kotor1/ src/engines/kotor/kotor2/ the sub-engines?

DrMcCoy:
(I do like the general idea of having a meta-engine base that KotOR and KotOR2 branch off of. That fits with how close the engines are, and that was something I wanted for a while. Same with Dragon Age: Origins and Dragon Age II, probably.)

seedhartha:
Yes, I started with a separate "kotormeta" engine folder, but changed it to "meta" subdirectory later on. The reason is that I don't like having an abundance of engine folders and compile options (they are confusing). I also didn't want to rename the KotOR folder to not clutter the commit history or interfere with other peoples' work. Personally, this is the best directory structure I can think of:

/src/engines/kotor/meta
/src/engines/kotor/kotor1
/src/engines/kotor/kotor2

clone2727:
"Meta" isn't the right word to use here. That implies that it's something about the engine rather than an engine or part of one. "Base" is more appropriate.

seedhartha:
I renamed "meta" to "base" engine and also limited the changes to ScriptContainer, Room and Object base classes as well as a shared types.h file. I think it's mergeable now.

src/engines/kotor/base/object.h
```diff
@@ -19,150 +19,126 @@
...
 	                           float length = 0.0f,
 	                           float speed = 1.0f);
 
 protected:
 	ObjectType _type; ///< The object's type.
-
-	Common::UString _name;        ///< The object's display name.
+	Common::UString _name; ///< The object's display name.
 	Common::UString _description; ///< The object's description.
-
```
DrMcCoy: Can you keep the previous formatting style of the file, instead of removing all the grouping newlines?

seedhartha: Okay. What about the grouping of methods though? I find it much easier to navigate class declarations when their mehods are grouped by type (getters, setters, actors, event handlers, etc.) rather then functionality.

DrMcCoy: I find it more useful to group them according to functionality.

seedhartha:
Moved the base engine from "src/engines/kotor/base" to "src/engines/kotorbase" and reverted formatting of object.h. I think it's best not to create a separate namespace at this point since it will require changes to a lot of files and those files will be part of the base engine eventually anyway.

P.S. KotOR / Jade Empire GUI should probably be part of the base engine as well.

Nostritius:
Maybe in this case it could be called "Odyssey"? Since all three engines have this engine in common. Then you could make a namespace like Engines::Odyssey::.

seedhartha:
Good catch! Then again, it only applies to the GUI system. There will be a lot of KotOR-specific logic in the base engine.

DrMcCoy:
Okay, merged as b7f08abbc...b1fe3a6e6, thanks! :)

PR #423, opened 2019-02-10
NWN2: Implement Roster-related script calls

rjshae:
This change will implement the following script calls: AddRosterMemberByTemplate, GetIsRosterMemberAvailable, GetIsRosterMemberCampaignNPC, and GetIsRosterMemberSelectable. The first eliminates multiple startup warnings about a missing script function.


src/engines/nwn2/roster.cpp
```diff
@@ -82,11 +105,25 @@ void Roster::load() {
...
-	member.rosterName = gff.getString("RosName");
+	member.rosterName = gff.getString("RosName", member.rosterName);
+	member.isAvailable = gff.getBool("RosAvailable", member.isAvailable);
+	member.isCampaignNPC = gff.getBool("RosCampaignNPC", member.isCampaignNPC);
+	member.isSelectable = gff.getBool("RosSelectable", member.isSelectable);
+	member.isLoadBefore = gff.getBool("RosLoadBefore", member.isLoadBefore);
 	_members.push_back(member);
 }
 
+void Roster::initMember(Member *member) {
```
DrMcCoy: Make it take a `Member &` instead, please.

Or you could even add a constructor to struct Member instead.

tests/engines/nwn2/roster.cpp
```diff
@@ -76,3 +76,18 @@ GTEST_TEST(NWN2Roster, rosterNames) {
...
+	Common::ScopedPtr<TestRoster> roster(new TestRoster);
+
+	EXPECT_TRUE(roster->getIsRosterMemberAvailable("ammon_jerro"));
+	EXPECT_FALSE(roster->getIsRosterMemberAvailable("khelgar"));
+	EXPECT_TRUE(roster->getIsRosterMemberCampaignNPC("qara"));
+	EXPECT_FALSE(roster->getIsRosterMemberCampaignNPC("neeshka"));
+	EXPECT_TRUE(roster->getIsRosterMemberSelectable("shandra"));
+	EXPECT_FALSE(roster->getIsRosterMemberSelectable("npc_bevil"));
+}
+
```
DrMcCoy: Extra empty line at the end of the file

src/engines/nwn2/roster.h
```diff
@@ -47,12 +47,21 @@ class Roster {
 	/** Increment to the next member in the list. */
 	Common::UString getNextRosterMember();
 
+	/** Fetch roster member information. */
+	bool getIsRosterMemberAvailable(Common::UString name) const;
```
DrMcCoy: Those name parameters can all be `const Common::UString &`

src/engines/nwn2/roster.cpp
```diff
@@ -65,6 +65,29 @@ Common::UString Roster::getNextRosterMember() {
...
+	std::list<Member>::const_iterator it = getRosterMember(name);
+	return (it != _members.end()) ? it->isCampaignNPC : false;
+}
+
+bool Roster::getIsRosterMemberSelectable(Common::UString name) const {
+	std::list<Member>::const_iterator it = getRosterMember(name);
+	return (it != _members.end()) ? it->isSelectable : false;
+}
+
+std::list<Roster::Member>::const_iterator Roster::getRosterMember(Common::UString name) const {
```
DrMcCoy: xoreos has been opened for C++11, so you could theoretically use std::find() (in `<algorithm>`) with a lambda here. It's not necessary, though.

tests/engines/nwn2/roster.cpp
```diff
@@ -76,3 +76,18 @@ GTEST_TEST(NWN2Roster, rosterNames) {
 	EXPECT_STREQ(roster->getFirstRosterMember().c_str(), "ammon_jerro");
 	EXPECT_STREQ(roster->getNextRosterMember().c_str(),  "bishop");
 }
+
+/**
+ * Test the roster member settings in the ROSTER.rst file data.
+ */
+GTEST_TEST(NWN2Roster, rosterSettings) {
+	Common::ScopedPtr<TestRoster> roster(new TestRoster);
```
DrMcCoy: Likewise, you can use std::unique_ptr here now. In fact, I'd prefer it. I'd like to migrate away from our ScopedPtr a bit.

src/engines/nwn2/roster.cpp
rjshae: Changed to a constructor.

src/engines/nwn2/roster.cpp
rjshae: I'll have to revisit this when I do the 'set' calls, so I can experiment then.

src/engines/nwn2/roster.h
```diff
@@ -57,11 +60,12 @@ class Roster {
 
 private:
 	struct Member {
-		Member() : rosterName(""),
+		Member() : rosterName(""),      cTemplate(""),
```
DrMcCoy: Common::UString already constructs to empty, so you don't really need to do this here as well. (But it doesn't break anything either).

For another C++11-ism, you could inline brace-initializers for these member variables instead of doing that in a constructor. I.e. instead of
```
		bool isAvailable;           ///< Not currently in a party?
```
do
```
		bool isAvailable { false };           ///< Not currently in a party?
```

Again, not strictly necessary.

rjshae: Understood. I think I'll leave it there though so it's known it was intentional. Thanks.

DrMcCoy:
Merged as b68e3b8f3...dc6d13a05, thanks! :)

PR #425, opened 2019-02-10
GRAPHICS: GuiQuad scaling and blending

Nostritius:
This PR adds the capability to scale and multiply blend guiquads. This is part of my work for the scaleform renderer, since some elements need to be scaled or differently blended.


DrMcCoy:
Merged as dc6d13a05...da49d01f6, thanks! :)

PR #427, opened 2019-02-11
NWN2: Additional roster functonality

rjshae:
This revision will allow roster member settings to be changed, control the
party limit on NPC companions selectable from the UI, and retrieve party
information from the roster file.


src/engines/nwn2/roster.cpp
```diff
@@ -34,29 +36,44 @@ namespace Engines {
...
+ * because the UTC files will not be available.
+ */
+bool Roster::addRosterMemberByTemplate(Common::UString name, Common::UString cTemplate, bool unitTest) {
 	// Disallow null strings
 	if (name.size() == 0 || cTemplate.size() == 0)
 		return false;
 
 	// Don't permit duplicate roster names
-	if (getRosterMember(name) != _members.end())
+	auto matchName = [&name](const Member &m) bool {return m.rosterName == name;};
```
DrMcCoy: Return type is done with `->`. See (2) here: https://en.cppreference.com/w/cpp/language/lambda

src/engines/nwn2/roster.cpp
```diff
@@ -34,29 +36,44 @@ namespace Engines {
...
+	std::list<Member>::const_iterator it = std::find_if(_members.begin(), _members.end(), matchName);
```
DrMcCoy: However, I'd have done it like

```cpp
auto member = std::find_if(_members.begin(), _members.end(), [&](const Member &m) {
	return m.rosterName == name;
});
```

src/engines/nwn2/roster.cpp
```diff
@@ -34,29 +36,44 @@ namespace Engines {
...
+void Roster::setRosterNPCPartyLimit(uint32 limit) {
+	_partyLimit = limit;
+}
+
+/**
+ * Add a creature by template name to the roster list.
+ * The 'unitTest' flag is required for unit testing
+ * because the UTC files will not be available.
+ */
+bool Roster::addRosterMemberByTemplate(Common::UString name, Common::UString cTemplate, bool unitTest) {
```
DrMcCoy: I'm not really all that fond of that unitTest param :/

rjshae: Is there an alternative other than to not run the test? Like a unit test state function I can call?

DrMcCoy: Not really, no.

I can't think of a better solution. We can't mock the ResourceManager either.
Maybe some way to tell the ResourceManager to lie? An anti-blacklist of sorts?

Probably a discussion for the future. Let's keep that param for now, I guess. Please mark it with a "HACK" comment.

rjshae: Okay, I just had a thought to just use a special token in the name field, such as "[GTEST]". If the ResMan lookup fails, then it can check for the token. Thanks.

DrMcCoy:
Merged as 42926c25a...b7f08abbc, thanks! :)

PR #430, opened 2019-02-12
NWN2: Prep for implementing SetListeningPattern call

rjshae:
The listening patterns are primarily used by the creature AI, but I
believe they can be used by other object types so I implemented
this at the object class level.


rjshae:
I'm planning to use std::map to store the listening patterns for quick lookup.
Does that make sense?

rjshae:
Since the AppVeyor test is hosed anyway, I went ahead and implemented the SetListeningPattern call as part of this pull request. Hope that's okay.

DrMcCoy:
Merged as b1fe3a6e6...2e8dc08a7, thanks! :)

PR #431, opened 2019-02-13
Odyssey engine

seedhartha:
This PR creates the Odyssey engine from what used to be the base GUI for KotOR and Jade Empire. I also refactored the list box widget to allow creation of custom item widgets through the use of factory method.


rjshae:
Looks like you hit the same problem I did with the AppVeyor network connection.

seedhartha:
Yeah, it's been there for a while :)

DrMcCoy:
Merged as 72918b5f0, thanks! :)

(And nasm.us seems to be up again, so AppVeyor builds run through again. At least for now :P)

PR #432, opened 2019-02-13
GRAPHICS: Shader based debug console rendering

mirv-sillyfish:
Fix console and ttfont to use shader based rendering if the
experimental renderer is enabled.

### DIFF
--- a/src/engines/aurora/console.cpp
+++ b/src/engines/aurora/console.cpp
@@ -28,6 +28,8 @@
 
 #include <boost/bind.hpp>
 
+#include "glm/gtc/matrix_transform.hpp"
+
 #include "src/common/util.h"
 #include "src/common/strutil.h"
 #include "src/common/filepath.h"
@@ -56,6 +58,10 @@
 
 #include "src/engines/aurora/console.h"
 #include "src/engines/aurora/util.h"
+
+#include "src/graphics/mesh/meshman.h"
+#include "src/graphics/shader/surfaceman.h"
+#include "src/graphics/shader/materialman.h"
 
 
 static const uint32 kDoubleClickTime = 500;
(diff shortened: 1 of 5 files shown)


src/engines/aurora/console.cpp
```diff
@@ -57,6 +57,11 @@
 #include "src/engines/aurora/console.h"
 #include "src/engines/aurora/util.h"
 
+#include "src/graphics/mesh/meshman.h"
+#include "src/graphics/shader/surfaceman.h"
+#include "src/graphics/shader/materialman.h"
+#include "glm/gtc/matrix_transform.hpp"
```
DrMcCoy: Could you move the glm include between the boost and common includes, please? Separated with a new line in both directions.

src/graphics/aurora/ttffont.cpp
```diff
@@ -173,6 +186,75 @@ void TTFFont::buildChars(const Common::UString &str) {
...
 
+void TTFFont::renderBind(const glm::mat4 &transform) const {
+	glUseProgram(_renderable->getProgram()->glid);
+	_material->bindProgram(_renderable->getProgram(), 1.0f);
+	_material->bindGLState();
+	_renderable->getSurface()->bindProgram(_renderable->getProgram(), &transform);
+	_renderable->getSurface()->bindGLState();
+	_mesh->renderBind();
+
+	/**
```
DrMcCoy: Since that's supposed to be a "normal" multi-line comment, and not a Doxygen comment, change that into just plain `/*`, please.

src/graphics/aurora/ttffont.cpp
```diff
@@ -173,6 +186,75 @@ void TTFFont::buildChars(const Common::UString &str) {
...
+
+
+	float v_pos[12];
+	float v_uv[8];
+	float v_rgba[4*4];
+
+	size_t page = cC->second.page;
+	assert(page < _pages.size());
+
+	/**
```
DrMcCoy: Same here

src/graphics/aurora/ttffont.h
```diff
@@ -60,6 +64,14 @@ class TTFFont : public Graphics::Font {
 
 	void buildChars(const Common::UString &str);
 
+	/**
+	 * @brief Bind the font for rendering. Must be performed before render is called.
+	 * @param transform  Base modelview transform. Under most circumstances this is expected to be the identity matrix.
+	 */
```
DrMcCoy: The format we're using should be

```cpp
	/** Bind the font for rendering. Must be performed before render is called.
	 *
	 *  @param transform  Base modelview transform. Under most circumstances
	 *                    this is expected to be the identity matrix.
	 */
```

DrMcCoy:
Merged as 3e2d26203, thanks! :)

PR #434, opened 2019-02-16
ENGINES: Create base creature for KotOR games

seedhartha:
This PR creates base creature and support classes for KotOR games.


DrMcCoy:
Merged as 8f92feccb, thanks! :)

PR #435, opened 2019-02-16
NWN2: Lock, key, and hit point functionality for situated objects

rjshae:
I tried to add a unit test for doors, but it aborts for an unknown reason so I left it out.
(I suspect it fails because of the dependency of the Door class on the NWN2 engine.)


DrMcCoy:
Merged as 3e2d26203...6a8aa616d, thanks! :)

PR #437, opened 2019-02-18
GRAPHICS: Fix texture index 2 assignment to material

mirv-sillyfish:
Properly assign texture index 2 to a generated ShaderMaterial, if
that index will be required.


DrMcCoy:
Merged, thanks! :)

PR #438, opened 2019-02-19
KOTORBASE: Modify enumeration of inventory slots to match nwscript.nss

seedhartha:


src/engines/kotorbase/item.cpp
```diff
@@ -60,8 +60,8 @@ const Common::UString &Item::getName() const {
 	return _name;
 }
 
-EquipmentSlot Item::getEquipableSlots() const {
-	return _equipableSlots;
+bool Item::isSlotEquipable(InventorySlot slot) const {
```
DrMcCoy: You should probably check for slot being < 32 here and explictly return false otherwise, because shifting >= the width of the type is Undefined Behaviour

seedhartha: Done.

DrMcCoy:
Merged, thanks! :)

PR #439, opened 2019-02-19
More base objects for KotOR games

seedhartha:
This PR extracts the following objects: situated, placeable, door, trigger, waypoint, sound and pathfinding classes.


DrMcCoy:
Merged, thanks! :)

PR #440, opened 2019-02-20
ACTIONSCRIPT: Update

Nostritius:
Another actionscript update with many more changes

### DIFF
--- a/src/aurora/actionscript/asbuffer.cpp
+++ b/src/aurora/actionscript/asbuffer.cpp
@@ -332,7 +332,7 @@ void ASBuffer::actionAdd2() {
 	Variable b = _stack.top();
 	_stack.pop();
 
-	_stack.push(a + b);
+	_stack.push(b + a);
 
 	debugC(kDebugActionScript, 1, "actionAdd2");
 }
--- a/src/aurora/actionscript/asbuffer.cpp
+++ b/src/aurora/actionscript/asbuffer.cpp
@@ -480,9 +480,15 @@ void ASBuffer::actionCallMethod(AVM &avm) {
 		counter += 1;
 	}
 
-	for (size_t i = 0; i < arguments.size(); ++i) {
-		avm.storeRegister(arguments[i], counter);
-		counter += 1;
+	if (function->hasRegisterIds()) {
+		for (size_t i = 0; i < arguments.size(); ++i) {
+			avm.storeRegister(arguments[i], function->getRegisterId(i));
+		}
+	} else {
+		for (size_t i = 0; i < arguments.size(); ++i) {
+			avm.storeRegister(arguments[i], counter);
+			counter += 1;
+		}
 	}
 
 	avm.setReturnValue(Variable());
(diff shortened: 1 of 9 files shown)


src/aurora/actionscript/avm.cpp
```diff
@@ -41,6 +41,9 @@ namespace ActionScript {
 
 AVM::AVM() {
 	_registers.resize(256);
+	for (int i = 0; i < 256; ++i)
```
DrMcCoy: Please use _registers.size() here (and size_t i), or iterators, or a range-based for, so that we're not repeating the 256 here.

src/aurora/actionscript/asbuffer.cpp
```diff
@@ -464,6 +464,8 @@ void ASBuffer::actionCallMethod(AVM &avm) {
 	byte counter = 1;
 	Variable prevThis;
 
+	avm.pushRegisters(function->getNumRegisters());
```
DrMcCoy: getNumRegisters() returns an unsigned int, but pushRegisters takes a byte?

src/aurora/actionscript/object.cpp
```diff
@@ -53,6 +53,9 @@ std::vector<Common::UString> Object::getSlots() const {
 }
 
 bool Object::hasMember(const Common::UString &id) {
+	if (_members.find("constructor") != _members.end())
+		if (_members["constructor"].asObject()->hasMember(id))
```
DrMcCoy: find() returns an iterator you can use directly, instead of having to go through operator[] again

src/aurora/actionscript/object.cpp
```diff
@@ -64,7 +64,10 @@ Variable Object::getMember(const Variable &id) {
 		throw Common::Exception("Object::getMember id is not a string");
 
 	const Common::UString idString = id.asString();
-	if (_members.find(idString) != _members.end())
+
+	if (hasMember(idString) && hasMember("constructor") && _members["constructor"].asObject()->hasMember(idString))
+		return _members["constructor"].asObject()->getMember(id);
+	else if (_members.find(idString) != _members.end())
```
DrMcCoy: Same here

Nostritius:

src/aurora/actionscript/object.cpp
```diff
@@ -65,7 +65,11 @@ Variable Object::getMember(const Variable &id) {
 		throw Common::Exception("Object::getMember id is not a string");
 
 	const Common::UString idString = id.asString();
-	if (_members.find(idString) != _members.end())
+
```
DrMcCoy: Likewise, `getMember()` could/should be const, no?

Nostritius: getMember might create new object, so i think it should stoy non const.

Nostritius:
The issues except for getMember should now be fixed

DrMcCoy:
Merged as c63240c19...66f8ca7a3, thanks! :)

PR #441, opened 2019-02-20
NWN: Display dialog boxes with experimental renderer

mirv-sillyfish:
Add support for displaying dialog boxes when using shader based
rendering.


DrMcCoy:
Merged as 74da9bbe1, thanks! :)

PR #442, opened 2019-02-21
NWN2: Change cursor appropriately during mouse-over of a creature

rjshae:


src/engines/nwn2/cursor.h
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+#ifndef ENGINES_NWN2_CURSOR_H
+#define ENGINES_NWN2_CURSOR_H
+
+namespace Engines {
+
+namespace NWN2 {
+
+/**
```
DrMcCoy: This is not a Doxygen comment, so please use just plain `/*` here.

src/engines/nwn2/object.cpp
```diff
@@ -300,7 +303,7 @@ uint32 Object::getFaction() const {
 /** Return this object's reputation with the source. */
 uint8 Object::getReputation(Object *source) const {
 	assert(source);
-	return 0; // Default to 'enemy'
+	return getArea()->getModule().getFactions().getReputation(source, getFaction());
```
DrMcCoy: getArea() might return 0, so please check before dereferencing

rjshae:
Changes applied. Thanks.

DrMcCoy:
Merged as c81b917b0...08c5cf7fa, thanks! :)

PR #443, opened 2019-02-21
NWN: legal billboard fade with shader rendering

mirv-sillyfish:
Properly update fade alpha when rendering legal billboard with
shader render path.


DrMcCoy:
Merged, thanks! :)

PR #444, opened 2019-02-21
NWN: Translucent ingame dialog with shader rendering

mirv-sillyfish:
Make dialog backdrop translucent with shader rendering.


DrMcCoy:
Merged, thanks! :)

PR #447, opened 2019-02-22
GRAPHICS: GLM takes rotation in radians

mirv-sillyfish:
Correctly apply deg2rad for render transform calculation when using
GLM and shader based rendering.


DrMcCoy:
Merged, thanks! :)

PR #448, opened 2019-02-23
KOTOR: Fix random crash when selecting a class in character generation

seedhartha:


DrMcCoy:
Merged as 8c482956d, thanks! :)

PR #450, opened 2019-02-23
Refactor CharacterGenerationInfo in KotOR games

seedhartha:
This PR extracts creature and model creation from the CharacterGenerationInfo class, which removes a circular dependency on the Creature class and, IMO, better represents the class intent.


DrMcCoy:
Hmm, but the classes in KotOR and KotOR2 are different. So now you have KotOR-only stuff in KotORBase

seedhartha:
True. Character and CharacterGenerationInfo classes should be subclassed in KotOR as well.

seedhartha:
I have extracted KotOR-specific code for creatures, chargeninfo and saved games from the base engine. As for the issues you were having, they seem to have existed for some time before the base engine PR. It can wait for a separate PR, I think.

DrMcCoy:
(Another one, when selecting a character class (a female soldier, I think it was): https://gist.github.com/DrMcCoy/78899f951dd27dd768d87fb7b7e7daf9 )

DrMcCoy:
Ah, nope, that one now always happens with this PR applied. When clicking on "OK" in the portrait selection.

seedhartha:
Fixed.

seedhartha:
I also fixed the sanitizer errors you mentioned. Not sure about the segfault though.

DrMcCoy:
The Sanitizers now complain within the pathfinding code, when Trask wants to move to the PC: https://gist.github.com/DrMcCoy/84a443f4c85c728fde56f694843c4aa0

seedhartha:
I guess it has to do with the commits Supermanu pushed recently.

DrMcCoy:
Ah, yes, you're right. master - Supermanu's changes + this PR works.

DrMcCoy:
Okay, merged as 1b04f4792...15f47ff05, thanks! :)

I'll open an issue for the new pathfinding thing.

PR #452, opened 2019-02-23
NWN2: Set the cursor type for a mouse-over of a door or placeable

rjshae:
This should complete the basic cursor type functionality for NWN2,
at least until the UI is added. Note that I verified the DefAction enum
values via a save file cross-check. The order is slightly different than
the menu picks in the toolset.


DrMcCoy:
Merged as 75840370a...09cb2fdbf, thanks! :)

PR #457, opened 2019-02-24
GRAPHICS: Project using perspective matrix

mirv-sillyfish:
When performing a project operation, use 3D perspective matrix.


DrMcCoy:
Looks good now. Merged as f72310e20, thanks! :)

PR #459, opened 2019-02-25
NWN2: Create an item class

rjshae:
This pull request implements some of the basic functionality for NWN2 items. It is the first step in trying to display the inventory interface for the chest in the tutorial area, which seems like a relatively simple starting point for building out the UI.


DrMcCoy:
Merged as edd15e154...3297b5a58, thanks! :)

PR #460, opened 2019-02-25
ENGINES: PC and party management refactoring for KotOR games

seedhartha:
- Party management was extracted to a separate class (PartyController).
- Ownership of PC and party members was moved from the Module to the Area.
- PC != controllable creature. PC is a character generated by the player. We control the party leader. Changed the source code accordingly.

As a bonus, party members are now interactable, follow the party leader and will be saved on module change.

Known issues: 
- Trask will not immediately follow the PC after joining the party. This is because of a script that resets his actions.


DrMcCoy:
Merged as 3297b5a58...87e25a4a3, thanks! :)

PR #461, opened 2019-02-26
ENGINES: Implement heartbeat, use native KotOR AI

seedhartha:
I turns out that AI in KotOR games is handled by a special script (which is open source, BTW). I implemented the heartbeat event to invoke that script, as well as some of the missing functions to make Trask follow the PC.


src/engines/kotorbase/module.cpp
```diff
@@ -673,6 +677,17 @@ void Module::handleActions() {
 	}
 }
 
+void Module::handleHeartbeat() {
+	const int kHeartbeatInterval = 1000; // ms
```
DrMcCoy: Is every second correcct for KotOR? In NWN, the heartbeat is invoked every 6 seconds (= the length of a combat round)

seedhartha: I'm not really sure. What would be the right place to look for a reference?

DrMcCoy: On Deadly Stream, a KotOR modding site, they also say 6s: https://deadlystream.com/topic/6371-on-heartbeat/ (the Lexicon is https://nwnlexicon.com/) , https://deadlystream.com/topic/2062-tsl-textures-remastered/?page=9&tab=comments#comment-23797

DrMcCoy: In the future, we should probably have a "round handler" or something in the KotOR engine, that synchronously does all the things required per round. I.e. calling all heartbeat scripts, evaluating pathfinding, checking and firing script events like perception, going through combat initiative, etc.

seedhartha: Changed the interval to 6 seconds.

DrMcCoy:
Merged as ef40916e2...82273fc5f, thanks! :)

PR #462, opened 2019-02-26
ENGINES: Hide selection in dialog, container and party menu

seedhartha:


seedhartha:
Commit moved to #461 because it depends on it.

PR #463, opened 2019-02-27
ENGINES: Update selection mechanics to match KotOR games

seedhartha:
This PR updates selection mechanics in KotOR to match the original games. We now have two selection circles: persistant one for the current target and temporary one for the hovered object. In order to use the object, you have to target it first. Also, locked situated objects will now have (empty) target buttons above them.

Known issues:
- Cursor should only change on hover when the object is targeted
- Selection circles only work for situated objects currently (but creatures still need to be targeted prior to activation)

Note: Should probably port that to KotOR 2 as well.


DrMcCoy:
Merged, thanks! :)

PR #465, opened 2019-02-28
COMMON: Use random number generation from C++11

### LINKED ISSUE #449: Non-random numbers using Visual C++ runtime
std::rand always returns the same sequence when compiling xoreos using Visual Studio. Apparently, this is because we call std::srand from the main thread and std::rand from the game thread.

seedhartha:
This PR replaces srand/rand functions with random number generation from C++11. This fixes #449.


DrMcCoy:
Merged, thanks! :)

PR #466, opened 2019-02-28
ENGINES: Refactor GUI for KotOR games

seedhartha:
- GUI class from KotOR 2 is replaced with an utility function (it was limiting reuse possibilities)
- Duplicate dialog logic was moved into the base dialog GUI while the game-specific logic was extracted from it
- In-game GUI and HUD were merged between the two games, which effectively enabled selection mechanics in KotOR 2

### DIFF
--- a/src/engines/kotor/gui/dialog.cpp
+++ b/src/engines/kotor/gui/dialog.cpp
@@ -22,25 +22,7 @@
  *  Dialog GUI for Star Wars: Knights of the Old Republic.
  */
 
-#include "src/common/configman.h"
-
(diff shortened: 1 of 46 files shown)


DrMcCoy:
Merged as cdfaf8d37...aca58a632, thanks! :)

There is, however, an invalid read in the Xbox version of KotOR, when clicking on the chest (the one with your starting stuff in it). Probably because the Xbox interface is different. https://gist.github.com/DrMcCoy/89fe871819710b20feae655c9e2386a7

PR #467, opened 2019-03-01
NWN2: Additional item functionality

rjshae:


src/engines/nwn2/inventory.cpp
```diff
@@ -83,6 +83,26 @@ Item *Inventory::getNextItemInInventory() {
 	return 0;
 }
 
+Item *Inventory::createItem(const Common::UString &blueprint, uint16 stackSize, const Common::UString &tag) {
+	if (blueprint.empty() || stackSize == 0)
+		return 0;
+
+	// Create the item
+	Item *item = new Item(blueprint, stackSize, tag);
+	if (item) {
```
DrMcCoy: Can you rewrite this method to be less nested, by returning directly? I.e.

```
if (!item)
	return 0;
```

etc.

src/engines/nwn2/inventory.cpp
```diff
@@ -83,6 +83,26 @@ Item *Inventory::getNextItemInInventory() {
...
+		// Was the tag set?
+		if (!item->getTag().empty()) {
+			// TODO: Check for a merge into a stack of similar items
+			uint16 slot = getFirstEmptySlot();
+			inventoryItems[slot] = item;
+		} else {
+			delete(item);
```
DrMcCoy: No need for parens on delete.

As an alternative, you could make item a std::unique_ptr (or a Common::ScopedPtr, but I'd like to phase usage of ScopedPtr out in favour of unique_ptr) and return item.release() on success.

src/engines/nwn2/inventory.cpp
```diff
@@ -101,6 +121,28 @@ void Inventory::clear() {
...
+	const uint16 max = inventoryItems.rbegin()->first;
+	if (max < total)
+		return total;
+
+	// Look for the first open slot
+	uint16 slot;
+	std::map<uint16, Item *>::const_iterator item;
+	for (item = inventoryItems.begin(), slot = 0; item != inventoryItems.end(); ++item, slot++)
+		if (slot < item->first)
+			return slot;
```
DrMcCoy: What exactly is that doing? That looks...confused.

Should the inventory be a `std::map<uint16, Item *>` in the first place? Wouldn't it make more sense to have a flat `std::vector<Item *>`, of the size of the inventory, with the item being nullptr if unused?

(Or maybe even a `std::vector<std::unique_ptr<Item>>`, or `std::vector<std::shared_ptr<Item>>`, if the inventory is supposed to manage the item.)

Finding the first non-used inventory slot would then be a `std::find_if`. You could even just return the iterator on this method (and that would also give you a way to express "no free slot", by returning the end iterator). If you need the index later (because a script call wants it or something), you can use `std::distance()` then.

src/engines/nwn2/inventory.cpp
rjshae: Okay.

src/engines/nwn2/inventory.cpp
rjshae: That's my attempt to optimize the search for a gap in the map.

1) Most creatures are going to have an empty inventory, so it checks that right away.
2) The next case is for inventory with no gaps, which will be true for most creatures as well as for PC's that have sorted their inventory. In that case just checking the length is enough to see if there are gaps.
3) Finally, I do the slow crawl through the list looking for the first gap.

The PC inventory could potentially have a lot of gaps, so I just choose map for optimum storage. But yes it could use a vector, if you don't mind the large amount of wasted storage. I suppose I could allocate a row at a time.

DrMcCoy: I'd say optimizing for a large amount of gaps here would be premature optimization. Especially considering the usual number of items in an inventory and that you're giving up cache locality with a map. I'm not really sure such a map would even be faster than a vector here, to be frank.

When we see that crawling through the inventory is a bottle neck, we can optimize it further. Preferably with some profiling, so that we can see how different structures perform.

rjshae: Okay, I will simplify it down. Thanks.

src/engines/nwn2/inventory.cpp
Mingun: It `Item` has no overrided `operator new`, then checking for `nullptr` is not required -- an exception will be thrown if memory can not be allocated

DrMcCoy: Oh, I didn't see that it's just a plain new and not a function() call. My bad.

(However, whether an exception is thrown is optional. The operating system might simply overcommit. But yes, in either case, checking for nullptr is useless.)

src/engines/nwn2/inventory.cpp
rjshae: I'm intentionally making the items so that they would be transferable between inventories. Items can have inventories of other items, so it could potentially get recursive. That's the reason I was avoiding ScopedPtr. Does that make sense or am I being too paranoid? I think eventually an item will also need to carry a back pointer to its inventory so that a delete object can be performed.

DrMcCoy: You can still use a ScopedPtr/unique_ptr locally, and call the release() method, which returns the raw pointer and releases it from the managed pointer. This is just so that you can return from the function in the error case without having to worry about manually deleting the object.

rjshae: If I'm understanding the std::unique_ptr functionality correctly, I can't return the object itself with the getFirstItemInInventory & getNextItemInInventory calls(?).

DrMcCoy: You can.

```cpp
Foo *createFoo() [
  std::unique_ptr<Foo> foo(new Foo(23, 5);

  ...
  if (somethingFailed)
    return nullptr;
  ...

  return foo.release();
}

Comment (DrMcCoy) [reply src/engines/nwn2/inventory.cpp]: Or maybe I don't understand your question?

Comment (rjshae) [reply src/engines/nwn2/inventory.cpp]: Well I don't think I want to release the ownership. The pointer is being accessed by script calls via the getFirstInInventory & getNextInInventory. But it should still remain in the inventory.

Comment (DrMcCoy) [reply src/engines/nwn2/inventory.cpp]: No, that releases the pointer from the temporary local unique_ptr. The purpose of that local unique_ptr would be solely to manage the memory inside the function. With the return release, you transfer the ownership to the caller.

Comment (DrMcCoy) [reply src/engines/nwn2/inventory.cpp]: We can talk about that in-depth on IRC if you want :)
Later today, though, because it's still early here and I'd like to try for some more sleep :P

Comment (rjshae) [reply src/engines/nwn2/inventory.cpp]: I'm thinking a std::shared_ptr might work better in this instance.

Comment (DrMcCoy) [reply src/engines/nwn2/inventory.cpp]: Maybe. But that change is more intricate, so maybe that should be left for the future. While just swapping out the raw pointer within the function for a unique_ptr would be easy to do now.

Comment (rjshae) [reply src/engines/nwn2/inventory.cpp]: For example, I've got an item in a crafting table inventory. The PC runs a script that modifies the itemProperty list on the item, then exits, leaving the item still in the table inventory. The script will need to grab the item instance, make some changes, then exit.

Comment (rjshae) [reply src/engines/nwn2/inventory.cpp]: Yeah, I guess I'm not understanding how removing the unique_ptr from the inventory to dispatch it to a temporary script is helping. Oh well.

Comment (DrMcCoy) [reply src/engines/nwn2/inventory.cpp]: Huh? Swapping in the std::unique_ptr for the raw pointer here changing nothing outside of the function. You're still returning the raw pointer. The release is only there to release it from the temporary unique_ptr that was created for the function.

Literally the only thing that changes is that you won't have to manually delete the item when an error in the middle of the function happens. That's the only purpose of the local unique_ptr. The benefit is that you can't forgot to delete it, so there's never a leak.

Comment (DrMcCoy) [reply src/engines/nwn2/inventory.cpp]: I.e. this is what I'm talking about, rewriting this function to use a unique_ptr (and keeping the rest as-if, like the map, for clarity):

```cpp
Item *Inventory::createItem(const Common::UString &blueprint, uint16 stackSize, const Common::UString &tag) {
	if (blueprint.empty() || stackSize == 0)
		return nullptr;

	std::unique_ptr<Item> item(new Item(blueprint, stackSize, tag));
	if (item->getTag().empty())
		return nullptr;

	const uint16 slot = getFirstEmptySlot();
	inventoryItems[slot] = item.release();

	return inventoryItems[slot];
}
```

The unique_ptr makes sure that the object is always deleted correctly, unless release was called (which transferes ownership).

rjshae: Guess I'm confused then. According to the docs., calling release replaces the stored pointer in the unique_ptr with null. The inventory no longer has access to that item instance. That's not what I want to do.

DrMcCoy: Yes, it replaces it with null *and returns the previous pointer*, so that it can be used somewhere else. Like, to store it inside the map here.

That's what `inventoryItems[slot] = item.release();` does, it transfers the ownership from the unique_ptr into the inventory map.

That local unique_ptr is there so that, effictively, the function owns the item, until it's released in the inventory map.

rjshae: Ah, okay. I was thinking at a completely different level, believing you wanted the vector element to be a unique_ptr. Gotcha. No you just want it there long enough to assign to the slot. Sure, no problem.

DrMcCoy: Yes, exactly. Sorry for the confusion :)

rjshae: Thank you for helping me work through it. :)

rjshae:
I did a rewrite using std::vector, but something looks to have gone horribly wrong.

DrMcCoy:
Oh, yeah, sorry, my bad.

But I don't get why the CIs aren't building this PR. You're still basing off of a revision before the change. The conflict with the current main branch should only be relevant when merging the PR, and I could have resolved the conflict myself then.

rjshae:
Darn, more collisions.

rjshae:
Okay, hopefully that'll work now.

DrMcCoy:
Merged as 35d5293a7...a3c216cc1, thanks! :)

A few comments, maybe something for future PRs:
- `Inventory::getFirstItemInInventory()` looks complicated. Can that maybe be simplified using the stuff in `<algorithm>`? The loops smell like a find_if, for example
- Same for `Inventory::getFirstEmptySlot()`
- You normally don't need to care about std::vector allocating by line. The usual implementation of vector already allocates exponentially, i.e. it doubles its capacity when it runs out
- But for indices into std::vector, etc., we usually use size_t. _lastRetrieved could be changed into an iterator, though

Nothing tragic, but I think the code can be made a bit more readable that way. :)

rjshae:
Okay, I'll see what I can do. Thanks.

rjshae:
> You normally don't need to care about std::vector allocating by line. The usual implementation of vector already allocates exponentially, i.e. it doubles its capacity when it runs out

The problem I ran into is that the '[]' operator doesn't assign anything unless it has already been filled with a resize() call.

DrMcCoy:
Yes, of course, but you can just resize() one more, the capacity is not necessarily equal the size

rjshae:
Well it might not be just one more, but 20 more, depending on how the player moves stuff around the inventory (leaving large gaps). The extra padding was me just trying to code in performance again, as I'd read that many single slot resizes can be expensive. But okay, one it is.

DrMcCoy:
Yeah, you just need to resize up to where you want the object to go. :)

When a std::vector reallocates, it is pretty expensive, yeah, since you allocate a new continous block and copy the old data into it. That's exactly why the usual growth strategy in basically all implementations is to double the capacity when running out. "Amortization" is the term that's commonly used to describe this feature.

PR #468, opened 2019-03-01
ENGINES: Implement OpenLock action execution, fixes, HUD refactoring

seedhartha:
- Object click is first handled in Module, then in Area (this fixes premature door activation)
- Make doors not usable when opened
- Hide selection when entering menus or unloading the module


DrMcCoy:
Merged as aca58a632...997de0b46, thanks! :)

PR #469, opened 2019-03-01
ENGINES: Refactor camera in KotOR games, implement collision detection

seedhartha:
This PR implements collision detection for the satellite camera in KotOR games.

Known issues:
- Collision with situated objects is not handled (would probably need Supermanu's input on that).


DrMcCoy:
Merged as 997de0b46...bf8b25810, thanks! :)

The camera handling in the Ebon Hawk, in the prologue of KotOR2, feels far better now, yes :)

PR #472, opened 2019-03-02
BUILD: Fix compilation on Windows 7 using VS and CMake

seedhartha:
That was a tough one. Linking to mincore.lib adds a dependency on a DLL that is not present in Windows 7. The engine would compile just fine, but would not run.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,10 @@ if(CMAKE_HOST_WIN32)
     endforeach(d)
 
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${WARNINGS}")
-    link_libraries(winmm.lib imm32.lib mincore.lib version.lib)
+    link_libraries(winmm.lib imm32.lib version.lib)
+    if((${CMAKE_SYSTEM_VERSION} VERSION_EQUAL 6.2) OR (${CMAKE_SYSTEM_VERSION} VERSION_GREATER 6.2))
+      link_libraries(mincore.lib)
+    endif()
   endif()
 endif()
 


DrMcCoy:
Huh, okay? Weird.

But what about the other libs in there? Are none of them needed on Windows 7?

DrMcCoy:
And does it otherwise work fine on newer and on older Windowses? I.e. when is mincore needed and when isn't it?

Seems like Beren added these for AppVeyor? I don't really know anything about that at all :P

seedhartha:
[Here](https://docs.microsoft.com/en-us/windows/desktop/apiindex/windows-8-api-sets) it is stated that binaries that link to MinCore.lib or MinCore_Downlevel.lib are not designed to work on Windows 7, Windows Server 2008 R2 or earlier.

They do seem to be a requirement for Windows 8 and above however (AppVeyor is most likely Windows Server 2012 R2). Basically, as is, we only support Visual Studio builds on Windows 8 and Windows 10, which is a shame (Visual Studio really feels superior to MinGW setup on Windows and Windows 7 is still widely used).

As for the other libraries, the engine compiles just fine without them on my machine.

DrMcCoy:
> Visual Studio really feels superior to MinGW setup on Windows 

I mean, I disagree there, but that's me. :P

> They do seem to be a requirement for Windows 8 and above

So the check should probably be for "GREATER 6.1"? Or however the Windows version numbers work.

seedhartha:
That should do it, I think. For the reference: [comparison](https://en.wikipedia.org/wiki/Comparison_of_Microsoft_Windows_versions) of Windows versions.

DrMcCoy:
Merged as 9f46a0ae3, thanks! :)

Issue #473, opened 2019-03-02
BUILD: Linking failure using MinGW and CMake


seedhartha:
Trying to build the engine using latest MSYS 2, MinGW-w64 and CMake on Windows 7 results in the linking failure. [Here](https://gist.github.com/seedhartha/968478ab690026cd9cc74cfa893be3ef) are the errors. There are no compilation errors and the engine builds just fine using autotools.

DrMcCoy:
Can you build with "make VERBOSE=1", so that the whole compiler/link command lines are printed?

I don't get it. Somehow, stuff is missing there, and it makes no sense

seedhartha:
Sure, [here](https://gist.github.com/seedhartha/41cb45732d57cc3a91b1bb2a018b91e0) is a verbose log.

seedhartha:
Okay, yeah, changing the order or libraries in rules.mk fixes most of the errors.

DrMcCoy:
Aaah, okay, that makes sense :)

seedhartha:
Fixed.

PR #475, opened 2019-03-02
BUILD: Fix build using MinGW and CMake

seedhartha:
This PR, when paired with #474, fixes the #473. The issue was caused by circular dependencies in the video and graphics subsystems, as well as, an invalid order of components in the engines subsystem.

### DIFF
--- a/src/engines/rules.mk
+++ b/src/engines/rules.mk
@@ -36,9 +36,9 @@ src_engines_libengines_la_SOURCES += \
     $(EMPTY)
 
 src_engines_libengines_la_LIBADD = \
-    src/engines/aurora/libaurora.la \
-    src/engines/odyssey/libodyssey.la \
     src/engines/kotorbase/libkotorbase.la \
+    src/engines/odyssey/libodyssey.la \
+    src/engines/aurora/libaurora.la \
     $(EMPTY)
 
 # Subdirectories
--- a/src/graphics/aurora/rules.mk
+++ b/src/graphics/aurora/rules.mk
@@ -19,10 +19,7 @@
 
 # Aurora-specific graphics-related loaders.
 
-noinst_LTLIBRARIES += src/graphics/aurora/libaurora.la
-src_graphics_aurora_libaurora_la_SOURCES =
-
-src_graphics_aurora_libaurora_la_SOURCES += \
+src_graphics_libgraphics_la_SOURCES += \
     src/graphics/aurora/types.h \
     src/graphics/aurora/texture.h \
     src/graphics/aurora/texturehandle.h \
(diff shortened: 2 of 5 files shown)


seedhartha:
BTW, I think it's worth mentioning in the wiki, that "make -jX", where X is core count + 1 can drastically improve compilation speed on modern CPU.

DrMcCoy:
Yeah, I just assume that's common knowledge, but perhaps not.

DrMcCoy:
Merged as 817402634, thanks! :)

Can you check xoreos-tools and Phaethon as well?

seedhartha:
Just checked: xoreos-tool builds fine using autotools but fails to do so using CMake and either MinGW or Visual Studio. [This](https://gist.github.com/seedhartha/1ff0e9b324c64cf44c71effa3ee0739e) is a log from the MinGW build.

Visual Studio case looks like a problem with the latest version of Boost (easily solved by including <set> in problematic files). Also, this:

`[xoreos-tools root]\src\archives\files_dragonage.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj`

seedhartha:
Phaethon: builds OK using MinGW with both autotools and CMake, but fails using Visual Studio with the following erros:

`
Error	C3615	constexpr function 'qCountLeadingZeroBits' cannot result in a constant expression (compiling source file D:\Source\phaethon\src\gui\panelbase.cpp)	gui	c:\msys64\mingw64\include\qtcore\qalgorithms.h	847	
Error	C3615	constexpr function 'qCountLeadingZeroBits' cannot result in a constant expression (compiling source file D:\Source\phaethon\src\gui\panelmanager.cpp)	gui	c:\msys64\mingw64\include\qtcore\qalgorithms.h	847	
Error	C3615	constexpr function 'qCountLeadingZeroBits' cannot result in a constant expression (compiling source file D:\Source\phaethon\src\phaethon.cpp)	phaethon	c:\msys64\mingw64\include\qtcore\qalgorithms.h	847	
`

DrMcCoy:
> Visual Studio case looks like a problem with the latest version of Boost (easily solved by including `<set>` in problematic files).

I.e. this is probably our fault to begin with, not properly including what we're using?

> `[xoreos-tools root]\src\archives\files_dragonage.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj`

Argh :/
Is there a way to enable bigobj for that file? Or can we just enable bigobj globally when using MSVC?

> Error C3615 constexpr function 'qCountLeadingZeroBits' cannot result in a constant expression (compiling source file [phaethon root]\src\gui\panelbase.cpp) gui c:\msys64\mingw64\include\qtcore\qalgorithms.h 847

Narf, that seems to be a bug in Qt: https://bugreports.qt.io/browse/QTBUG-67259

PR #476, opened 2019-03-02
KOTOR: Improve camera handling

seedhartha:
This PR improves the satellite camera in KotOR games by fixing the movement lag and also interpolating camera position when it hits the obstacle and returns back to normal.


DrMcCoy:
Merged, thanks! :)

PR #477, opened 2019-03-02
KOTOR: Fix room visibility not being evaluated on start

seedhartha:
Also use set for visible rooms instead of a vector.

### DIFF
--- a/src/engines/kotorbase/area.cpp
+++ b/src/engines/kotorbase/area.cpp
@@ -722,7 +722,7 @@ void Area::updateRoomsVisiblity() {
 		bool visible = o->isVisible();
 
 		if (visible && !shouldBeVisible)
-			o->hideSoft();
+			o->notifyNotSeen();
 		else if (shouldBeVisible && !visible)
 			o->show();
 	}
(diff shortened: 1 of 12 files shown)


src/engines/kotorbase/area.cpp
```diff
@@ -241,13 +241,20 @@ void Area::show() {
 
 	GfxMan.lockFrame();
 
+	std::set<Common::UString> visibleRooms(getRoomsVisibleByPartyLeader());
+
 	// Show rooms
-	for (RoomList::iterator r = _rooms.begin(); r != _rooms.end(); ++r)
-		(*r)->show();
+	for (auto r : _rooms) {
```
DrMcCoy: should be auto &, and likewise below

seedhartha: Why though? It would be a reference to a pointer. Makes no sense to me :P

seedhartha:
Fixed foreach loops as requested. I have also restored the logic where doors where always visible (because they may not be recognized as part of the room).

DrMcCoy:
Merged as df7dfd260...6d4fb1313, thanks! :)

PR #478, opened 2019-03-03
KOTOR: Weapon drawing, party leader switch (TAB), selection of all objects types

seedhartha:


DrMcCoy:
Merged as 6d4fb1313...58fb1f372, thanks! :)

PR #481, opened 2019-03-04
BUILD: Fix CMake compilation using MSVC x64

seedhartha:
Also move MSVC linker options from warnings to a more generic section.


DrMcCoy:
Merged as 35d5293a7, thanks! :)

PR #482, opened 2019-03-04
KOTOR: Add CreatureInfo class, save PC state on module change

seedhartha:


src/engines/kotorbase/creatureinfo.cpp
```diff
@@ -0,0 +1,268 @@
...
+	_skills = other._skills;
+	_abilities = other._abilities;
+
+	return *this;
+}
+
+int CreatureInfo::getClassLevel(Class charClass) const {
+	auto it = std::find_if(_levels.begin(),
+	                       _levels.end(),
+	                       [&](const ClassLevel &cl) -> bool { return cl.characterClass == charClass; });
```
DrMcCoy: I generally prefer this to be formatted as

```cpp
	auto it = std::find_if(_levels.begin(), _levels.end(), [&](const ClassLevel &cl) {
		return cl.characterClass == charClass;
	});
```

(and the `-> bool` isn't necessary here, since it can be deduced.)

DrMcCoy:
Merged as 4a521692e...d15ec350d, thanks! :)

(I took the liberty of reformating that find_if :P)

Issue #479, opened 2019-03-04
KOTOR: Not seen objects are interactable


seedhartha:
Currently it is possible to interact with all objects in the area, regardless of how far they are or whether they are seen by the player.

seedhartha:
It could partially be solved by limiting the activation range and issuing a move to command to the party leader. Otherwise, not sure how to handle it right.

DrMcCoy:
How does the original game handle this? Does it check for visibility/occlusion?

rjshae:
At least in NWN2, I think it must queue an ActionMoveToObject action followed by an ActionInteractObject. I'm not sure how they compute the range for the ActionMoveToObject call, unless it's using one of the appearance.2da columns such as PERSPACE.

PR #483, opened 2019-03-04
KOTOR: Fix selection, use object action, actions refactoring

### LINKED ISSUE #480: KOTOR: Target selection projection bug
Currently, if you select an object and then start moving in the opposite direction, selection will appear on screen (the object is behind the camera).

seedhartha:
- Fix #480 as suggested
- Issue a "use object" action on click, like in original games (partially solves #479, but still)
- Use ExecutionContext structure in ActionExecutor

### DIFF
--- a/src/engines/kotorbase/actionexecutor.cpp
+++ b/src/engines/kotorbase/actionexecutor.cpp
@@ -40,20 +40,23 @@ namespace Engines {
 
 namespace KotORBase {
 
-void ActionExecutor::executeActions(Creature &creature, Area &area, float dt) {
-	const Action *action = creature.getCurrentAction();
+void ActionExecutor::executeActions(const ExecutionContext &ctx) {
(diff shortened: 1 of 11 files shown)


DrMcCoy:
Merged, thanks! :)

Sometimes, when clicking on an object, the ingame menu pops up instead, though. Which is weird.

(Also, just as a style comment, relatively recently I started finding it useful to mark non-changing local variables, like that `bool onScreen` as const. It's just one less thing I have to worry about when reading the code. "Oh, this is const, so I don't have track any potential changes going forward", basically.)

DrMcCoy:
Hmm, clicking on objects with conversations doesn't work anymore now, though.

Like the "Skip prologue" object in KotOR2 (after selecting "continue prologue" on the auto dialogue), or the computer terminal where you can open the medbay doors to where Kreia is in the first real area, Peragus, of KotOR2.

PR #486, opened 2019-03-05
NWN2: Convert inventory crawl to use find algorithm

rjshae:
I'm not sure if it's an improvement in terms of reducing complexity,
but here's the first draft of the converted inventory class.

### DIFF
--- a/src/engines/nwn2/inventory.cpp
+++ b/src/engines/nwn2/inventory.cpp
@@ -24,6 +24,7 @@
 
 #include <iterator>
 #include <vector>
+#include <algorithm>
 
 #include "src/common/error.h"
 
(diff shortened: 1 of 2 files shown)


rjshae:
I think there's a way to refactor the insertItem call so it's a little cleaner, so I'm going to take a look at it tonight. Perhaps you'll want to hold off your review until then? Thanks.

rjshae:
I finished the update. Thank you.

rjshae:
P.S. I have a concern that a suitably-written NWN2 script may cause the ItemSlots vector to change between calls to getNextItemInInventory(), which is why I'm still using a numerical counter rather than storing the iterator. Perhaps that concern is misplaced?

rjshae:
I'll merge this into a Store class pull request.

PR #487, opened 2019-03-06
GRAPHICS: Fix head animation glitches in KotOR 2 (WIP)

seedhartha:
This PR attempts to fix head animation glitches with some models in KotOR 2 (particularly, pmhc06 and p_handmaidenh). The issue seems to be caused by head, tongue (and tongie) nodes being children to a bone, rather than a root node.


DrMcCoy:
Wait, why did you close this? Didn't it work?

seedhartha:
It only partially solved the issue (found out that the bone mapping was also not interpreted correctly). Besides, I believe I can think of a cleaner solution.

PR #491, opened 2019-03-09
GRAPHICS: Fix head animations in K2, refactor skeletal animation

seedhartha:
This PR fixes a long-standing issue with certain head models in KotOR 2 (particularly, p_handmaidenh and pmhc06).

The issue was caused by two things. Firstly, we interpreted the bone map as a map from node numbers to bone indices, while it was a map from node indices (in array of root node's children) to bone indices. Secondly, in our current implementation, if a model node is both skinned and is a child to a bone node, then it will be transformed twice, both by the renderer and by the skeletal animation routine. Now, in a majority of models, head and tongue nodes are children to a root node, while in these particular heads they were children to bone nodes.

I have also refactored animations by extracting the skeletal animation code to a separate class and using recursion to compute node transformation matrices.

### DIFF
--- a/src/engines/nwn2/area.cpp
+++ b/src/engines/nwn2/area.cpp
@@ -495,8 +495,8 @@ void Area::loadTileModels() {
 		t->model->setOrientation(0.0f, 0.0f, 1.0f, rotation);
 
 		// Rotate static floors back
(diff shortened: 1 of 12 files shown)


DrMcCoy:
Merged as 4650dce51, thanks! :)

PR #494, opened 2019-03-12
ALL: Use thread and atomic from std, refactor animation thread

seedhartha:
Also fix a race condition in the animation thread, where a model list could be changed while flushing.


DrMcCoy:
Moving to the C++11 thread stuff is a good idea, yeah. :)

But could you please split moving from Boost.Atomic to std::atomic, from SDL_Thread to std::thread and refactoring the animation thread into individual commits? This makes the history cleaner and easier to understand. Also, easier to move the individual pieces over to xoreos-tools and Phaethon.

seedhartha:
Done.

src/common/thread.cpp
```diff
@@ -55,16 +57,19 @@ bool Thread::createThread(const UString &name) {
 	_name = name;
 
 	// Try to create the thread
-	if (!(_thread = SDL_CreateThread(threadHelper, _name.empty() ? 0 : _name.c_str(), static_cast<void *>(this))))
+	try {
+		_thread = std::thread(threadHelper, static_cast<void *>(this));
```
DrMcCoy: Aw, we're using the capability of naming our thread here. With `SDL_CreateThread`, the name would appear in the debugger, process monitors, etc.

But it seems like `std::thread` just doesn't have that functionality. Guess we're going to have to live with that

src/common/rules.mk
```diff
@@ -58,7 +58,6 @@ src_common_libcommon_la_SOURCES += \
     src/common/platform.h \
     src/common/debugman.h \
     src/common/debug.h \
-    src/common/atomic.h \
```
DrMcCoy: We can then also remove remove Boost.Atomic from
- our build files
  - https://github.com/xoreos/xoreos/blob/master/CMakeLists.txt#L264
  - https://github.com/xoreos/xoreos/blob/master/configure.ac#L345
  - https://github.com/xoreos/xoreos/blob/master/Makefile.am#L74
- our documentation
  - https://github.com/xoreos/xoreos/blob/master/BUILDDEP.md
- the distribution build files
  - https://github.com/xoreos/xoreos/blob/master/dists/fedora/xoreos.spec#L36
  - https://github.com/xoreos/xoreos/blob/master/dists/debian/control#L5

seedhartha:
Updated the build files and documentation as requested.

DrMcCoy:
Thanks, merged as 12bbc2c06...ca8c86ab1! :)

One thing, though, is it just me or are the KotOR animation now again less smooth with vsync enabled? Especially noticable with the walking animation.

seedhartha:
Yes, it could be. I intentionally made the renderer skip some animation frames to not suffer the performance penalty (using a single mutex was fast, but not very thread-safe unfortunately). I'm planning to further optimize the animation system in the future, possibly using a thread pool.

PR #495, opened 2019-03-13
NWN2: Include ItemProperty class list in the Item class

rjshae:


src/engines/nwn2/item.cpp
```diff
@@ -51,6 +52,14 @@ Item::Item(const Common::UString &blueprint, uint16 stackSize, const Common::USt
 }
 
 Item::~Item() {
+	clear();
+}
+
+void Item::clear() {
+	for (std::list<ItemProperty *>::const_iterator prop = _itemProperties.begin();
```
DrMcCoy: Hmm, my thought process here:

- Why isn't that using a range-based for?
- Wait, why is that a `std::list<ItemProperty *>` if the list solely manages its memory? Can't that be a `std::list<std::unique_ptr<ItemProperty>>`?
- Wait, why are that ItemProperties pointers in the first place? That can be a `std::list<ItemProperty>`
- Heck, that can even be a `std::vector<ItemPropery>`. The set of item properties is only seldom changed, and in either case, an ItemProperty is only a handful of bytes big

In summary, please change `_itemProperties` into a `std::vector<ItemProperties>`. Add a `_itemProperties.reserve(iprp.size());` into `loadItemProperties()` for good measure, so that it's only allocated once during loading.

rjshae: There are script functions to add and remove item properties on the fly (to allow for crafting and spells), which is why I used a list.

DrMcCoy:
Merged as ca8c86ab1...fd3f17b1d, thanks! :)

(I did simplify it a tiny bit, by using emplace_back() and find_if, and removing the, essentially, copy constructor where the default one can be used)

PR #498, opened 2019-03-16
GRAPHICS: Fix animation smoothness, optimize animations

seedhartha:
This PR should fix a regression in the animation smoothness with VSync ON, while also retaining thready-safety and acceptable performance.

I also tried to implement the animation thread pool, but it did not give a noticeble performance boost while introducing it's own set of synchronization problems.

### DIFF
--- a/src/graphics/aurora/animationthread.cpp
+++ b/src/graphics/aurora/animationthread.cpp
@@ -35,57 +35,59 @@ namespace Graphics {
 
 namespace Aurora {
 
+const int kPauseDuration = 10;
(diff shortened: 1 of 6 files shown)


DrMcCoy:
Merged as e13054f65...ed82e8cb5, thanks! :)

DrMcCoy:
@seedhartha, we've got leaks in KotOR2 now, though: https://gist.github.com/DrMcCoy/06c014ab7c3d8235af0b90a9ed3dd967

PR #499, opened 2019-03-17
KOTOR: Base console, new commands, fix creature loading regression

seedhartha:
This PR:
- Adds a base console for KotOR games.
- Adds console commands "getactiveobject" and "actionmovetoobject".
- Fixes a regression, where Sith body models were not loaded on Korriban. The problem was that those models only have 3 texture variations, but are equipped with clothing that have texture variations of 5 and 6.


DrMcCoy:
Merged as ed82e8cb5...4721e84ca, thanks! :)

PR #500, opened 2019-03-18
Fix memory leaks

seedhartha:


DrMcCoy:
The code itself looks good, thanks :). But I want to check in the evening if it works

DrMcCoy:
Merged as 217eefeb7...ad788a50d, thanks! :)

PR #501, opened 2019-03-18
ACTIONSCRIPT: Update

Nostritius:
A new small ActionScript Update

### DIFF
--- a/src/aurora/actionscript/array.cpp
+++ b/src/aurora/actionscript/array.cpp
@@ -46,8 +46,8 @@ static Variable as_pop(AVM &avm) {
 }
 
 Array::Array(const std::list<Variable> &values) : _values(values) {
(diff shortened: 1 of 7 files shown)


DrMcCoy:
Looks good, thanks! :)

Will merge in the evening.

One comment, though. The four bool flags are getting kind of...unwieldy and confusing. In the past, I've written functions with multiple bool parameters as well, but they've always been kind of awkward. On the call-site, you never see what they mean, they're easy to mix-up, etc.

Maybe those preload* flags should be a class enum, or multiple of those. So that you can say, for example, Preload::Global or something (depending on what makes sense there).

Not something that should be added to this PR, but maybe something we should think about.

DrMcCoy:
Merged as 4721e84ca...217eefeb7, thanks! :)

PR #502, opened 2019-03-18
AURORA: GFF3Writer Update

Nostritius:
This PR updates the GFF3Writer and some of it's components to reuse complex values and through that reduce the size of the generated Files. The original games are already doing that.

### DIFF
--- a/src/aurora/gff3writer.cpp
+++ b/src/aurora/gff3writer.cpp
@@ -42,6 +42,13 @@ GFF3WriterStructPtr GFF3Writer::getTopLevel() {
 }
 
 void GFF3Writer::write(Common::WriteStream &stream) {
+	// Extract all individual values of the fields.
+	std::list<Value> individualValues;
+	for (const auto &field : _fields) {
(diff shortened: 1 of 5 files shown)


DrMcCoy:
What the heck is happening there: https://ci.appveyor.com/project/DrMcCoy/xoreos/builds/23165421#L6689 :/

Looks like GFF3Writer.WriteNestedStructs is crashing? I see no sanitizer complaints on my machine, tho...

Nostritius:
I have now found and fixed the bug, which caused the program to fail in appveyor. For anyone having a similar issue, the reason for that is, that msvc enforces strict weak ordering in std::map, with certain invariants. Is an invariant violated, the program will fail. I made thee < operators more strix to prevent as far as possible something like that for the future.

DrMcCoy:
Oooh, interesting! Thanks for investigating and fixing this! :)

Nostritius:
Now the c++11 bug in clang should also be solved.

DrMcCoy:
Merged as 8e12d510d...5a16bf71e, thanks! :)

PR #504, opened 2019-03-19
NWN2: Implement itemproperty declaration calls

rjshae:
I looked at save file dumps to try and figure out what went into the unused fields in the various ItemType creation functions, but it seemed inconsistent (typical values of 0, 1, or 255, for example). I finally settled on setting them to the maximum integer for the field type, so at least the script calls would be consistent for newly-made ItemProperty instances.


DrMcCoy:
It's very possible that the original just doesn't initialize the unused variables at all, so what you see as inconsistent values is simply garbage memory.

For us to initialize it to something consistent is a good idea, yes.

rjshae:
Thanks. I asked about the ItemPropertyDamageReduction call on a forum and was informed by a knowledgeable person that the property can only be added to an item via the Toolset or as an effect, so we can probably either ignore that function or have it return an invalid item property.

src/engines/nwn2/itemproperty.cpp
```diff
@@ -76,10 +76,14 @@ uint8 ItemProperty::getItemPropertyCostTableValue() const {
...
 bool ItemProperty::getIsItemPropertyValid() const {
+	// Check if marked invalid
+	if (_type == kItemPropertyInvalid || _type >= kItemPropertyMAX)
+		return false;
+
 	// Load the item properties row
 	const Aurora::TwoDAFile &twoDA = TwoDAReg.get2DA("itempropdef");
 	const size_t count = twoDA.getRowCount();
-	if (_type >= count)
+	if (_type >= (ItemPropertyType) count)
```
DrMcCoy: Please do these casts the other way round, i.e. cast `_type` to a `size_t`. Probably won't hurt to use `static_cast` more too, instead of C-style casts.

src/engines/nwn2/script/functions_itemproperty.cpp
```diff
@@ -102,6 +105,704 @@ void Functions::getIsItemPropertyValid(Aurora::NWScript::FunctionContext &ctx) {
 		ctx.getReturn() = prop->getIsItemPropertyValid();
 }
 
+void Functions::itemPropertyAbilityBonus(Aurora::NWScript::FunctionContext &ctx) {
+	ItemPropertyType type = kItemPropertyAbilityBonus;
+	uint16 ability = (uint16) ctx.getParams()[0].getInt();
+	uint8 bonus = (uint8) ctx.getParams()[1].getInt();
+
+	// Check the range
+	if (ability >= (uint16) kAbilityMAX || bonus < 1 || bonus > 12)
```
DrMcCoy: We're probably going to need these different item property ranges in other places too, right? Might be useful to create `static constexpr` min/max values for that in src/engines/nwn2/types.h (like the `static const uint32 kClassInvalid` there; that one should be a constexpr now as well).

rjshae:
I've implemented your recommendations. Please have a look. Thanks.

Mingun:
`functions_itemproperty.cpp` now contains a lot of slightly different code. Can be to hide it behind a macro?

DrMcCoy:
I'm not sure I want to introduce macros here. A template, maybe. But I'm not sure this can be done without introducing even more lines for every use of the template.

Mingun:
The repeating code in many places differs only with a called method and names of variables. Use of a template here will be not really convenient therefore I offer a simple local macro, just for increase in readability and an insurance from typos.

DrMcCoy:
Hmm, true

rjshae:
Perhaps that can be done with a future commit? That way there's a baseline copy in case we need to revert.

DrMcCoy:
Merged as edb4d001d...ea12d5927, thanks! :)

PR #505, opened 2019-03-21
NWN2: Add unit test for NWN2 doors

rjshae:
For some reason the door 'isOpen()' call is returning true even
though the "AnimationState" parameter is set to zero for all four
doors. A cursory check of the code didn't turn up a reason, so for
now I left that test disabled.

Ed.: resolved.


rjshae:
Hmm, the AppVeyor error seems to be occurring in the 'door1changes' test. Most of those 'set' calls are fairly straight forward. The one that might be a concern is the 'Door::setLocked' call, which is running 'Door::evaluateLink'. As a test, I gave the first door a "LinkedTo" value of "exterior_door1", which doesn't actually exist, plus a "LinkedToFlags" value of "1". Perhaps during the '_module->getFirstObjectByTag()' call it is returning a valid object and going off the rails? But it runs fine on my PC.

rjshae:
Yep, it runs fine when I comment out the 'setLocked' call.

rjshae:
It looks like the 'isOpen()' call is returning true because of the exception handling in  'Door::loadAppearance'. That sets _invisible to true, which causes '_state' to be set to 'kStateOpened1'. Hence, that test can be safely ignored, since it doesn't actually check the imported 'AnimationState' value.

DrMcCoy:
Merged as 40035ba53, thanks! :)

PR #508, opened 2019-03-23
NWN2: Implement Store class

rjshae:
I believe this will require a minor merge with a couple of the script files
in pull request #504.


src/engines/nwn2/area.cpp
```diff
@@ -559,6 +564,14 @@ void Area::loadCreatures(const Aurora::GFF3List &list) {
 	}
 }
 
+void Area::loadStores(const Aurora::GFF3List &list) {
+	for (Aurora::GFF3List::const_iterator s = list.begin(); s != list.end(); ++s) {
+		Store *store = new Store(**s);
```
Mingun: For what reason create `Store` on the heap? I think, this introduce memory leak. Stack allocation will be enough

DrMcCoy: `loadObject()`: https://github.com/xoreos/xoreos/blob/master/src/engines/nwn2/area.cpp#L515

That method puts the object into the area's objects list, which then manages the memory of the object.
Yes, that it takes a reference was maybe not a good way to signal that it takes over the object.

DrMcCoy: Since we're now C++11, that ObjectList should maybe be a `std::list<std::unique_ptr<Engines::NWN2::Object>>`? And loadObject then takes a `std::unique_ptr<Engines::NWN2::Object>` and `std::move()`s it into the list?

That's something for a future PR, though, I guess. And also something the other engines need to be changed to do as well.

rjshae: Do I need to create a unique_ptr then release it to the loadObject call?

DrMcCoy: No. Right now, the code is fine as is. You're using loadObject as intended.

It could be just a bit clearer/nicer, but that's something for a new PR.

Mingun: Ok, then may be add comment about this in this PR, just to reduce confusing for newcomers?

src/engines/nwn2/store.cpp
```diff
@@ -71,6 +71,46 @@ void Store::setStoreMaximumBuyPrice(int32 max) {
 	_maxBuyPrice  = (max < 0) ? -1 : max;
 }
 
+Item *Store::getFirstItemInInventory() {
+	// Cycle through the inventory types
+	for (InventoryType it = 0; it < kInventoryTypes; it++) {
+		// Retrieve the first item in this inventory
+		Item *item = _inventory[it]->getFirstItemInInventory();
```
DrMcCoy: Since you're filling in the `_inventory` array by reading from the GFF, you can't be certain all types are in there. So you need to check that `_inventory[it]` is != nullptr before dereferencing.

Not just here, but everywhere you're accessing parts of `_inventory`.

src/engines/nwn2/inventory.h
```diff
@@ -55,13 +55,13 @@ class Inventory {
 	ItemSlots equippedItems;
 	ItemSlots inventoryItems;
 
-	uint16 _lastRetrieved { UINT16_MAX }; ///< Index of last retrieved item.
+	size_t _lastRetrieved { UINT32_MAX }; ///< Index of last retrieved item.
```
DrMcCoy: `SIZE_MAX` is the correct value to use.

src/engines/nwn2/area.cpp
rjshae: I added a comment.

DrMcCoy:
Why are the CIs breaking for this PR? o_O

DrMcCoy:
@rjshae, you're just doing a push -f from your local repo into rjshae/xoreos on GitHub, right? Nothing special, nothing weird?

rjshae:
@DrMcCoy I'm not good at this git stuff so I just did a reset and added the last four changes back in, then did a forced push.

rjshae:
Thank you for the review. Hopefully my edits have addressed your concerns. Sorry about the mix-up.

DrMcCoy:
Merged as f72310e20...ba987288d, thanks! :)

PR #511, opened 2019-03-27
NWN2: Implement Journal class and associated functionality

rjshae:
Campaigns have their own journals, so I pass the campaign  'module.jrl' file content
down to the module instance for storage. Each time a new module is loaded from the
campaign, it just replicates the same campaign journal as the module journal. For
stand-alone modules, it instead loads the corresponding 'module.jrl' file content for
the module. The journal state information is carried on the PC.

As far as I know, the individual module journal files are empty for a campaign, which
is why I didn't try to load them on top of the campaign journal. At least the ones I
checked in the OC were empty. But this possibility is the reason about why I had
misgivings about whether to store the campaign journal separately and whether to
copy it each time. For now I played it safe.

I added a note about the unavoidable iconv() error that now shows up when the OC
is loaded. Hopefully that can be addressed at some point.


src/engines/nwn2/journal.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+Journal::Journal(const Aurora::GFF3Struct &journal) {
+
+	load(journal);
+}
+
+Journal::~Journal() {
+}
+
+uint32 Journal::getJournalQuestExperience(const Common::UString plotID) const {
```
DrMcCoy: `const Common::UString &plotID`

src/engines/nwn2/journal.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+	return (category != _categories.end()) ? category->XP : 0;
+}
+
+void Journal::load(const Aurora::GFF3Struct &journal) {
+	if (!journal.hasField("Categories"))
+		return;
+
+	const Aurora::GFF3List &categories = journal.getList("Categories");
+	for (Aurora::GFF3List::const_iterator c = categories.begin(); c != categories.end(); ++c) {
```
DrMcCoy: You can use a range-based for-loop for iterating through a GFF3List as well.

Should be
```cpp
for (const Aurora::GFF3Struct *c : categories) {
}
```

src/engines/nwn2/journal.cpp
```diff
@@ -0,0 +1,86 @@
...
+		category.name       = (*c)->getString("Name");
+		category.identifier = (*c)->getString("Tag");
+		category.comment    = (*c)->getString("Comment");
+		category.priority   = (*c)->getUint("Priority");
+		category.XP         = (*c)->getUint("XP");
+
+		category.entries.clear();
+		if ((*c)->hasField("EntryList")) {
+			const Aurora::GFF3List &entries = (*c)->getList("EntryList");
+			for (Aurora::GFF3List::const_iterator e = entries.begin(); e != entries.end(); ++e) {
```
DrMcCoy: (The range-based for-loop works in that nested loop here as well)

src/engines/nwn2/journal.h
```diff
@@ -0,0 +1,72 @@
...
+#include "src/common/ustring.h"
+
+namespace Engines {
+
+namespace NWN2 {
+
+class Journal {
+public:
+	Journal(const Aurora::GFF3Struct &journal);
+	~Journal();
```
DrMcCoy: Since the destructor doesn't do anything special, you can mark that as

```cpp
~Journal() = default;
```

(and then remove the `Journal::~Journal()` from the .cpp file)

rjshae:
Upon further investigation, it looks like the Journal entry states are stored in the character .bic files as local int variables that are named: "NW_JOURNAL_ENTRY" + sPlotID

For now I'm just adding some 'TODO' notes in the appropriate locations.

rjshae:
The files have been updated. Thank you for the review.

DrMcCoy:
Merged as 1b3ee9564...e77e4b7cf, thanks! :)

PR #512, opened 2019-03-27
GRAPHICS: Flash Font

Nostritius:
This PR adds support for SWF flash fonts used by both dragonage games. Since the rendering needs to know what are the neighbour characters of a character, because of kerning adjustment, I decided to move the functionality into a class independent from the rest of the font system. This PR also adds cairo as an additional dependency which is used by the FlashFont class and will be used more extensively in future scaleform PRs.


src/graphics/aurora/flashfont.h
```diff
@@ -0,0 +1,107 @@
...
+	double _ascent;
+	/** Distance between the font baseline and the bottom. */
+	double _descent;
+
+	/** Advance values of font. */
+	std::vector<int16> _advanceTable;
+	/** Kerning values of the font. */
+	std::vector<::Aurora::GFXCharacter::KerningCode> _kernings;
+
+	std::map<uint16, Glyph> _glyphs;
```
DrMcCoy: If you're using `std::vector` and `std::map`, you should `#include` `<vector>` and `<map>`.

src/graphics/aurora/flashfont.h
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Aurora {
+
+class FlashFont {
+public:
+	/**
+	 * Create a new flash font based on a font extracted from a gfx
+	 * file
+	 * @param font the font to render
+	 */
```
DrMcCoy: Minor style remark: that Doxygen comment is not in our usual style.

src/graphics/aurora/flashfont.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "src/common/encoding.h"
+#include "src/common/rect.h"
+
+#include "src/aurora/gfxfile.h"
+
+#include "src/graphics/aurora/flashfont.h"
+
+#ifdef ENABLE_CAIRO
+
+static cairo_user_data_key_t kUserDataKey;
```
DrMcCoy: Why does that need to be a global?

src/graphics/aurora/flashfont.cpp
```diff
@@ -0,0 +1,208 @@
...
+	cairo_user_font_face_set_render_glyph_func(_font, &FlashFont::renderGlyph);
+	cairo_user_font_face_set_text_to_glyphs_func(_font, &FlashFont::textToGlyphs);
+}
+
+FlashFont::~FlashFont() {
+	cairo_font_face_destroy(_font);
+}
+
+cairo_status_t FlashFont::renderGlyph(cairo_scaled_font_t *scaled_font, unsigned long glyph, cairo_t *cr,
+									  cairo_text_extents_t *extents) {
```
DrMcCoy: Please use spaces for alignment.

src/graphics/aurora/flashfont.cpp
```diff
@@ -0,0 +1,208 @@
...
+	FlashFont *font = reinterpret_cast<FlashFont *>(
+			cairo_font_face_get_user_data(
+					cairo_scaled_font_get_font_face(scaled_font),
+					&kUserDataKey
+			)
+	);
+
+	info("%i", static_cast<uint16>(glyph));
```
DrMcCoy: `%i` (which is the same as `%d` for printf) is signed. `glyph` is an unsigned long, so `%ul` would be the correct format expression, and you don't need a `static_cast` there then.

src/graphics/aurora/flashfont.cpp
```diff
@@ -0,0 +1,208 @@
...
+					cairo_scaled_font_get_font_face(scaled_font),
+					&kUserDataKey
+			)
+	);
+
+	const std::map<uint16, Glyph> &glyphmap = font->_glyphs;
+	const std::vector<::Aurora::GFXCharacter::KerningCode> &kernings = font->_kernings;
+
+	(*glyphs) = cairo_glyph_allocate(utf8_len);
+	(*num_glyphs) = utf8_len;
```
DrMcCoy: You don't need these parentheses here.

But in fact, since we're C++, you might just make the method take a reference instead of a pointer. Especially since then no one can give it a nullptr (which you don't check for, so that would explode).

src/graphics/aurora/flashfont.cpp
```diff
@@ -0,0 +1,208 @@
...
+			cairo_font_face_get_user_data(
+					cairo_scaled_font_get_font_face(scaled_font),
+					&kUserDataKey
+			)
+	);
+
+	const std::map<uint16, Glyph> &glyphmap = font->_glyphs;
+	const std::vector<::Aurora::GFXCharacter::KerningCode> &kernings = font->_kernings;
+
+	(*glyphs) = cairo_glyph_allocate(utf8_len);
```
DrMcCoy: You probably need to check that
- `utf8_len` is `>=` (or `>`?) 0
- `cairo_glyph_allocate()` returns != nullptr

src/graphics/aurora/flashfont.cpp
Nostritius: Because the methods renderGlyph and textToGlyphs are static and need a key to get the proper FlashFont class which is set as user data in a cairo_font_face_t. But it can be made const.

DrMcCoy: But this basically equivalent to having a global `FlashFont *`.

I see why you're doing this, because the glyph callback needs to know about the font, and the usual Cairo method is by going through the user data (instead of, for example, a `void *` parameter).

However, the way you're doing this, this means you can ever only have one Font at the same time.

If it were an integer, I would have suggested `Common::generateIDNumber()`, but Cairo does it in a weird way, by using the memory address.

So you should probably rather have (non-static) `cairo_user_data_key_t` member variable in FlashFont, and use that.

(This might still clash if we're using Cairo from different threads for certain thread implementations, but the probability there is low)

Nostritius: I think you misunderstand something here. The problem with your suggested approach is, that renderGlyph and textToGlyph are static and don't have a direct reference or pointer to the specific FlashFont object. The FlashFont class is for example extracted in line 80 to 85 in flashfont.cpp using the user data key, meaning i cannot get access to the user data without the key. The key associated user data is unique in every font as written [https://www.cairographics.org/manual/cairo-cairo-font-face-t.html#cairo-font-face-set-user-data](here), meaning you could use for every FlashFont object set as user data of a cairo_font_face_t the same key since every cairo_font_face_t contains exactly one FlashFont. I know that this solution is not ideal, but i cannot think of a better approach for this c-style callbacks.

DrMcCoy: Aaah, okay, if the key is unique to every font, yeah, that would work.

That's just...very weird of Cairo.

src/graphics/aurora/flashfont.cpp
DrMcCoy: (Oh, right, since the function's signature is predetermined by Cairo, you of course can't make those references. My bad :P)

Nostritius:
The issues should now be fixed, and the requested additional things except for the automake stuff are added.

DrMcCoy:
Thanks! :)

I can write the automake/autoconf macro for that, no problem, when we're getting around to merging it.

Nostritius:
I took the freedom of adding a small FAQ entry explaining my reasons for using cairo

DrMcCoy:
Now that the release is finally done (yeah, sorry for taking so long), I think I can "unfreeze" this.

Just to make sure, @Nostritius, this is still something you want to see merged, right? If so, I'll re-review this in the coming days and write the autotools macro to check for Cairo.

The conflicts are just in the distribution files, that should be a quick fix, nothing breaking.

Nostritius:
To be honest, after a quick search for the current state on software based libraries for shape rendering, I found [Blend2D](https://blend2d.com/) which is relatively new and comes without all that stuff cairo has and we don't need. According to their own benchmark they are even way faster then cairo (which is in our case highly important, since I realized, that generating all shapes required for example for the main menu of dragonage takes a lot of time, about 20 seconds). It was published after I made the decision to stick with cairo.
What do you think about it dependency integration wise and compared to cairo? I want to give it a try at least, to see if it fullfills our needs better than cairo.

DrMcCoy:
If it's something small we can drop into xoreos, I'd vastly prefer that over adding a dependency on something huge like Cairo, yes.

Looking at Blend2D, though... powered by asmjit? Oof, not sure that's a good idea. CPU architecture support: x86/amd64, ARM currently only partially. That doesn't really sound like that's really future-proof for porting xoreos.

DrMcCoy:
To further my thoughts on this:

What I would absolutely prefer is if we could pull in a something small, potentially single-file. Not sure something like that exists, specifically for all the features we need.

I'm not all too happy with adding a dependency on Cairo, since it's a huge beast (and it increases by cross-compilation burden). But if we need it to get everything we're using, okay, then that's what we need to do. Cairo is at least an established project, so we can be sure it's reasonable stable and works.

I'm going to veto anything that decreases portability (by depending on specific architectures, for example), sorry. And do we even need heavy-duty JIT stuff? Or are we throwing mountains at ants here?

Does this make sense?

Nostritius:
After thinking closely about this and taking a look into blend2d, I would say, it would be better to stick for the moment with cairo, and keep an eye on potential alternatives.

DrMcCoy:
Okay, thanks for researching this. :)

What's the status on this PR now, then? Is it in a state you want to see merged?

Nostritius:
Yes, I would cocnsider this ccode as effectively finished

DrMcCoy:
(WIP branch with this PR merged onto the current master + autotools support: https://github.com/xoreos/xoreos/tree/cairo . I haven't looked at the code yet beyond making it compilable.)

PR #513, opened 2019-03-31
COMMON: Add ability to set thread names

clone2727:
Now that we're using std::thread instead of SDL threading, we lost the ability to set thread names. This restores that on some platforms.

I was only able to test this on Linux, but @DrMcCoy says this at least compiles in MinGW right now. It would be good if someone could test compiler this on FreeBSD/OpenBSD/Mac/MSVC and then make sure it actually works on said platforms.

Since Linux is limited to 15 characters and we need to truncate beyond that, I did not bother with a corresponding "get" function. It's really a set-and-forget function that should only be used for debugging.

### DIFF
--- a/src/common/thread.cpp
+++ b/src/common/thread.cpp
@@ -29,6 +29,23 @@
 #include "src/common/thread.h"
 #include "src/common/util.h"
 
+// Include whatever headers are necessary to allow for naming the thread
+#if defined(__linux__)
+	#include <sys/prctl.h>
+#elif defined(__APPLE__)
+	#include <dlfcn.h>
+#elif defined(__MINGW32__) || defined(__FreeBSD__) || defined(__OpenBSD__)
+	#include <pthread.h>
+#endif
+
+#if defined(_WIN32)
+	#include <boost/scope_exit.hpp>
+
+	#define WIN32_LEAN_AND_MEAN
+	#define NOMINMAX
+	#include <windows.h>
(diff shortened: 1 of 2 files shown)


Nostritius:
@clone2727 I tested your naming code under OSX and the threads are properly named.

clone2727:
@Nostritius Awesome, thanks!

DrMcCoy:
(@seedhartha, since you're on Windows and using MSVC IIRC, can you check if that works there? The thread name should show up when running xoreos in the debugger, and a recent WinDbg on Windows 10 release 1607 should also show them when attaching while xoreos is already running)

seedhartha:
Sure, I'll check.

seedhartha:
Not sure about WinDbg, but the internal debugger in VS certainly does display thread names.

DrMcCoy:
Nice, thanks! :)

Considering we haven't tested it on the BSDs, but I haven't really tested *anything* on the BSDs, is it good to be merged? Before or after the coming release?

seedhartha:
I think it's good to be merged, yes.

DrMcCoy:
Hmm, no idea why that fell from my view. I'll recheck that this still compiles and works in the evening, then merge if so.

DrMcCoy:
Okay, merged as de2e9ad1ed7251db34a2f27527c0aa871f44d74a, thanks! :)

PR #514, opened 2019-04-04
ODYSSEY: Adjust icon to be in the middle of the widget

Nostritius:
A small fix to adjust the lock icon in both kotor games like it is in the original


DrMcCoy:
Merged as 53da8a8ac, thanks! :)

PR #517, opened 2019-04-07
KOTORBASE: Minigame preparation

Nostritius:
I investigated a bit, how the swoop racing in both kotor games works. The swoop racing tracks are stored inside modules suffixed with "mg" they are essentially basic areas with some extra information and some specialy objects like obstacles and accelerators. This PR adds some basic features for the swoop races.


src/engines/kotorbase/script/functions_time.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "src/aurora/nwscript/functioncontext.h"
+
+#include "src/engines/kotorbase/script/functions.h"
+
+namespace Engines {
+
+namespace KotORBase {
+
+void Functions::getTimeHour(Aurora::NWScript::FunctionContext &ctx) {
+	ctx.getReturn() = static_cast<int32>(((EventMan.getTimestamp() / 1000) / 60) / 60);
```
DrMcCoy: Are you sure that's correct?

In NWN, these functions return the current in-game time, which moves faster than wall-time, and resets when starting a new game, etc.

Nostritius: Pretty sure, the functions are used in the swoop racer minigame to calculate your achieved time.

DrMcCoy:
Merged as 9fa0be1da...3772f1a7b, thanks!

PR #521, opened 2019-04-16
NWN: Fix shader based scrollbar rendering

mirv-sillyfish:
Abuse texture wrap to render scrollbars, modifying the texture
matrix as necessary.


DrMcCoy:
Merged as 51bba525b, thanks! :)

(I did fix-up the style a bit, though.)

The assertion is still there; but it's not happening on every scrollbar. I think that's something else than the scrollbar.

```
xoreos: /home/drmccoy/projects/xoreos/xoreos/src/graphics/aurora/texturehandle.cpp:86: Graphics::Aurora::Texture& Graphics::Aurora::TextureHandle::getTexture() const: Assertion `!_empty' failed.
```

PR #522, opened 2019-04-18
KOTORBASE: Add factions

Nostritius:
This PR adds initial support for factions, which determine if an object is an enemy or friendly.


DrMcCoy:
Merged as 51bba525b...8e12d510d, thanks! :)

PR #525, opened 2019-05-03
BUILD: Remove boost regex

Nostritius:
This PR removes boost regex from the depndencies and replaces it through c++11 regex


DrMcCoy:
Merged as ba1d6ffe0...f5c46f838, thanks! :)

(I added commits to remove Boost.Regex from our autotools files as well, as well as the documentation and distribution build files)

PR #527, opened 2019-05-06
WITCHER: Add CAuroraSettings data structure and initialization

rcxrdx:
The CAuroraSettings data structure shared between C++/LUA is added from reversing the game in version 1.4.5.1304.
The fields have their original names (and thus original naming convention) to make the individual fields bindings easier.


DrMcCoy:
Oh, neat, thanks! :)

(I'm a bit swamped with a move at the moment, so replies might be spotty. Sorry about that.)

However...I don't really want the original naming convention in xoreos' code. It clashes with our own naming conventions. Could you change that?

A few other things:
- Could you meld the fixup in 2739aec25208168d6a1184a538ebfb975f2217e1 into the commit that originally introduced that line? That way, our history stays clear and concise.
- b3f459b1832f057955c40d68f76763610839c575 and 7ed093429219372d1b76d8d110c37013621083aa also look like they should be one commit, not two.
- Can you get rid of the merge commit and instead rebase your changing on top of the current master? If you don't know how to do that, no worries, I'll do that when I merge the PR once it's good to go.

rcxrdx:
Thanks for the comments, I have done those modifications.

src/engines/witcher/lua/bindings.h
```diff
@@ -424,6 +424,667 @@ class LuaBindings {
 	static int luaBor(lua_State *state);
 };
 
+struct CAuroraSettingsData {
+	uint32_t _staticTimeFlow;
```
DrMcCoy: Can you use `uint32` instead of `uint32_t`, to be consistent with the rest of the codebase?

(Yes, doing that was a mistake on my part, back when I started xoreos, but now we're kinda stuck with it.)

src/engines/witcher/lua/bindings.h
```diff
@@ -424,6 +424,667 @@ class LuaBindings {
 	static int luaBor(lua_State *state);
 };
 
+struct CAuroraSettingsData {
```
DrMcCoy: Just make it `AuroraSettingsData`, please.

src/engines/witcher/lua/bindings.h
```diff
@@ -424,6 +424,667 @@ class LuaBindings {
...
+	uint32_t _showFloatingEmotionalState;
+	uint32_t _dontYouEverTouchMyGamma;
+	uint32_t _skipCutscenesOneByOne;
+	uint32_t _speedTreeRenderBranches;
+	uint32_t _speedTreeRenderFronds;
+	uint32_t _speedTreeRenderLeafCards;
+	uint32_t _speedTreeRenderBillboards;
+	uint32_t _dumpLoadingTextures;
+
+	CAuroraSettingsData();
```
DrMcCoy: This won't compile, since you add the constructor definiton in a later commit.

But also, can you do C++11-style inline initializations here instead? I.e.
`float _attackLockDist { 6.0f };`

src/engines/witcher/lua/bindings.cpp
```diff
@@ -142,11 +142,17 @@ void LuaBindings::registerBindings() {
 	LuaScriptMan.endRegister();
 }
 
+CAuroraSettingsData g_auroraSettings;
```
DrMcCoy: Does this have to be a *global* global? Can it be a static within `LuaBindings::luaGetAuroraSettings()` instead?

DrMcCoy:
Hi @rcxrdx, are you still interested in working on this?

PR #528, opened 2019-05-06
KOTORBASE: Enemy coloring

Nostritius:
This PR adds code which colors all objects, which belong to a hostile faction in the red style


DrMcCoy:
Merged as 7bc48cfe0...0bcca6775, thanks! :)

PR #529, opened 2019-05-07
GRAPHICS: New renderer fps counter

mirv-sillyfish:
Display fps counter using shader render path, if enabled.


DrMcCoy:
Merged as d73475099, thanks! :)

PR #530, opened 2019-05-10
KOTORBASE: Add that non-creature object dialogs are properly started

Nostritius:
This PR fixes the start of non-creature dialogs. It enables the playing of the dialog in the third room and properly assigns the Sith troopers as enemys.


DrMcCoy:
Merged as 1c963f406, thanks! :)

PR #532, opened 2019-05-18
GRAPHICS:  Integrate ImGui

Nostritius:
This PR integrates ImGui into the engines rendering pipeline. It makes use of freetype for font rendering and is able to switch between OpenGL2 and OpenGL3 rendering, but i was not able to test the second one due to my own hardware limitations at the moment. I added a wrapper class usable for defining own imguis as Renderable. I also added  an implementation of the wrapper with the ImGui demo for debugging purposes.
Bildschirmfoto zu 2019-05-18 19-13-21


DrMcCoy:
Oh, neat :)

What's that with gl3w, though? Can we use glew instead?

Nostritius:
Yeah, i have to define the variable IMGUI_IMPL_OPENGL_LOADER_GLEW in automake, but have no clue how to do that in automake. You can see my last try for that in 6adc14e. Can you give me a hint how to do that properly in automake? I have up to no experience with that.

DrMcCoy:
`AC_DEFINE` is correct in theory. It does, however, put the preprocessor variable into the generated config.h file. I.e. from the source, it only works if you include that config.h beforehand.

Which we can do for our own sources, but that's not the correct way if you want to set something for third-party code. Unless you want to modify it by adding that include directive to it.

Otherwise, you can put that, with a "-D" into a `SOMETHING_CFLAGS` variable like we do for `GLEW_CFLAGS`, call `AC_SUBST` on it and add that to our Makefile.am. That way, it'll be added to the compiler call as a general command line parameter.

Nostritius:
Now, the automake stuff should work.

DrMcCoy:
Nothing personal, but I'd like to keep that PR until after the coming release.

This PR looks like it could be the base for a complete rewrite of our debug console at least, i.e. it needs more work until it makes sense to ship with a release.

Does this make sense?

Nostritius:
Hmm, considering the fact, that the ImGui renderer basically hooks into the render path (that's by the way one of the reasons why I chose cairo over another shape rendering library to circumvent an extension of the render path) and changes some vital parts of the graphics related systems, I would approve to keep this rather big change for after the release.

DrMcCoy:
Since you're talking about Cairo here, does this PR depends on #512?

Nostritius:
No, definitely not

DrMcCoy:
Okay, I merged this as 56e822c5c...95db9720b, thanks! :)
I did, however, upgrade it to Dear ImGui 1.77.

I guess this now opens up the work field of adapting our debug console (and FPS counter) to use Dear ImGui.

PR #533, opened 2019-05-19
KOTORBASE: Prevent infinite distance from camera

Nostritius:
I discovered, that especially in the first kotor2 level on peragus, the camera got stuck and applied really long distances between the character. This PR should prevent this by applying an upper bound to the distance.


DrMcCoy:
Merged as ed62aeff2, thanks! :)

PR #535, opened 2019-06-10
IMAGES: Add support for animated tpc images

Nostritius:
This PR ports the changes from the xoreos-tools tpc class to xoreos for animated tpc files.


DrMcCoy:
Merged as beb6b8afe...8b869d766, thanks! :)

PR #536, opened 2019-06-22
BUILD: Add build folder to the include path for win32/xoreos.rc

seedhartha:
Otherwise, the MinGW build fails because of the `config.h` file not being
found.


DrMcCoy:
Merged, thanks! :)

PR #539, opened 2019-06-27
COMMON: Restore SDL based Semaphore class

seedhartha:
This fixes deadlock with anti-aliasing enabled.

### DIFF
--- a/src/common/rules.mk
+++ b/src/common/rules.mk
@@ -89,6 +89,7 @@ src_common_libcommon_la_SOURCES += \
     src/common/aabbnode.h \
     src/common/random.h \
     src/common/mutex.h \
+    src/common/semaphore.h \
     $(EMPTY)
 
 src_common_libcommon_la_SOURCES += \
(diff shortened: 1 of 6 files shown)


DrMcCoy:
Hmm, I thought the mutex + condition_variable basically is a semaphore?

I'm too tired right now to really thing it through, but IIRC this was the StackOverflow thread I read about it back then? https://stackoverflow.com/questions/4792449/c0x-has-no-semaphores-how-to-synchronize-threads

seedhartha:
Correct. However, semaphore is a particular system object in Windows and Linux (SDL uses that). It is also a concept of its own, which we currently do not implement in our code.

seedhartha:
As a side note, I believe it was a bad idea to switch to C++11 threads after all: there were definitely some performance and compatibility losses.

DrMcCoy:
Hmm, okay, that is a bit disappointing. I had hoped moving to C++11 threads would *increase* performance and compatibility. :/

DrMcCoy:
Merged, thanks :)

PR #540, opened 2019-07-21
JADE: Character Generation

Nostritius:
This PR tries to implement as much as possible the Jade Empire character generation menu and generates the player character.

### DIFF
--- a/src/engines/jade/gui/chargen/rules.mk
+++ b/src/engines/jade/gui/chargen/rules.mk
@@ -20,7 +20,9 @@
 # Character generation in Jade Empire.
 
 src_engines_jade_libjade_la_SOURCES += \
+    src/engines/jade/gui/chargen/characterinfo.h \
     $(EMPTY)
 
 src_engines_jade_libjade_la_SOURCES += \
+    src/engines/jade/gui/chargen/characterinfo.cpp \
     $(EMPTY)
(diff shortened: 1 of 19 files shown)


src/engines/jade/gui/chargen/rules.mk
```diff
@@ -0,0 +1,26 @@
...
+# You should have received a copy of the GNU General Public License
+# along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+
+# Character generation in Jade Empire.
+
+src_engines_jade_libjade_la_SOURCES += \
+    $(EMPTY)
+
+src_engines_jade_libjade_la_SOURCES += \
+    $(EMPTY)
```
DrMcCoy: Missing a new line at the end of the file

Nostritius:
The issues should now be fixed. About the exception, you mentioned, it's about the background of the main menu. While i fixed the error for the character generation, this will almost certainly make more problems for example, when going back to the main menu from the game.

DrMcCoy:
Now starting the game doesn't work anymore. It just returns to back the menu when it should display the first area.

Nostritius:
Ok, now it should work

DrMcCoy:
Yeah, now it does

Merged, thanks! :)

PR #543, opened 2019-09-22
KOTORBASE: Load creature abilities and weapon properties, damage calculation, death

seedhartha:
This PR effectively opens up three more doors on the Endar Spire.


DrMcCoy:
Merged as ccbcb88df...207b377b5, thanks! :)

PR #546, opened 2019-10-06
KOTORBASE: Add SwitchPlayerCharacter nwscript method

Nostritius:
This PR implements the SwitchPlayerCharacter nwscript method. It allows the switch in the kotor2 prologue from the main character to T3M4. I hope, i did everything right, concerning object creation and destruction.


DrMcCoy:
Apparently not, since it invokes UB :P

sanitizer log here: https://gist.github.com/DrMcCoy/94d189fb6136cf2f5e5316dde6fdaaa3

src/engines/kotorbase/module.cpp
```diff
@@ -179,6 +180,26 @@ Creature *Module::getPC() {
 	return _pc;
 }
 
+void Module::switchPC(unsigned int index) {
+	if (!_partyController.isAvailableCreature(index))
+		throw Common::Exception("%u is not an available npc id", index);
+
+	Creature *newPC = createCreature(_partyController.getAvailableNPCTemplate(index));
+	ObjectMan.registerObject(newPC);
```
seedhartha: Object registration is handled automatically when the object is created and destroyed. See https://github.com/xoreos/xoreos/blob/90003390a2c8b908335e90517e6580142e3b4a49/src/engines/kotorbase/object.cpp#L54

src/engines/kotorbase/module.cpp
```diff
@@ -179,6 +180,26 @@ Creature *Module::getPC() {
...
+	int oldLeaderID = _partyController.getPartyLeaderID();
+
+	Creature *oldPC = _partyController.getPartyLeader();
+	_partyController.setPartyLeader(index);
+	_partyController.removePartyMember(oldLeaderID);
+	_pc = newPC;
+	notifyPartyLeaderChanged();
+
+	_area->removeObject(oldPC);
+	removeObject(*oldPC);
```
seedhartha: `Module::removeObject` is called implicitly in `Area::removeObject`. See https://github.com/xoreos/xoreos/blob/90003390a2c8b908335e90517e6580142e3b4a49/src/engines/kotorbase/area.cpp#L872

src/engines/kotorbase/module.cpp
```diff
@@ -179,6 +180,26 @@ Creature *Module::getPC() {
...
+void Module::switchPC(unsigned int index) {
+	if (!_partyController.isAvailableCreature(index))
+		throw Common::Exception("%u is not an available npc id", index);
+
+	Creature *newPC = createCreature(_partyController.getAvailableNPCTemplate(index));
+	ObjectMan.registerObject(newPC);
+	_partyController.addPartyMember(index, newPC);
+	int oldLeaderID = _partyController.getPartyLeaderID();
+
+	Creature *oldPC = _partyController.getPartyLeader();
```
seedhartha: Player character will not always be the party leader. Maybe we should modify `Module::_pc` instead?

Nostritius:
The double free should now be fixed. About the use of _pc, I'm unsure. Can _pc be considered the player generated character or the currently controlled character and therefore being equal to the party leader?

seedhartha:
Currently, Player Character and Party Leader are two separate roles, which may or may not be represented by a single object. PC is instantiated once per module (`Module::loadPC()`), while the party leader a.k.a. currently controlled character, is a changing reference to either the PC or one of the party members. SwitchPlayerCharacter should probably recreate the PC object (`Module::_pc`), regardless of his status in the party, while persisting `Module::_pcInfo` and `Module::_chargenInfo` of the original PC.

DrMcCoy:
Yeah, I guess we have *three* separate concepts here, in the KotOR games.

1. The originally created player character
2. The current player character, which can be different when the game switches to another POV (during the tutorial, for example)
3. The party leader

Maybe we should call 1. something else, and keep that around permanently too. Especially because we need to remember the inventory and other things that might modify this entitiy during gameplay.

Module::_pc would then be pointer or copy of this entity (again, maing sure to keep everything in sync).

DrMcCoy:
The crash is gone, btw. But skipping the prologue doesn't work now:

```
WARNING: Failed running dialog script "a_end_001"
    Because: Failed running engine function "SwitchPlayerCharacter" (11)
    Because: 4294967295 is not an available npc id
```

Also, maybe we should investigate how the original keeps track of the original PC first?

DrMcCoy:
Just to make sure, this is still currently blocked by us not really knowing how this should be implemented, right?

Might be worth waiting to see what seedhartha will do in his reone project there :P

Nostritius:
Yes, I'm still not sure how to best realize this.

PR #547, opened 2019-10-12
KOTORBASE: Script parameters in dialogues

Nostritius:
This PR implements scripting parameters for kotor2 dialogues which enable to play the real prologue dialogue.


src/engines/kotorbase/gui/dialog.cpp
```diff
@@ -56,6 +56,14 @@ DialogGUI::DialogGUI(Module &module) :
...
 	try {
 		_dlg.reset(new Aurora::DLGFile(name, owner));
+		_dlg->setParameterCallback(
+				std::bind(
+						&DialogGUI::callbackParameter,
+						this,
+						std::placeholders::_1,
+						std::placeholders::_2
+				)
+		);
```
DrMcCoy: Hmmm.. I'm not so sure about that.

Shouldn't the parameters belong to the script, and as such be a property of `Aurora::NCSFile` that can be set and queried?

That would get rid of the need to add a callback there. And I also kind want to add any external state to a script that can be overwritten by another script. (In the future, I think we will need to run several scripts somewhat in parallel, at least for the heartbeat scripts?)

Nostritius:
I have changed it now, so that the parameters are a part of the NCSFile Object and are accessed from that.

DrMcCoy:
Merged, thanks! :)

PR #548, opened 2019-10-16
BUILD: Add arm64 build and exclude osx from it

Nostritius:
After i realized, that travis now offers the possibility to build for arm, i added some extensions to the travis file to build all linux targets for arm64.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,18 +10,24 @@ os:
   - linux
   - osx
 
+arch:
+  - amd64
+  - arm64
+
 env:
   global:
     # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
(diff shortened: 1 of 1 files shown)


DrMcCoy:
Oh, nice. But you also need to make sure Coverity Scan doesn't trigger for arm builds. I.e. you need to add that to the `coverity_scan_run_condition`.

Nostritius:
I've added now a condition to the coverity scan to only compile with x86_64 architecture

DrMcCoy:
Merged, thanks! :)

DrMcCoy:
Oh, btw, can you do that for xoreos-tools and phaethon as well, please?

Nostritius:
Of course, i will do so

PR #549, opened 2019-10-27
ACTIONSCRIPT: Time and interval handling

Nostritius:
This PR enables time and interval handling in the actionscript engine with an abstract interface and implements this in the newly introduces eclipse engine.


Nostritius:
@jdm Thanks for the hint, i fixed that one

PR #551, opened 2019-12-14
AURORA: Add oodle1 decompression

Nostritius:
For anothe project of mine, I started to look into gr2 meshes, and wrote code which can be ported to xoreos. The beginning for this work is the oodle1 decompression which is heavily used in the gr2 format. The code for that is taken from @Arbos nwn2mdk and and "xoreosified" and wrapped into an interface similar to the other compression functions. Do I have to add also the Apache2 license, since the code is under it?

### DIFF
--- a/src/aurora/oodle.cpp
+++ b/src/aurora/oodle.cpp
@@ -1,0 +1,371 @@
+/* xoreos - A reimplementation of BioWare's Aurora engine
+ *
+ * xoreos is the legal property of its developers, whose names
+ * can be found in the AUTHORS file distributed with this source
+ * distribution.
+ *
+ * xoreos is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * xoreos is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


DrMcCoy:
Ooooh, very nice, thanks! :)

However, there is one major-ish issue I see with the `Parameters` bitfield. Or rather, the issue is that *is* a bitfield. Bitfields are, unfortunately, not portable. The layout of the bits are implementation-defined, so you can't really be sure that what you specify there works for all platforms and all compilers.

Accessing the bitfield parts needs to be replaced with the equivalent bitwise-ORs and shifts. They're only used in the Dictionary constructor, right, and read-only?

Nostritius:
I have now replaced bitfield struct through reading with our BitStream class

DrMcCoy:
I'll take a further look at that tomorrow. But I think it should be in a state fit for merging already, yeah?

Nostritius:
Yes, I do not intend to further work on it. As I said in the beginning, this is mostly the exact code from nwn2mdk, wrapped into a decompress function. I have already successfully read gr2 files with that code in the mentioned project.

DrMcCoy:
Merged it as 856ce1e95e64a0012622fdeaf1e0bcef1e20d499, thanks! :)

PR #553, opened 2020-02-24
BUILD: Fix arm64 travis build

Nostritius:
This PR (hopefully) fixes the arm64 build with some modifications on the travis script.


DrMcCoy:
Merged, thanks! :D

Looks good so far, let's hope that works for a while :)

PR #556, opened 2020-04-17
ACTIONSCRIPT: Fix cascading inheritance

Nostritius:
This PR fixes a really nasty bug with inheritance in actionscript, where it was not possible to use objects of classes with more than one parent class.


DrMcCoy:
Yeah, this is a PR that slipped by me when I was in the release funk. I'll take a look at it tomorrow, thanks :)

(The CI failure was just an intermittant broken macOS VM. That should be fixed already)

DrMcCoy:
Merged as 856ce1e95...56e822c5c, thanks! :)

PR #561, opened 2020-08-12
COMMON: Replace SWAP with std::swap

clone2727:


DrMcCoy:
Merged as eb788fe1cafc5109902aeb887848e059df091e96, thanks! :)

PR #562, opened 2020-08-12
COMMON: Implement ARRAYSIZE using constexpr

clone2727:


DrMcCoy:
Merged as c4de39cdd97678e69c8fb24b130597bd280b6438, thanks! :)

PR #563, opened 2020-08-25
GRAPHICS: Use std::numeric_limits over literals

mirv-sillyfish:
Use std::numerics_limits<type>::max() to indicate invalid index
values rather than the more error prone literals of 0xFFFFFFFF.


DrMcCoy:
Merged as 9a608b37d5dae6d0a3b081c8acb5c7430c6504e7, thanks! :)

PR #564, opened 2020-08-25
KOTORBASE: Fix ODR violation

lacc97:
GCC 10.2 reports an ODR warning due to a mismatch in type Engines::KotORBase::Game. Specifically, the type ambiguity of the `_console` member means in some source files it gets resolved as Engines::Console and sometimes as Engines::KotORBase::Console.

This fixes the ambiguity.

### DIFF
--- a/src/engines/kotorbase/game.h
+++ b/src/engines/kotorbase/game.h
@@ -66,7 +66,7 @@ public:
 	virtual void run() = 0;
 
 protected:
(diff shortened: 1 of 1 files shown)


DrMcCoy:
Good catch! Merged as b347a6514bfe1c65d86778b7f27478f620b0e601, thanks! :)

PR #565, opened 2020-10-17
AURORA: Add support for gr2 files

Nostritius:
This PR adds support for gr2 files and fixes two things in the oodle1 decompressor.


DrMcCoy:
Merged as 5f0d7f3e3...746e65bae, thanks! :)

PR #572, opened 2021-12-30
COMMON: Add some prep work for working up to removing UString

clone2727:
As has long been wanted, we want to switch from using `Common::UString` to using `std::string` everywhere. It's saner to do this in a piecemeal fashion rather than attempting to do this all at once.

This starts off some of the work to create a new set of `Common::String` utilities in `common/string.h` while doing a bit of other related cleanup.

### DIFF
--- a/src/common/string.cpp
+++ b/src/common/string.cpp
@@ -29,6 +29,8 @@
 
 #include "src/common/string.h"
 
+#include "external/utf8cpp/utf8.h"
+
 namespace Common {
 namespace String {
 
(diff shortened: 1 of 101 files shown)


src/common/string.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include <memory>
+#include <system_error>
+
+#include "src/common/string.h"
+
+namespace Common {
+namespace String {
+
+std::string format(const char* format, ...) {
```
DrMcCoy, line 38: That should be `const char *format`. Same in `formatV()` and in the .h file too

DrMcCoy (approved):
Apart from the little style thing in string.h/.cpp, this looks good to me, thanks. Feel free to fix it up and merge (via rebase, no merge commit) :)

src/common/string.cpp
clone2727: If that's my only asterisk mistake, I'll gladly take it ;)

Issue #335, opened 2018-08-19
AURORA: ERF Compression 2 and 3


Nostritius:
The xbox version of Dragon Age 2 uses two new compressions with the ids 2 and 3 in the erf files which cannot be recognized from ther ERFFile class at the moment.

Nostritius:
My suspicion is, that it is lzo since bioware used this compression before Dragon Age 2 in the Mass Effect 2 pcc archives.

logzero:
According to @gibbed

None = 0, // all platforms
BiowareZlib = 1, // all platforms
LZMA = 2, // PS3
XMemLZX = 3, // XBOX 360
HeaderlessZlib = 7, // all platforms

http://svn.gib.me/public/bioware/trunk/Gibbed.Bioware.FileFormats/EncapsulatedResourceFile.cs

gibbed:
Thanks for the ping, you made me realize I hadn't removed the old repo for my Bioware code. That code has been migrated to GitHub:

[EncapsulatedResourceFile.cs#L384](https://github.com/gibbed/Gibbed.Bioware/blob/master/projects/Gibbed.Bioware.FileFormats/EncapsulatedResourceFile.cs#L384)

(I'll be removing the old repo when I have time, so I don't want to leave you hanging in that case)

clone2727:
With the merge of #577 and #578, this should now be handled.

PR #577, opened 2022-01-19
AURORA: Add support for DA2 PS3 ERF LZMA decompression

clone2727:
BioWare seems to have created their own (?) custom chunked LZMA format here, which at least seems minimal on top of LZMA1. This parses through that.

DA2 PS3 gets further along now, actually able to parse the initial room before failing to load any mesh (new format?).

This should be half of #335.

### DIFF
--- a/src/aurora/erffile.cpp
+++ b/src/aurora/erffile.cpp
@@ -28,6 +28,10 @@
 
 #include <cassert>
 
+#ifdef HAVE_CONFIG_H
+	#include "config.h"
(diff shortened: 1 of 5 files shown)


DrMcCoy (approved):
lgtm, thanks! :D

Can you apply it to xoreos-tools and Phaethon, too, please?

clone2727:
Need to make a change to fix some files that potentially add extra bytes mid-stream.

PR #578, opened 2022-01-21
AURORA: Add support for Xbox LZX compressed ERFs

clone2727:
As used by Dragon Age II Xbox 360, which should fix the other half of #335.

This uses code from libmspack for the actual decompression routine (internalized since the decompression API is not exposed normally) and UEViewer (for the light Xbox wrapping of the normal MS LZX decompression).

All of libmspack is included, but we may cut down on that. I left it as compiling as C code for ease-of-maintenance -- there are no changes to the library so it's very much just a copy/paste in to handle an update.

### DIFF
--- a/external/mspack/rules.mk
+++ b/external/mspack/rules.mk
@@ -1,0 +1,78 @@
+# xoreos - A reimplementation of BioWare's Aurora engine
+#
+# xoreos is the legal property of its developers, whose names
+# can be found in the AUTHORS file distributed with this source
+# distribution.
+#
+# xoreos is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 3
+# of the License, or (at your option) any later version.
+#
+# xoreos is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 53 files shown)


external/mspack/rules.mk
```diff
@@ -0,0 +1,78 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with xoreos. If not, see <http://www.gnu.org/licenses/>.
+
+# libmspack (<https://www.cabextract.org.uk/libmspack/>)
+
+noinst_LTLIBRARIES += external/mspack/libmspack.la
+external_mspack_libmspack_la_SOURCES =
+
+external_mspack_libmspack_la_CFLAGS = -Iexternal/mspack
```
DrMcCoy, line 25: That needs to be `-I$(srcdir)/external/mspack`, otherwise out-of-tree builds fail.

src/common/lzx.cpp
```diff
@@ -0,0 +1,167 @@
...
+	WriteStream *output;
+	int rest;
+};
+
+int readWrapper(mspack_file *file, void *buffer, int bytes) {
+	StreamHolder *stream = static_cast<StreamHolder *>(file);
+	if (!stream->input)
+		return 0;
+
+	ReadStream& input = *stream->input;
```
DrMcCoy, line 79: `ReadStream &input` :P

DrMcCoy (approved):
Apart from these two changes, the PR lgtm, thanks! :)

PR #601, opened 2024-04-07
DOCS: Update copyright year to 2024

BJNFNE:


DrMcCoy:
In general, I tend to update the copyright year for the xoreos repositories only after I have actually added/changed something in the repository. The reasoning being, that as long as there haven't been any changes, there's really no "legitimate" claim I could stake for that time frame.

Also, it seems this PR breaks the encoding of the xoreos.rc file. The change replaces the copyright symbol with the Unicode replacement character (0xEFBFBD, UTF-8 encoded).

(Also, also, it seems I need to update the GitHub actions pipeline for MSVC. Looks like it's related to the xz fallout, since currently the GitHub repo for that is down, and vcpkg pulls that in as a dependency. Hopefully just updating vcpkg is enough, but it's always a bit brittle.)

BJNFNE:
OK, after a long time of being the PR opened, i decided to close it. Thank you for your Review :)

PR #603, opened 2024-11-07
BUILD: Remove boost::shared_ptr and boost::scoped_array usage

clone2727:


DrMcCoy:
Merged as c3bf15ac8...3b13030fd, thanks! :)

PR #604, opened 2024-11-08
BUILD: Switch to C++17

clone2727:
Switches to C++17 and replaces some related boost-using code with std-using code.


DrMcCoy:
Merged, thanks! :)

PR #610, opened 2025-04-10
Make CMake find SDL2 on Haiku

Peppersawce:
A small tweak to the FindSDL2 file so CMake can find SDL2 on Haiku without user input. By @Begasus


DrMcCoy:
Merged, thanks! :)

